PR #12, opened 2020-10-19
Containerised releases

### LINKED ISSUE #11: Github Actions build
Use GHA to build a new container image when a tag is created.

Shillaker:
Closes #11 

- Add release container build on push
- Split container into gRPC base (slow to build, not changed often) and faabric 
- Move containers to faasm org in Dockerhub
- Use CMake external projects for 3rd party deps


cmake/ExternalProjects.cmake
```diff
@@ -0,0 +1,62 @@
+include(FindGit)
```
Shillaker, line 1: This file captures all the 3rd party dependencies now

docker/faabric.dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM faasm/grpc-root:0.0.5
```
Shillaker, line 1: This image will be rebuilt as part of the CI/ release process

docker/grpc-root.dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM ubuntu:20.04
```
Shillaker, line 1: This image only needs to be rebuilt infrequently as it takes ages (building gRPC and protobuf)

src/mpi/CMakeLists.txt
```diff
@@ -24,13 +24,16 @@ set_target_properties(faabricmpi
...
+    install(TARGETS faabricmpi
+        ARCHIVE DESTINATION ${LIB_DIRECTORY}
+        LIBRARY DESTINATION ${LIB_DIRECTORY}
+        PUBLIC_HEADER DESTINATION ${HEADER_DEST}
+    )
 else()
-    set(LIB_DIRECTORY ${CMAKE_SYSROOT}/lib)
+    install(TARGETS faabricmpi
+        PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_PREFIX}/include/faabric/mpi
+    )
```
Shillaker, line 37: This native install didn't quite work, but we want to keep the wasm version the same

.github/workflows/tests.yml
```diff
@@ -8,41 +8,50 @@ on:
...
   tests:
+    if: github.event.pull_request.draft == false
     runs-on: ubuntu-20.04
     env:
       HOST_TYPE: ci
       REDIS_QUEUE_HOST: redis
       REDIS_STATE_HOST: redis
     container:
-      image: faabric/base:0.0.4
+      image: faasm/faabric:0.0.9
```
csegarragonz, line 18: Could we use enviroment variables here in some way?

docker/faabric.dockerfile
csegarragonz: I am sorry about being so annoying with versions but I don't understand the policy used here.

docker/grpc-root.dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+RUN apt-get update
+RUN apt-get install -y software-properties-common
+
+RUN apt install -y \
+   autoconf \
+   clang-10 \
+   clang-format-10 \
+   build-essential \
```
csegarragonz, line 10: re-order to be alphabetically sorted

docker/faabric.dockerfile
Shillaker: Unfortunately there isn't a policy at the moment, the different bits of the application have different versions and for now we'll just have to keep them in sync. Improving this is definitely a worthy task to add to the backlog.

.github/workflows/tests.yml
Shillaker: I get what you're saying, but AFAICT we would either have to template the GH actions file or set the variables at the repo/ organisation level. Do you have an idea of what we might do?

Shillaker:
Related to https://github.com/faasm/faasm/issues/314

PR #15, opened 2020-10-23
Faasm-SGX Integration

J-Heinemann:


include/faabric/util/func.h
```diff
@@ -51,7 +51,7 @@ namespace faabric::util {
 
     bool isValidFunction(const faabric::Message &msg);
 
-    std::string funcToString(const faabric::Message &msg, bool includeId);
+    std::string funcToString(const faabric::Message &msg, bool includeId, bool includeSGX = false);
```
Shillaker, line 54: Instead of adding another parameter we can use the `issgx` parameter of the messages

src/executor/FaabricMain.cpp
```diff
@@ -17,6 +22,11 @@ namespace faabric::executor {
 
         conf.print();
 
+#if(FAASM_SGX)
+        // Check for SGX capability and create shared enclave
+        sgx::checkSgxSetup();
```
Shillaker, line 27: The Faasm subclass of this class should invoke `checkSgxSetup` rather than have it here as an `extern`. If we do it here we are polluting the superclass with knowledge of its subclasses. This also brings knowledge of the Faasm repo into Faabric which we want to avoid.

There may need to be a `beforePoolSetup` hook that subclasses can create so that we provide the ability for them to do this at the right point.

Shillaker (changes requested):
Couple of changes, but I'll implement them

PR #17, opened 2020-11-10
Containerised CLI

Shillaker:


docker/docker-compose-cli.yml
```diff
@@ -0,0 +1,22 @@
+version: "3"
+
+services:
+  redis:
+    image: redis
+
+  cli:
+    image: ${FAABRIC_CLI_IMAGE}
+    volumes:
+        - ..:/code/faabric
```
csegarragonz, line 10: Just for consistency, I think it would be better to mount at `/usr/local/code/faabric`.

tests/test/util/test_files.cpp
```diff
@@ -25,17 +25,15 @@ TEST_CASE("Test reading from a URL", "[util]")
     auto conf = faabric::util::getSystemConfig();
 
     // Skip if we're in CI
```
csegarragonz, line 27: Remove the comment (and update the `localPath` if you end up changing it).

csegarragonz (changes requested):
Nice! I've just added a very minor comment.

On another note, is leaving a newline at the end of shell scripts something we want to enforce? I wouldn't make it a _strong_ requirement as we currently have no way to lint/format it automatically, but just asking.

docker/docker-compose-cli.yml
Shillaker: I'd actually rather move away from `/usr/local/code` as it's not really a good place to put code (it means only a single user can have their code checked out on a shared machine). Unfortunately there are still a couple of places in Faasm where it assumes `/usr/local/code`. The pattern for our Docker containers should be `/code` for code and `/build` for build.

PR #18, opened 2020-11-11
Removed index and cold start interval

Shillaker:
Both are no longer necessary


src/util/json.cpp
```diff
@@ -22,101 +22,138 @@ std::string messageToJson(const faabric::Message& msg)
     d.AddMember("function",
```
Shillaker, line 22: The formatting in this file was quite messed up so I tidied it

csegarragonz (approved):

PR #20, opened 2020-11-13
Adding MPI Stub Definitions

csegarragonz:
This work is part of the necessary support for LAMMPS-MPI, and was therefore living in a `lammps-support` branch.

However, as it's getting too big, and more implementation than expected will be required, I've decided to chunk contributions into smaller, easier to review pieces.

PS: might be worth to do a release once _all_ LAMMPS is working (still some time for that to happen, tho :sweat_smile: ).

### DIFF
--- a/include/faabric/mpi/mpi.h
+++ b/include/faabric/mpi/mpi.h
@@ -419,6 +419,27 @@ extern "C"
                      MPI_Datatype recvtype,
                      MPI_Comm comm);
 
+    int MPI_Cart_create(MPI_Comm old_comm,
+                        int ndims,
+                        const int dims[],
+                        const int periods[],
+                        int reorder,
+                        MPI_Comm* comm);
+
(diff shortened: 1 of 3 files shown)


include/faabric/mpi/mpi.h
```diff
@@ -183,8 +189,13 @@ extern "C"
     extern struct faabric_info_t faabric_info_null;
 #define MPI_INFO_NULL &faabric_info_null
 
+// Misc constants
+#define MPI_ANY_SOURCE -1
+#define MPI_UNDEFINED -1
```
csegarragonz, line 194: `MPI_UNDEFINED` is used as a rank for processes that have been left out of a _new_ communicator.

For instance, creatnig a 3x3 grid of processes and running with `-np 10` will make one be completely idle (this is expected behaviour).

Shillaker (approved):

PR #21, opened 2020-11-13
MPI_Scan implementation

csegarragonz:
Implementation of the scan call, as in [`MPI_Scan`](https://www.open-mpi.org/doc/v4.0/man3/MPI_Scan.3.php).

Scan performs an inclusive prefix reduce. This is, process `i` will return the accumulated value for processes `[0, ..., i]` and pass it's result to the next process.


Shillaker (approved):
Great, nice and clean :+1:

PR #22, opened 2020-11-13
Adding cartesian communicator methods

csegarragonz:
As suggested in faasm/faasm#351

### DIFF
--- a/src/scheduler/MpiWorld.cpp
+++ b/src/scheduler/MpiWorld.cpp
@@ -206,6 +206,24 @@ void MpiWorld::getCartesianRank(int rank, int* dims, int* periods, int* coords)
     }
 }
 
+void MpiWorld::getRankFromCoords(int* rank, int* coords)
+{
+    int sideLength = static_cast<int>(std::floor(std::sqrt(this->size)));
+    int prank = 0;
+    int factor = 1;
+
+    int i = MPI_CART_MAX_DIMENSIONS - 1;
+    for (; i >= 0; --i) {
+        if (coords[i] == MPI_UNDEFINED) {
+            throw std::runtime_error(
+              "Cartesian rank with undefined coordinates.");
+        }
+        prank += factor * (coords[i] % sideLength);
+        factor *= sideLength;
+    }
+    *rank = prank;
+}
+
 int MpiWorld::isend(int sendRank,
                     int recvRank,
                     const uint8_t* buffer,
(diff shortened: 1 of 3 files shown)


src/scheduler/MpiWorld.cpp
```diff
@@ -176,6 +176,54 @@ std::string MpiWorld::getHostForRank(int rank)
...
+    }
+
+    // Compute the corresponding rank in a 2-dim grid given the original process
+    // rank.
+    // Note - this operation, when restricted to 2dims is quivalent to:
+    // coords = {rank / sideLength, rank % sideLength}
+    int sideLength = static_cast<int>(std::floor(std::sqrt(this->size)));
+    int nprocs = sideLength * sideLength;
+    if (rank >= nprocs) {
+        for (uint8_t i = 0; i < MPI_CART_MAX_DIMENSIONS; i++) {
```
Shillaker, line 193: Using an 8-bit unsigned integer as an array index and loop variable seems strange, what's the reason for this?

Shillaker, line 193: Using an 8-bit unsigned integer as an array index and loop variable isn't wrong, but it's not in the same style as the rest of the code (which would use an int). What's the reason for this?

src/scheduler/MpiWorld.cpp
```diff
@@ -176,6 +176,54 @@ std::string MpiWorld::getHostForRank(int rank)
...
+}
+
+void MpiWorld::getRankFromCoords(int* rank, int* coords)
+{
+    int sideLength = static_cast<int>(std::floor(std::sqrt(this->size)));
+    int prank = 0;
+    int factor = 1;
+
+    int i = MPI_CART_MAX_DIMENSIONS - 1;
+    for (; i >= 0; --i) {
```
Shillaker, line 216: Then the style of this for loop is also different, we usually initialise loop variables in the loop declaration (and this is using an int whereas the one above is using `uint8_t`)

Shillaker (changes requested):
Looks good other than questions around for loop syntax

src/scheduler/MpiWorld.cpp
csegarragonz: Funny enough, I had to change the type in this case (as using an unsigned int would overflow) but didn't bother looking what was used elsewhere.

Shillaker (approved):

PR #23, opened 2020-11-16
Add container flag and zsh comp.

csegarragonz:
Mimicking what's currently done in `faasm`.


Shillaker (approved):

csegarragonz:
Just as a self-reminder, this won't work until we bump the version, or force a new tag, as the image gets build from a tagged version of the repo.

PR #25, opened 2020-11-16
Encapsulate reduce operator

csegarragonz:
In this PR I abstract the current reduce logic from `reduce` itself to `op_reduce`. Reducing an operator, this is accumulating two arrays of `faabric_datatype_t` using a `faabric_operator_t`, is something we might need in more collective communication calls other than `reduce`.

For instance, `MPI_Scan`, which performs a [partial reduce](https://www.open-mpi.org/doc/v4.0/man3/MPI_Scan.3.php), would also use this method.

Further, I find it cleaner for implementation, testing, and debugging purposes to have the logic separated. Specially as we include more types and operators.

### DIFF
--- a/include/faabric/scheduler/MpiWorld.h
+++ b/include/faabric/scheduler/MpiWorld.h
@@ -123,6 +123,12 @@ class MpiWorld
                    int count,
                    faabric_op_t* operation);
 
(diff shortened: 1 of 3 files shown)


Shillaker (approved):
Nice, this is much better

PR #27, opened 2020-11-20
MPI_sendRecv support: implementation and test

csegarragonz:
[`MPI_sendRecv`](https://www.open-mpi.org/doc/v4.0/man3/MPI_Sendrecv.3.php) performs both a send and a receive, and ensures completion (i.e. no deadlocks) regardless of the order they are issued in.

In this PR I include the implementation in `faabric` together with a test case. As mentioned in the `TODO` I will now change the signature of `doISendRecv` to allow async `recv`s to have a message type as well. But this will be in a different PR.

This is, hopefully, the last call to be implemented prior to running LAMMPS @ Faasm :tada:

### DIFF
--- a/include/faabric/scheduler/MpiWorld.h
+++ b/include/faabric/scheduler/MpiWorld.h
@@ -86,6 +86,16 @@ class MpiWorld
 
     void awaitAsyncRequest(int requestId);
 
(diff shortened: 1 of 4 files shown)


Shillaker (approved):
Nice, looks great.

PR #31, opened 2020-11-23
Support for MPI_MIN in reduce operations

csegarragonz:
Profiling tells us what MPI calls are made, but not the actual parameters. Reducing with the `MPI_MIN` operation is needed in LAMMPS.

### DIFF
--- a/src/scheduler/MpiWorld.cpp
+++ b/src/scheduler/MpiWorld.cpp
@@ -804,6 +804,7 @@ void MpiWorld::op_reduce(faabric_op_t* operation,
             throw std::runtime_error("Unsupported type for max reduction");
         }
     } else {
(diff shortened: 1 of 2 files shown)


src/scheduler/MpiWorld.cpp
```diff
@@ -748,62 +748,95 @@ void MpiWorld::op_reduce(faabric_op_t* operation,
 
     logger->trace(
       "MPI - reduce op: {} datatype {}", operation->id, datatype->id);
-    if (operation->id == faabric_op_sum.id) {
+    if (operation->id == faabric_op_max.id) {
```
csegarragonz, line 751: Re-shuffled the operations to match the order in `include/faabric/mpi/mpi.h`.

Shillaker (approved):

PR #35, opened 2020-11-30
Change the build flag to avoid extra CMake arguments

csegarragonz:
Before we had to pass a specific `CMake` flag (`FAASM_BUILD_TYPE`) in order to install the header and library files in the right `sysroot`. This is unnnecessary as we can directly query the system name in a WebAssembly build.

Note that this was making the `toolchain` installation and subsequent builds fail, as it would create broken symlinks to non-existing `faabric` header files.

### DIFF
--- a/src/mpi/CMakeLists.txt
+++ b/src/mpi/CMakeLists.txt
@@ -22,7 +22,8 @@ set_target_properties(faabricmpi
     PROPERTIES PUBLIC_HEADER "${PUBLIC_HEADERS}"
     )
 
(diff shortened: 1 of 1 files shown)


Shillaker (approved):

Shillaker:
Is there anywhere in the code we can now remove the `-DFAASM_BUILD_TYPE=wasm` argument?

csegarragonz:
Not in `faabric` but there are several references in `faasm`.

PR #38, opened 2020-12-09
Mirror testing scheme and compose tooling from Faasm + Version bump

csegarragonz:
In this PR I update `faabric`'s GH actions to match the ones we use in `faasm`. In particular:
* Split `formatting` and `tests` jobs to enhance parallelism and hopefuly reduce action times.
* Change from `check_black.sh` to `change_python.sh` and include the `flake8` dependency.
* Include `clang-tidy` check (ran it, and no file was changed).

As I have changed the dependencies in the `dockerfiles` I've also tagged a new release.

I have also moved the `docker-compose.yml` file to the repository root so that now the deployment can easily invoked/removed using `docker-compose up -d` and `docker-compose down` (the former is done through `bin/cli.sh` hence why I haven't included it in the docs).


Shillaker (approved):
Nice

PR #39, opened 2020-12-09
Improvements to using Faabric as an external library

Shillaker:


docker/faabric-cli.dockerfile
```diff
@@ -3,17 +3,6 @@ FROM faasm/faabric:$FAABRIC_VERSION
...
-RUN apt-get install -y \
-    libpython3-dev \
-    python3-dev \
-    python3-pip \
-    python3-venv
-
-WORKDIR /code/faabric
-COPY requirements.txt .
-RUN pip3 install -r requirements.txt
-
```
Shillaker, line 16: Do we still need this Dockerfile? Almost empty

.github/workflows/release.yml
```diff
@@ -32,11 +32,3 @@ jobs:
...
           tags: faasm/faabric:${{ env.TAG_VERSION }}
           build-args: FAABRIC_VERSION=${{ env.TAG_VERSION }}
-      - name: "Build Faabric CLI container"
-        uses: docker/build-push-action@v2
-        with:
-          push: true
-          file: docker/faabric-cli.dockerfile
-          context: .
-          tags: faasm/faabric-cli:${{ env.TAG_VERSION }}
-          build-args: FAABRIC_VERSION=${{ env.TAG_VERSION }}
```
Shillaker, line 42: The `faabric-cli` container is now redundant as we do everything in the `faabric` container. In addition, the `faabric` container isn't actually used anywhere as it is, so it makes sense to combine them.

.github/workflows/tests.yml
```diff
@@ -63,20 +63,45 @@ jobs:
...
-        run: ninja faabric_tests
-        working-directory: /build/faabric
+        run: inv dev.cc faabric_tests
       # --- Tests ---
       - name: "Run tests"
         run: ./bin/faabric_tests
-        working-directory: /build/faabric
+        working-directory: /build/faabric/static
+
+  examples:
```
Shillaker, line 76: This step checks building and using Faabric as a shared library

CMakeLists.txt
```diff
@@ -13,26 +13,17 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
-        add_library(${lib_name} SHARED ${lib_deps})
-    endfunction()
-    function(faabric_private_lib lib_name lib_deps)
-        add_library(${lib_name} STATIC ${lib_deps})
+function(faabric_lib lib_name lib_deps)
+    add_library(${lib_name} ${lib_deps})
+
+    add_library(${lib_name}_obj OBJECT ${lib_deps})
+
+    if(BUILD_SHARED_LIBS)
```
Shillaker, line 22: This function is just for building libraries internally, so we want to add the normal static library for linking between internal libraries as normal, then the `OBJECT` library to support bundling everything together at the end.

bin/run_clang_tidy.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-BUILD_PATH=/build/faabric/
+BUILD_PATH=/build/faabric/static/
```
Shillaker, line 5: Building staticallly is the default, so we assume this in development environments.

cmake/ExternalProjects.cmake
```diff
@@ -2,8 +2,6 @@ include(FindGit)
 find_package(Git)
 include (ExternalProject)
 
-include_directories(${THIRD_PARTY_INSTALL_DIR}/include)
```
Shillaker, line 5: We want to install 3rd party deps in the main `CMAKE_INSTALL_PREFIX` location, but we don't want to include everything in there when building, hence we do more fine-grained includes for each library below.

cmake/ExternalProjects.cmake
```diff
@@ -62,7 +66,9 @@ if(FAABRIC_BUILD_TESTS)
         GIT_TAG "v2.13.2"
         CMAKE_ARGS "-DCATCH_INSTALL_DOCS=OFF \
             -DCATCH_INSTALL_EXTRAS=OFF"
-        CMAKE_CACHE_ARGS "-DCMAKE_INSTALL_PREFIX:STRING=${THIRD_PARTY_INSTALL_DIR}"
+        CMAKE_CACHE_ARGS "-DCMAKE_INSTALL_PREFIX:STRING=${CMAKE_INSTALL_PREFIX}"
     )
+    ExternalProject_Get_Property(catch_ext SOURCE_DIR)
+    include_directories(${CMAKE_INSTALL_PREFIX}/include/catch2)
```
Shillaker, line 72: Catch2 requires different `#includes` depending on how it's installed. Doing it like this or from source means we have to write `#include <catch.hpp>` (as they do in their docs), rather than `#include <catch2/catch.hpp>`. This means there are hundreds of changes in the tests below.

include/faabric/endpoint/Endpoint.h
```diff
@@ -1,9 +1,9 @@
 #pragma once
 
+#include <faabric/proto/faabric.pb.h>
```
Shillaker, line 3: This is the other side-effect of this change which I think makes much more sense, having the Faabric protobuf-generated headers in the normal include location avoids having to specify a special `include_directories` when using Faabric as a 3rd party dep.

src/mpi/CMakeLists.txt
```diff
@@ -13,10 +13,11 @@ set(PUBLIC_HEADERS
...
 
-add_library(faabricmpi STATIC 
-        mpi.cpp
-        ${PUBLIC_HEADERS}
-    )
+if(BUILD_SHARED_LIBS)
+    add_library(faabricmpi SHARED mpi.cpp ${PUBLIC_HEADERS})
+else()
+    add_library(faabricmpi STATIC mpi.cpp ${PUBLIC_HEADERS})
+endif()
```
Shillaker, line 20: This just needs to respect the static/ shared request of the user.

src/proto/CMakeLists.txt
```diff
@@ -1,41 +1,59 @@
-set(OUTPUT_FILES
-        ${CMAKE_CURRENT_BINARY_DIR}/faabric.pb.cc
-        ${CMAKE_CURRENT_BINARY_DIR}/faabric.pb.h
-        ${CMAKE_CURRENT_BINARY_DIR}/faabric.grpc.pb.cc
-        ${CMAKE_CURRENT_BINARY_DIR}/faabric.grpc.pb.h
-        )
+# ----------------------------------------------
```
Shillaker, line 1: Changes in this file are just a refactor and steps to copy the generated headers into place.

include/faabric/endpoint/Endpoint.h
csegarragonz: Been there :rofl:

csegarragonz (approved):

Issue #43, opened 2020-12-11
Fix broken build


csegarragonz:
The current _clean_ build is broken, as it fails to link with some headers. To replicate run:
```bash
inv dev.cmake --clean
inv dev.cc faabric
```

PR #44, opened 2020-12-11
Fix includes for inclusion as an external project

Shillaker:
When using Faabric as an external project it can either be added with `add_subdirectory` in which case we need the public `target_include_directories` or as an installed library in which case the transitive header includes should still work.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,17 +13,8 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 # Compile comamnds for clang tools
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-# Library funcs
-function(faabric_lib lib_name lib_deps)
-    add_library(${lib_name} ${lib_deps})
-
-    add_library(${lib_name}_obj OBJECT ${lib_deps})
-
-    if(BUILD_SHARED_LIBS)
-        target_compile_options(${lib_name} PRIVATE "-fPIC")
-        target_compile_options(${lib_name}_obj PRIVATE "-fPIC")
-    endif()    
-endfunction()
+# Global include dir
+set(FAABRIC_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/include)
 
 # Output directories
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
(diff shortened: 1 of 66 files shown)


CMakeLists.txt
```diff
@@ -13,32 +13,37 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
+set(FAABRIC_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/include)
+
+# Output directories
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+
+# External libraries
+include(cmake/ExternalProjects.cmake)
+
```
Shillaker, line 26: Just moved this bit up to put all config at the top

CMakeLists.txt
```diff
@@ -13,32 +13,37 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
     add_library(${lib_name}_obj OBJECT ${lib_deps})
 
+    # Include directories
+    target_include_directories(${lib_name} 
+        PUBLIC ${FAABRIC_INCLUDE_DIR}
+        PUBLIC ${PISTACHE_INCLUDE_DIR}
+        PUBLIC ${SPDLOG_INCLUDE_DIR}
+        PUBLIC ${RAPIDJSON_INCLUDE_DIR}
+        )
+
```
Shillaker, line 40: This is the key change, rather than using a global `include_directories` for this repo

csegarragonz:
Closes #43

csegarragonz (approved):

csegarragonz:
@Shillaker I was trying to rebuild this myself, but the `examples` build seems to fail (albeit the tests "pass"). In fact, if you check the [latest action](https://github.com/faasm/faabric/pull/44/checks?check_run_id=1551196689), under the tab `Build Examples` you will see that the build step for the `server` fails.

To replicate just run the steps in the workflow file:
```bash
inv dev.cmake --shared
inv dev.cc faabric --shared
inv dev.install faabric --shared
inv examples
```

I will now have a look into why the tests actually pass.

csegarragonz (approved):

Issue #42, opened 2020-12-10
Run and profile standalone MPI applications


csegarragonz:
The aim of this issue is to track the implementation of the support to run standalone (i.e. non-WASM) MPI applications. Ideally, this would link against `faabric` and `emulator` and will allow for a faster development life-cycle, easier debugging, and a more interesting profiling when implementing the new scheduler.

My idea would be that, rather than having a long-lasting monstruous PR, we could split the work in smaller chunks and keep track of them through this issue (if we manage to keep it updated). In essence, there are two main bits to be added to `faabric`:
1. Running standalone MPI applications.
2. Profiling tools for MPI applications.

### Running standalone MPI applications

Technically we should be able to remove WebAssembly from the picture, and compile applications against our own MPI implementation and with the emulator as a host interface wrapper. A non-exhaustive to-do list would be:
* [x] Port a hello world MPI test.
* [x] Compile the basic MPI test.
* [x] Port all function definitions.
* [x] Pass all the MPI tests.
* [x] Compile LAMMPS.
* [x] Run LAMMPS.

### Profiling MPI applications

This second part, if done right, could be specially useful and a cornerstone to achieve any type of performance with the new `faabric` scheduler. In particular it would consist in implementing a set of tools to aid profiling an MPI application at _runtime_. An even less exhaustive to-do list would be:
* [x] Centralize (as in aggregate) all different log files.
* [ ] Port the static analyzer tool.
* [ ] Port the dynamic profiler (as in which calls are made and how much time).
* [ ] Include time measurements.
* [ ] Present data in a reasonable manner?

#### Related PRs

* [x] Mpi Standalone Implementation - First Bits (#46): implements the underlying architecture and ports a hello world example.

csegarragonz:
### Using Faabric as a Library

During an offline discussion, we decided on the following approach to run MPI applications natively. The root of the problem is that MPI assumes processes as it's execution unit/model, whereas Faabric uses threads. Additional isolation and multi-tenancy guarantees are accomplished through WebAssembly. Thus, it's not apparent how can we simulate MPI processes whilst using Faabric (which assumes regions of shared memory, among other things).

Architecturally, we will have `MPI_WORLD_SIZE` faabric instances each running a single-thread (highly inneficient but bear in mind that this is just a workaround to run MPI applications without cross-compiling to wasm and profile our implementation). We would need:
* A Python script that takes in `MPI_WORLD_SIZE` (M).
* Script instantiates M containers.
[...]

PR #46, opened 2020-12-15
MPI Standalone Implementation - First Bits

csegarragonz:
This PR is the first bit towards running standalone MPI applications #42 . In it I include a simple executor implementation for MPI methods. It consists of a single thread execution thread in a singleton thread pool. I also implement support to run both `threadPool` and `endpoint` either in the background or in the foreground (depending on where we want our program to hang).

As of now, the way to port an MPI application is to change the `int main` line for a custom function (`faabric::executor::mpiFunc`), and prepend a different main entrypoint (this could be wrapped in a macro).

To test the hello world program, from a fresh instance:
```bash
inv container.build-mpi-native

# outside the container
# modify the values in ./docker/mpi_native.env
./bin/run_mpi_native.sh
```

To-Do:
* [x] Compilation of `mpi_helloworld`
* [x] Execution of `mpi_helloworld`
* [x] Scalability test of `mpi_helloworld`
* [x] Regression test

After this gets merged in:
* [x] Update the main issue tracker (#42)
* [x] Port the actual implementation of most MPI methods. (#51)
* [x] Aggregate log files from all different workers in a shared directory (#50).

To be discussed:
* [x] Running `invoke` from outside the container is a bit dodgy, migrate to standalone script?
* [x] Wrap all code in a macro?


examples/mpi_helloworld.cpp
```diff
@@ -0,0 +1,58 @@
+#include <faabric/util/logging.h>
+
+#include <faabric/mpi-native/MpiExecutor.h>
+#include <faabric/mpi/mpi.h>
+
+#include <unistd.h>
+
+int main(int argc, char** argv)
```
csegarragonz, line 8: This main function is what we'd need to add to standalone MPI binaries, and then provide a definition for `int faabric::executor mpiFunc()`.

cmake/ExternalProjects.cmake
```diff
@@ -30,6 +30,7 @@ set(PROTOC_EXE /usr/local/bin/protoc)
 set(GRPC_PLUGIN /usr/local/bin/grpc_cpp_plugin)                                  
 
 # Pistache 
+link_directories(${CMAKE_INSTALL_PREFIX}/lib)
```
csegarragonz, line 33: In my local development environment (and for the CI) I needed this to build the tests (otherwise linking to `pistache` failed).

tasks/mpi.py
```diff
@@ -0,0 +1,102 @@
...
+from os import environ
+from os.path import exists, join
+from subprocess import run
+
+from tasks.util.env import PROJ_ROOT, FAABRIC_INSTALL_PREFIX
+
+from invoke import task
+
+EXAMPLES_DIR = join(PROJ_ROOT, "examples")
+BUILD_DIR = join(EXAMPLES_DIR, "build")
```
Shillaker, line 11: Is this variable used?

tasks/mpi.py
```diff
@@ -0,0 +1,102 @@
...
+
+INCLUDE_DIR = "{}/include".format(FAABRIC_INSTALL_PREFIX)
```
Shillaker, line 13: Is this variable used?

tasks/mpi.py
```diff
@@ -0,0 +1,102 @@
...
+
+    # Start up `np` faabric instances. Note that if there are any other
+    # scaled out processes they will be stopped and removed. Additionally,
+    # running /bin/cli.sh whilst having a scaled out cluster will stop & remove
+    # them as well.
+    scale_up_cmd = "docker-compose up -d --scale {}={} {}".format(
+        FAABRIC_WORKER_NAME,
+        np,
+        "--force-recreate" if clean else "--no-recreate",
+    )
```
Shillaker, line 69: This file is quite awkward, and re-engineers some docker-compose functionality. I would have thought it would be easier to create a new docker-compose file for this native MPI stuff, with two services, one for the workers and one for the client. Then you could use vanilla `docker-compose` commands to control the workers directly (without having to do the filtering and scaling in here). You'd then need to create a similar script to `cli.sh` for connecting to the client.

Try to do as much with the `docker-compose` CLI and separate config file as you can, then see whether it's still necessary to have scripting around that, and perhaps just try a bash script or two.

Shillaker (changes requested):

docs/native_mpi.md
```diff
@@ -0,0 +1,51 @@
...
+`./bin/cli.sh`.
+
+Then, to compile the library:
+```bash
+inv dev.cmake --shared
+inv dev.cc faabricmpi_native --shared
+inv dev.install faabricmpi_native --shared
+```
+
+## Adapting the source code
```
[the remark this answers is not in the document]
csegarragonz: With this I mean how to run a native MPI binary with our implementation (i.e. right now, paste a bunch of code). But I agree that it is not important right now.

include/faabric/endpoint/Endpoint.h
```diff
@@ -13,12 +16,19 @@ class Endpoint
...
+    void stop();
 
     virtual std::shared_ptr<Pistache::Http::Handler> getHandler() = 0;
 
   private:
+    bool isBackground;
     int port = faabric::util::getSystemConfig().endpointPort;
     int threadCount = faabric::util::getSystemConfig().endpointNumThreads;
+    std::thread servingThread;
+    std::unique_ptr<Pistache::Http::Endpoint> httpEndpoint;
```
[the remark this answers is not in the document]
csegarragonz: The endpoint is currently used to bootstrap execution. This is, before instantiating the `SingletonPool` object that starts the thread pool, state server, and http endpoint, we make a call to the scheduler in `examples/mpi_helloworld.cpp:35` this in turn loads the function defined in `faabric::executor::mpiFunc` and at this point we don't need the endpoint anymore.

If running the HTTP endpoint in the background is something we definately won't need down the road, then these changes may add complexity for a _single_ call, but as previously mentioned they are needed for initialisation.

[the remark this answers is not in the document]
Shillaker: Do we need to go via the HTTP endpoint to make the call to the scheduler? Can't we do it directly? That way we avoid having the endpoint running at all.

src/mpi-native/mpi_native.cpp
```diff
@@ -0,0 +1,74 @@
...
+    return MPI_SUCCESS;
+}
+
+int MPI_Comm_size(MPI_Comm comm, int* size)
+{
+    auto logger = faabric::util::getLogger();
+    logger->debug("MPI_Comm_size");
+
+    auto& world = getExecutingWorld();
+    *size = world.getSize();
```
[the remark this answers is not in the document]
csegarragonz: My gut feeling is that most implementations will indeed be one-liners, so I think it is worth templating/macroing the function definitions. As discussed before, I'd like to try (also for educational purposes) a more C++ native approach (rather than macros).

However, I'd prefer to include this when I tackle #51

tasks/mpi.py
csegarragonz: I agree with you on this one. The approach I've gone for in the end is the following:

1. Custom `docker-compose.yml` file in `docker/mpi-native-docker-compose.yml`: it has less dependencies than the original compose as it's not supposed to be used as a cli, rather just a service to run one-off functions.
2. Deployment scripted using `bin/run_mpi_native.sh` and feeding input variables via `docker/mpi-native.env`: using the `.env` file fits well within the `docker-compose` architecture and permits not handling the input variables in `bash` which is quite cumbersome (I am happy to support variables as CL arguments, but I'd rather not do so :grin: ).

One problem I've run with is that I've needed to build a separate docker image with a different entrypoint. This is because I need to compile the libraries and the examples at build time, so that I can set the corresponding MPI binary as entrypoint. Alternatively, I could:
* Modify the original `faabric` base image to build the `faabricmpi_native` target and the examples (which I'd rather not do).
* Script around `docker-compose` again to ensure that everything is in place before we bootstrap execution.

Right now, the build time is around 3' so it's bearable.

include/faabric/endpoint/Endpoint.h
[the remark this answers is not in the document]
csegarragonz: You are completely right about this. I missunderstood some of the endpoint functionalities, I really don't need it _at all_. I will revert to the implementation in `master`.

docker/mpi-native.env
```diff
@@ -0,0 +1,10 @@
+FAABRIC_VERSION=0.0.16
```
csegarragonz, line 1: This file should not be tracked, follows up the offline discussion.

cmake/ExternalProjects.cmake
csegarragonz: Turns out there's the exact thing we need in CMake through the usage of [`IMPORTED`](https://cmake.org/cmake/help/git-stage/guide/importing-exporting/index.html#id5) libraries.

I've also had to tune a bit the `ExternalProject_Add` command. The difference with `spdlog` and `rapidjson` is that they both are header-only libraries hence work just fine with pastint the headers in the right include direction.

cmake/ExternalProjects.cmake
```diff
@@ -31,12 +31,18 @@ set(GRPC_PLUGIN /usr/local/bin/grpc_cpp_plugin)
...
     CMAKE_CACHE_ARGS "-DCMAKE_INSTALL_PREFIX:STRING=${CMAKE_INSTALL_PREFIX}"
+    BUILD_BYPRODUCTS ${CMAKE_INSTALL_PREFIX}/lib/libpistache.so
 )
 ExternalProject_Get_Property(pistache_ext SOURCE_DIR)
 set(PISTACHE_INCLUDE_DIR ${SOURCE_DIR}/include)
+add_library(pistache SHARED IMPORTED)
+add_dependencies(pistache pistache_ext)
+set_target_properties(pistache
+    PROPERTIES IMPORTED_LOCATION ${CMAKE_INSTALL_PREFIX}/lib/libpistache.so
+)
```
Shillaker, line 45: Nice, this looks relatively clean. Can you move faasm and faasm-toolchain to this version of Faabric to check they both still work after (or before) merging this?

docs/native_mpi.md
```diff
@@ -0,0 +1,20 @@
...
+
+Faabric supports linking MPI binaries against our custom MPI implementation
+used in [Faasm](https://github.com/faasm/faasm). This way, you can test the
+compliance of your MPI application with our API (a subset of the standard)
+without the burden of cross-compiling to WebAssembly.
+
+To run native MPI applications you need to first modify your binary matching
+the examples provided, and then build the worker image running:
+```
+inv container.build-mpi-native
```
Shillaker, line 11: Is this a bash script now?

Shillaker (commented):
Couple of minor requests

cmake/ExternalProjects.cmake
csegarragonz: Here is [faasm's](https://github.com/faasm/faasm/pull/388/checks?check_run_id=1733786695) and here the [toolchain's](). Should we merge those PRs?

csegarragonz: Here is [faasm's](https://github.com/faasm/faasm/pull/388/checks?check_run_id=1733786695) and here the [toolchain's](https://github.com/faasm/faasm-toolchain/pull/25/checks?check_run_id=1734076349). Should we merge those PRs?

Note that the toolchain one would not pass for some missmatch in the submodule build (which I am looking into) but afaik everything else worked fine (as to what this PR is concerned).

csegarragonz: @Shillaker ok for merge?

tasks/container.py
```diff
@@ -79,3 +88,11 @@ def push_grpc(ctx):
     Push current version of gRPC container
     """
     _do_push(GRPC_IMAGE_NAME)
+
+
+@task
+def push_mpi_native(ctx):
+    """
+    Push current version of gRPC container
```
Shillaker, line 96: Tiny typo

Shillaker (approved):
Tiny typo, then good to go

PR #48, opened 2021-01-06
Updating to latest pistache

Shillaker:


Shillaker:
This is no longer needed, we can stick with the same version for now.

PR #49, opened 2021-01-11
Flush as a top-level RPC

Shillaker:
Flush was previously a bit of a hack, implemented as a flag on an existing message. Instead, it should really be a top-level RPC operation so that each host can handle its own flushing operation (rather than letting an individual executor handle it on each host).

### DIFF
--- a/bin/run_clang_tidy.sh
+++ b/bin/run_clang_tidy.sh
@@ -4,26 +4,50 @@ set -e
 
 BUILD_PATH=/build/faabric/static/
 CONFIG=${FAABRIC_ROOT}/.clang-tidy
+DUMP_OUTPUT=true
 
-FILES=$(git ls-files {src,tests}/{"*.h","*.cpp","*.c"})
+# Function to actually run clang-tidy
+function do_tidy {
+    FILES="($@)"
+    run-clang-tidy-10.py \
+        -config '' \
+        -j `nproc` \
+        -fix \
+        -format \
+        -style 'file' \
+        -p ${BUILD_PATH} \
+        -quiet \
+        ${FILES}
+}
 
-if [ -z "$1" ]; then
-    :
-else
+# Support passing a file, directory or nothing
+if [ -f "$1" ]; then
+    echo "Running on file $1"
+    FILES=($1)
+
+    # Assume we want to see output if running on a single file
+    unset DUMP_OUTPUT
+
+elif [ -d "$1" ]; then
     pushd $1 >> /dev/null
+    echo "Running on directory $1"
     FILES=$(git ls-files {"*.h","*.cpp","*.c"})
+
+elif [ -z "$1" ]; then
+    echo "Running on cwd at $(pwd)"
+    FILES=$(git ls-files {func,libs,src,tests}/{"*.h","*.cpp","*.c"})
+
+else
+    echo "First argument must be a file or directory"
+    exit 1
 fi
 
-# Run clang-tidy on a set of files
-run-clang-tidy-10.py \
-    -config '' \
-    -j `nproc` \
-    -fix \
-    -format \
-    -style 'file' \
-    -p ${BUILD_PATH} \
-    -quiet \
-    ${FILES} > /dev/null
+# Dump output or not
+if [ -z "$DUMP_OUTPUT" ]; then
+    do_tidy ${FILES[@]}
+else
+    do_tidy ${FILES[@]} > /dev/null
+fi
 
 popd >> /dev/null || true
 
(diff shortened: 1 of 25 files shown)


include/faabric/scheduler/Scheduler.h
```diff
@@ -80,6 +80,8 @@ class Scheduler
     std::string getThisHost();
 
     void broadcastFlush(const faabric::Message& msg);
+    
+    void flushLocalNodes(const faabric::Message& msg);
```
Shillaker, line 84: We really need to change "nodes" to "executors" but it's used in too many places to do in this PR

include/faabric/executor/FaabricExecutor.h
```diff
@@ -42,8 +42,6 @@ class FaabricExecutor
 
     virtual void postFinish();
 
-    virtual void postFlush();
```
Shillaker, line 45: No need to split this into two now. Executors directly implement `flush()`

include/faabric/scheduler/Scheduler.h
```diff
@@ -46,7 +46,9 @@ class Scheduler
 
     std::string getFunctionWarmSetNameFromStr(const std::string& funcStr);
 
-    void clear();
+    void reset();
+
+    void shutdown();
```
Shillaker, line 51: Having a single `clear` function didn't make sense as there are really two responsibilities

src/executor/FaabricExecutor.cpp
```diff
@@ -20,24 +20,6 @@ FaabricExecutor::FaabricExecutor(int threadIdxIn)
     currentQueue = scheduler.getBindQueue();
 }
 
-void FaabricExecutor::flush()
```
Shillaker, line 23: This stuff is now done by the scheduler

src/executor/FaabricExecutor.cpp
```diff
@@ -150,7 +132,7 @@ std::string FaabricExecutor::processNextMessage()
     faabric::Message msg = currentQueue->dequeue(timeoutMs);
 
     std::string errorMessage;
-    if (msg.isflushrequest()) {
+    if (msg.type() == faabric::Message_MessageType_FLUSH) {
```
Shillaker, line 135: Rather than having boolean flag properties like `isflushrequest` we should use the existing message type field.

src/proto/RPCServer.cpp
```diff
@@ -20,14 +20,14 @@ void RPCServer::start(bool background)
     _isBackground = background;
 
     if (background) {
-        logger->debug("Starting state server in background thread");
+        logger->debug("Starting RPC server in background thread");
```
Shillaker, line 23: Old copy-paste error

src/proto/faabric.proto
```diff
@@ -10,6 +10,9 @@ service FunctionRPCService {
     rpc ShareFunction (Message) returns (FunctionStatusResponse) {
     }
 
+    rpc Flush (Message) returns (FunctionStatusResponse) {
+    }
```
Shillaker, line 14: These methods have to take a request and return a response type, even though in this case both are ignored.

src/scheduler/CMakeLists.txt
```diff
@@ -13,4 +13,4 @@ set(LIB_FILES
 
 faabric_lib(scheduler "${LIB_FILES}")
 
-target_link_libraries(scheduler faabricmpi redis)
+target_link_libraries(scheduler state faabricmpi redis)
```
Shillaker, line 16: This dependency is now needed to allow the top-level function server to flush state

bin/run_clang_tidy.sh
```diff
@@ -4,26 +4,50 @@ set -e
 
 BUILD_PATH=/build/faabric/static/
```
Shillaker, line 5: This was copied from Faasm

include/faabric/util/queue.h
```diff
@@ -14,6 +14,14 @@ class QueueTimeoutException : public faabric::util::FaabricException
     {}
 };
 
+class ExecutorFinishedException : public faabric::util::FaabricException
```
Shillaker, line 17: This lets executors tell the pool they're finished

include/faabric/util/queue.h
```diff
@@ -57,6 +66,34 @@ class Queue
 
     T dequeue(long timeoutMs = 0) { return doDequeue(timeoutMs, true); }
 
+    void waitToDrain(long timeoutMs)
```
Shillaker, line 69: This is needed by the scheduler for waiting for a flush to have completed.

csegarragonz (approved):
Nice! LGTM

PR #52, opened 2021-01-21
MPI Standalone Implementation - Bulk of the API Code

### LINKED ISSUE #51: [MPI] Port the remaining MPI calls

csegarragonz:
In this PR I include the implementation for all the remaining MPI API calls that are currently supported in `faasm`. Sorry for the size of this, but the _vast_ majority of the code is just linking with the corresponding `MpiWorld` call.

Another bit that increases the size of the PR by a lot are the examples that I have used (ported from `faasm`) to test the actual implementation. These now live in `mpi-native/examples` and get run as part of the tests.

Closes #51


include/faabric/mpi/mpi.h
```diff
@@ -70,7 +70,11 @@ extern "C"
         int worldId;
         int rank;
         int size;
-        int wasmPtr;
+        union
```
csegarragonz, line 73: This is the only change I've needed to make to the original interface (kudos to you for that). The reason being that we are now dealing with real memory and not offsets in the WASM memory array.

include/faabric/mpi/mpi.h
```diff
@@ -86,7 +90,7 @@ extern "C"
 
     struct faabric_request_t
     {
-        int _unused;
+        int id;
```
csegarragonz, line 93: Refactor for better readability as in `faasm` we make no difference between a pointer to an object, and a pointer to an object's first attribute (same address, but in faasm it's hidden under a layer of indirection, hence the first propertie's name does not really matter).

include/faabric/scheduler/MpiWorld.h
```diff
@@ -40,6 +40,8 @@ class MpiWorld
 
     int getSize();
 
+    void getSizePtr(int* size);
```
csegarragonz, line 43: These methods are not _necessary_ per se. However, to be able to use the function pointer trick (and save lines of code) I must ignore the return value (i.e. void preferably). Additionally, function overloading does not work well with declaring a pointer to the function, hence the easy workaround was to define two methods with the properties I needed.

The immediate workaround if you'd rather see these methods gone would be not using the `fptr` trick and perform the regular calls to `MpiWorld`.

mpi-native/examples/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
+cmake_minimum_required(VERSION 3.13.0)
```
csegarragonz, line 1: Copied the file from `examples/CMakeLists.txt` to have the mpi examples be a separate target.

include/faabric/mpi/mpi.h
Shillaker: A `union` is nice because it means there's only one set of code shared between both compilation targets, but I'm not sure it's the most common way to change things based on compilation target. I would say it's more normal to use preprocessor directives with things like `#ifdef __wasm__`. In this case you can kind of guess that one gets used in wasm and the other doesn't based on the names I guess.

I would have expected a type called something like `MPI_POINTER_TYPE` with an `#ifdef __wasm__` around it to decide whether it's `int` or `void*`. It's more explicit that way but if you add a comment to explain the union then that's ok too.

include/faabric/mpi/mpi.h
Shillaker: Hmm I'm not sure this is any clearer, is this really an `id`? `id` suggests to me that it's an identifier, and somehow assigned by the application to have some meaning. It seems to me more like a `self` pointer, so maybe `_self` is an appropriate name?

tasks/examples.py
```diff
@@ -56,6 +57,51 @@ def build(ctx, clean=False):
     )
 
 
+@task()
+def build_mpi(ctx, clean=False):
```
Shillaker, line 61: Let's move this to a new `tasks/native_mpi.py` as it's not really related to the examples

src/mpi_native/mpi_native.cpp
```diff
@@ -47,28 +65,777 @@ int MPI_Init(int* argc, char*** argv)
...
+    auto fptr = &faabric::scheduler::MpiWorld::send;
+    callMpiFunc(
+      fmt::format("MPI_Send {} -> {}", executingContext.getRank(), dest),
+      fptr,
+      executingContext.getRank(),
+      dest,
+      (uint8_t*)buf,
+      datatype,
+      count,
+      faabric::MPIMessage::NORMAL);
```
Shillaker, line 114: I'm not convinced by this function pointer approach. It adds a lot of indirection and we're now having to modify the main MPI code to support it, yet in this case I'm not sure it actually saves any lines of code.

Without the func pointer stuff (in the case where we only need `logger` once, and `world` once):

```
faabric::util::getLogger()->debug("MPI_Send {} -> {}", executingContext.getRank(), dest);

getExecutingWorld().send(executingContext.getRank(), 
...
```

With the function pointer stuff:

```
callMpiFunc(fmt::format("MPI_Send {} -> {}", executingContext.getRank(), dest),
    &faabric::scheduler::MpiWorld::send,
    executingContext.getRank(),
    ...
```

I think the first one is much easier to follow personally, and doesn't seem that much more verbose. If you want to save more characters you could also add `using namespace faabric::util` and/ or `using namespace faabric::scheduler`.

include/faabric/scheduler/MpiWorld.h
Shillaker: See comment below. I'm not sure the `fptr` trick adds enough value to justify changing the main codebase.

src/mpi_native/mpi_native.cpp
```diff
@@ -47,28 +65,777 @@ int MPI_Init(int* argc, char*** argv)
 int MPI_Comm_rank(MPI_Comm comm, int* rank)
 {
     auto logger = faabric::util::getLogger();
-    logger->debug("MPI_Comm_rank");
+    logger->debug("S - MPI_Comm_rank");
```
Shillaker, line 68: This `S - ` syntax is a hangover from Faasm's host interface implementation and doesn't really mean anything any more. What value does it add here? If we want to make log parsing easier we could come up with a better scheme (e.g. using different loggers for different system components, or having a `[MPI]` prefix for MPI-related logs.

src/mpi_native/mpi_native.cpp
```diff
@@ -47,28 +65,777 @@ int MPI_Init(int* argc, char*** argv)
...
+      executingContext.getRank(), dest, (uint8_t*)buf, datatype, count);
+    (*request)->id = requestId;
+
+    return MPI_SUCCESS;
+}
+
+/*
+ * The API call returns a request ID that we can't capture with the function
+ * pointer trick. Thus we don't use it here.
+ */
```
Shillaker, line 721: Not quite sure what you mean here but I'm not sure it's worth not doing something just to facilitate the function pointer idea.

tasks/examples.py
```diff
@@ -31,7 +31,7 @@ def build(ctx, clean=False):
         [
             "cmake",
             "-GNinja",
-            "-DCMAKE_BUILD_TYPE=Release",
+            "-DCMAKE_BUILD_TYPE=Debug",
```
Shillaker, line 34: Do we need to change the release type here? Ideally the examples would be treating Faabric as an external project, hence using the `Release` build (not a big deal though if there's a reason to change it).

Shillaker (changes requested):

include/faabric/mpi/mpi.h
csegarragonz: I agree that preprocessor directives are a better choice for things based on compilation targets. I don't think though that the `MPI_POINTER_TYPE` is needed as this only happens once.

I've used CMake's `add_compile_definitions` to add a preprocessor directive if we detect a `Wasm` MPI build.

Shillaker: I don't think we need to use an extra compile definition do we? Can we just use `#ifdef __wasm__`?

src/mpi_native/mpi_native.cpp
csegarragonz: This will be addressed in #55

include/faabric/mpi/mpi.h
csegarragonz: I wasn't aware `wasi` set this value somewhere upstream (I literally redefined `__wasm__`).

Shillaker (approved):
Nice, looks good.

I'm wondering if we want to put all our MPI test functions into a separate `experiments-mpi` repo, as they're quite useful and might get added to with other stuff. Might be overkill for now but something to think about, then both Faasm and Faabric could have a submodule or fetch-content to run their tests (and we'd avoid having all that code in the main repos).

PR #56, opened 2021-01-29
Logging enhancements

### LINKED ISSUE #55: Richer logging capabilities

csegarragonz:
In this PR I slightly modify the behaviour of the logging functions in `faabric` to make it more extensible and flexible. The changes are summarized below:

1. Instead of a default logger, there now is a map `<logger_name> -> <logger_shared_ptr>` containing all the different loggers.
2. They are all lazily initialised. I.e. created when first used.
3. To use a specific (or different) logger we just need to call the `faabric::util::getLogger` method with the logger's name as a parameter. Otherwise, it will resort to using the `default` logger.
4. I also introduce a new configuration flag `LOG_FILE` which if set turns on logging to a file at `/var/log/faabric/<logger_name>.log` with the log-level specified in `conf.logLevel`.

Note that there's no control on _how many_ loggers are created, and wrongly typing the logger name could produce unexpected results.

Closes #55

### DIFF
--- a/include/faabric/util/config.h
+++ b/include/faabric/util/config.h
@@ -56,6 +56,8 @@ class SystemConfig
 
     // MPI
     int defaultMpiWorldSize;
+    std::string mpiLogLevel;
+    std::string mpiLogFile;
 
     // Endpoint
     std::string endpointInterface;
--- a/include/faabric/util/config.h
+++ b/include/faabric/util/config.h
@@ -21,6 +21,7 @@ class SystemConfig
     std::string cgroupMode;
     std::string netNsMode;
     std::string logLevel;
+    std::string logFile;
     std::string pythonPreload;
     std::string captureStdout;
     std::string stateMode;
@@ -56,8 +57,6 @@ class SystemConfig
 
     // MPI
     int defaultMpiWorldSize;
-    std::string mpiLogLevel;
-    std::string mpiLogFile;
 
     // Endpoint
     std::string endpointInterface;
--- a/src/mpi_native/mpi_native.cpp
+++ b/src/mpi_native/mpi_native.cpp
@@ -25,7 +25,7 @@ faabric::scheduler::MpiWorld& getExecutingWorld()
 
 static void notImplemented(const std::string& funcName)
 {
-    auto logger = faabric::util::getLogger();
+    auto logger = faabric::util::getLogger("mpi");
     logger->debug("{}", funcName);
 
     throw std::runtime_error(funcName + " not implemented.");
@@ -45,7 +45,11 @@ int MPI_Init(int* argc, char*** argv)
         executingContext.joinWorld(*call);
     }
 
+    // Initialise MPI-specific logging
     int thisRank = executingContext.getRank();
+    faabric::util::initMpiLogging(thisRank);
+    faabric::util::getLogger("mpi")->debug("MPI_Init");
+
     faabric::scheduler::MpiWorld& world = getExecutingWorld();
     world.barrier(thisRank);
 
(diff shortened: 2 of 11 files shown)


src/util/config.cpp
```diff
@@ -27,6 +27,7 @@ void SystemConfig::initialise()
     cgroupMode = getEnvVar("CGROUP_MODE", "on");
     netNsMode = getEnvVar("NETNS_MODE", "off");
     logLevel = getEnvVar("LOG_LEVEL", "info");
+    logFile = getEnvVar("LOG_FILE", "on");
```
csegarragonz, line 30: Policy-wise, I've opted for:
* 1 single log-level specified in `conf.logLevel`
* If an additional env variable (`conf.logFile`) is set, then logging also goes to a specific file (with the same log-level).

src/util/logging.cpp
```diff
@@ -1,42 +1,61 @@
 #include <faabric/util/config.h>
 #include <faabric/util/logging.h>
 
-#include <spdlog/sinks/stdout_color_sinks.h>
+#include "spdlog/sinks/basic_file_sink.h"
+#include "spdlog/sinks/stdout_color_sinks.h"
```
Shillaker, line 5: Why quotes here? I think for "system" libraries/ 3rd party deps we would normally use `<...>`

src/util/logging.cpp
```diff
@@ -1,42 +1,61 @@
...
+    }
 }
 
-std::shared_ptr<spdlog::logger> getLogger()
+std::shared_ptr<spdlog::logger> getLogger(const std::string& name)
 {
-    if (!isInitialised) {
-        initLogging();
+    // Lazy-initialize logger
+    // TODO consider using loggers.contains(<key>) when C++20 support
```
Shillaker, line 54: Not sure we need this TODO, we do the `count() == 0` trick all over the place and `contains()` only saves a few characters. I'd save TODOs for things that would improve the logic/ performance rather than the syntax.

include/faabric/util/config.h
```diff
@@ -21,6 +21,7 @@ class SystemConfig
     std::string cgroupMode;
     std::string netNsMode;
     std::string logLevel;
+    std::string logFile;
```
Shillaker, line 24: This should be added to the `SystemConfig` test alongside log level

src/mpi_native/mpi_native.cpp
```diff
@@ -9,6 +9,7 @@
 using namespace faabric::executor;
 
 static faabric::scheduler::MpiContext executingContext;
+static std::string logName;
```
Shillaker, line 12: Have you actually tested this? Each MPI rank is a thread, but this is a single global variable, won't it keep getting overwritten when each rank sets it? It should be `thread_local`-ised.

Having the logger name as the global variable seems odd, why not just a shared pointer to the logger? I would do something like:

```
static thread_local std::shared_ptr<spdlog::logger> mpiLogger = nullptr;

std::shared_ptr<spdlog::logger> getMpiLogger(int mpiRank) {
    if(mpiLogger == nullptr) {
        mpiLogger = faabric::util::getLogger(fmt::format("MPI-{}", mpiRank));        
    }
    return mpiLogger;
}
```

Then just call `getMpiLogger(rank)` inline.

src/util/config.cpp
Shillaker: I don't think file logging should be on by default. In most execution environments the task manager will handle piping stderr/stdout to somewhere useful (e.g. `kubectl`/ or `supervisord`). More importantly, something like supervisor will handle log file rotation. Having this on by default will lead to huge log files on our dev machines that we'll have to manually clear up. It should only be switched on when we need low-level debugging.

Shillaker (changes requested):

Shillaker (approved):

PR #59, opened 2021-03-11
Change __wasm__ flag

csegarragonz:
In faasm/faabric#56 I introduced a compiler preprocessor flag whose aim was to choose between two different implementations of the `faabric_win_t` for MPI. I chose `__wasm__`, not understanding that this flag would _only_ be on if run in code that's actually being compiled to WebAssembly.

The problem arose not when running the code natively or compiled to `wasm` but as a call mediating interfacing both.

After discussing off-line, the flag will be removed altogether and faabric will only know about non-wasm stuff. This is complemented by faasm/faasm#393 which implements a `wasm` specific `faabric_win_t` structure for interfacing `wasm-cpp` code.


src/scheduler/MpiWorld.cpp
```diff
@@ -1144,12 +1144,14 @@ void MpiWorld::checkRankOnThisHost(int rank)
     }
 }
 
-void MpiWorld::createWindow(const faabric_win_t* window, uint8_t* windowPtr)
+void MpiWorld::createWindow(const int winRank,
```
csegarragonz, line 1147: This also gets called from `faasm/src/wavm/mpi.cpp` as a consequence it can't depend on `faabric_win_t`. The fact we used the `worldMutex` also made it clone a `faasm`-native version.

Luckily, the implementation is simple enough to workaround not passing the actual struct, just what we need.

Shillaker (approved):

PR #61, opened 2021-03-22
Bulk-scheduling

Shillaker:
This changes scheduling to handle _one to many_ functions, rather than just one at a time.

Key changes:

- Every function chain has an application master. This is the host that executes the first function of that chain, and will act as the scheduler for that whole chain.
- Hosts can now query what resources other hosts have available. This is used by the application master to determine which hosts should execute a function.
- The list of warm hosts for each function is no longer held in Redis, the application master holds this list.
- Hosts no longer have an opinion on whether they should/ shouldn't execute a function. The master for that application will dictate which hosts will execute that function, and they'll just have to work it out
- There is no longer a limit on how many Faaslets can be executing at once, and the concept of the "in flight ratio" has been removed. This removes various scheduling parameters, making everything simpler.


include/faabric/util/config.h
```diff
@@ -40,7 +40,6 @@ class SystemConfig
 
     // Threading
     std::string threadMode;
-    int ompThreadPoolSize;
```
Shillaker, line 43: Not used any more

include/faabric/scheduler/FunctionCallClient.h
```diff
@@ -29,5 +29,11 @@ class FunctionCallClient
     void sendFlush();
 
     void sendMPIMessage(const faabric::MPIMessage& msg);
+
+    faabric::HostResources getResources(const faabric::ResourceRequest& req);
+
+    void executeFunctions(const faabric::BatchExecuteRequest & req);
+
+    void unregister(const faabric::UnregisterRequest& req);
```
Shillaker, line 37: This is to notify the master host that you're no longer "warm"

include/faabric/scheduler/Scheduler.h
```diff
@@ -12,67 +12,61 @@
...
-
-    void notifyNodeFinished(const faabric::Message& msg);
-
-    std::shared_ptr<InMemoryMessageQueue> getBindQueue();
-
-    std::string getFunctionWarmSetName(const faabric::Message& msg);
-
-    std::string getFunctionWarmSetNameFromStr(const std::string& funcStr);
+    std::vector<std::string> callFunctions(faabric::BatchExecuteRequest& req,
+                                           bool forceLocal = false);
```
Shillaker, line 27: This is the entry point into the scheduler, should be called by Faasm

src/scheduler/FunctionCallClient.cpp
```diff
@@ -1,11 +1,92 @@
...
 
 #include <faabric/proto/macros.h>
+#include <faabric/util/queue.h>
+#include <faabric/util/testing.h>
 
 namespace faabric::scheduler {
+
+// -----------------------------------
+// Mocking
+// -----------------------------------
```
Shillaker, line 15: This whole "mocking" section is for testing calls made to other hosts. It's easier than trying to run two servers on the same host.

cmake/ExternalProjects.cmake
```diff
@@ -45,12 +45,12 @@ set_target_properties(pistache
 )
 
 # RapidJSON
+set(RAPIDJSON_BUILD_DOC OFF CACHE INTERNAL "") 
+set(RAPIDJSON_BUILD_EXAMPLES OFF CACHE INTERNAL "") 
+set(RAPIDJSON_BUILD_TESTS OFF CACHE INTERNAL "")
```
Shillaker, line 50: This seems to actually work vs. the previous version which apparently didn't pass the args on

include/faabric/scheduler/FunctionCallClient.h
```diff
@@ -14,6 +14,33 @@ using grpc::ClientContext;
 using grpc::Status;
 
 namespace faabric::scheduler {
+
+// -----------------------------------
+// Mocking
+// -----------------------------------
```
Shillaker, line 20: The following functions are all used/ could be used for testing the client. I found it really difficult to test distributed RPCs without this

include/faabric/scheduler/Scheduler.h
```diff
@@ -9,88 +9,85 @@
 
 #include <shared_mutex>
 
-#define AVAILABLE_HOST_SET "available_nodes"
+#define AVAILABLE_HOST_SET "available_hosts"
```
Shillaker, line 12: The term `node` was really confusing, so I've now used Faaslet or host where relevant

include/faabric/util/func.h
```diff
@@ -61,6 +61,9 @@ std::string buildAsyncResponse(const faabric::Message& msg);
 faabric::Message messageFactory(const std::string& user,
                                 const std::string& function);
 
+faabric::BatchExecuteRequest batchExecFactory(
+  std::vector<faabric::Message>& msgs);
```
Shillaker, line 65: Equivalent of `messageFactory` for batch execution requests.

src/scheduler/FunctionCallClient.cpp
```diff
@@ -1,38 +1,165 @@
...
+// gRPC client
+// -----------------------------------
 FunctionCallClient::FunctionCallClient(const std::string& hostIn)
   : host(hostIn)
   , channel(grpc::CreateChannel(host + ":" + std::to_string(FUNCTION_CALL_PORT),
                                 grpc::InsecureChannelCredentials()))
   , stub(faabric::FunctionRPCService::NewStub(channel))
 {}
 
-void FunctionCallClient::shareFunctionCall(const faabric::Message& call)
```
Shillaker, line 16: `shareFunctionCall` no longer exists, instead the batch execute RPC is always used

src/util/config.cpp
```diff
@@ -40,15 +40,6 @@ void SystemConfig::initialise()
...
     noScheduler = this->getSystemConfIntParam("NO_SCHEDULER", "0");
-    maxNodes = this->getSystemConfIntParam("MAX_NODES", "5");
-    maxInFlightRatio = this->getSystemConfIntParam("MAX_IN_FLIGHT_RATIO", "1");
-    maxNodesPerFunction =
-      this->getSystemConfIntParam("MAX_NODES_PER_FUNCTION", "5");
-
-    // Threading
-    threadMode = getEnvVar("THREAD_MODE", "local");
-    ompThreadPoolSize =
-      this->getSystemConfIntParam("OMP_THREAD_POOL_SIZE", "0");
```
Shillaker, line 51: None of these 5 config parameters are used any more

src/util/json.cpp
```diff
@@ -26,7 +26,6 @@ std::string messageToJson(const faabric::Message& msg)
     d.AddMember("function",
                 Value(msg.function().c_str(), msg.function().size(), a).Move(),
                 a);
-    d.AddMember("hops", msg.hops(), a);
```
Shillaker, line 29: No longer tracking `hops` as the most a message will travel is once

src/scheduler/Scheduler.cpp
```diff
@@ -191,267 +192,271 @@ std::shared_ptr<InMemoryMessageQueue> Scheduler::getBindQueue()
...
+            if (remainder > 0) {
+                std::unordered_set<std::string>& thisRegisteredHosts =
+                  registeredHosts[funcStr];
+
+                std::unordered_set<std::string> allHosts = getAvailableHosts();
+
+                std::unordered_set<std::string> targetHosts;
+
+                // Build list of target hosts
+                // TODO - do this with set difference
```
csegarragonz, line 328: Will we actually do this? I feel that the `TODO` might rot here (differently to those about RPC copies, which are nice to have). To implement the difference I think it should be sufficient to have something like:
```cpp
std::unordered_set<std::string> targetHosts = getAvailableHosts();

for (const auto& h : thisRegisteredHosts) {
    targetHosts.erase(h);
}
```

src/scheduler/Scheduler.cpp
```diff
@@ -191,267 +192,271 @@ std::shared_ptr<InMemoryMessageQueue> Scheduler::getBindQueue()
...
 
-    // Accept if we have capacity
-    const std::string funcStrNoId = faabric::util::funcToString(msg, false);
-    SchedulerOpinion thisOpinion = opinionMap[funcStrNoId];
-    if (thisOpinion == SchedulerOpinion::YES) {
-        return thisHost;
-    }
+    // Execute as many as possible to this host
+    faabric::HostResources r = getHostResources(host);
+    int available = r.cores() - r.functionsinflight();
```
csegarragonz, line 418: Dangling reference to `cores`, could we change to `faaslets`?

src/scheduler/Scheduler.cpp
```diff
@@ -191,267 +192,271 @@ std::shared_ptr<InMemoryMessageQueue> Scheduler::getBindQueue()
...
+                          masterHost);
+
+            FunctionCallClient c(masterHost);
+            c.executeFunctions(req);
+        } else {
+            // At this point we know we're the master host, and we've not been
+            // asked to force full local execution.
+
+            // Work out how many we can handle locally
+            int cores = thisHostResources.cores();
```
csegarragonz, line 257: I thought we were not using the term `cores` anymore?

src/scheduler/Scheduler.cpp
```diff
@@ -470,43 +475,34 @@ Scheduler::getRecordedMessagesShared()
     return recordedMessagesShared;
 }
 
-bool Scheduler::hasHostCapacity()
+void Scheduler::incrementInFlightCount(const faabric::Message& msg)
 {
-    return _hasHostCapacity;
+    const std::string funcStr = faabric::util::funcToString(msg, false);
```
csegarragonz, line 480: Just to double check, this function needs no lock because it only runs in the master right?

tests/test/scheduler/test_function_client_server.cpp
```diff
@@ -182,13 +140,185 @@ TEST_CASE("Test sending flush message", "[scheduler]")
...
+    REQUIRE(sch.getBindQueue()->size() == nCalls);
+}
+
+TEST_CASE("Test get resources request", "[scheduler]")
+{
+    cleanFaabric();
+
+    faabric::scheduler::Scheduler& sch = faabric::scheduler::getScheduler();
+
+    int expectedCores;
```
csegarragonz, line 228: Change cores here

tests/utils/system_utils.cpp
```diff
@@ -29,9 +31,19 @@ void cleanFaabric()
     sch.shutdown();
     sch.addHostToGlobalSet();
 
+    // Give scheduler enough resources
+    faabric::HostResources res;
+    res.set_cores(10);
```
csegarragonz, line 36: Cores

tests/test/scheduler/test_scheduler.cpp
```diff
@@ -13,604 +17,338 @@ namespace tests {
...
 
-    // Check initial set-up
-    REQUIRE(redis.sismember(AVAILABLE_HOST_SET, thisHost));
-    REQUIRE(!redis.sismember(funcSet, thisHost));
+    sch.addHostToGlobalSet(otherHost);
 
-    // Call the function and check it's added to the function's warm set
-    s.callFunction(call);
+    // Set resources
+    int nCores = 5;
```
csegarragonz, line 33: cores

csegarragonz (approved):
Most comments are very minor:
* Dangling references to `cores`
* Locking granularity

Happy to merge it, the scheduling logic looks way simpler!

src/scheduler/Scheduler.cpp
```diff
@@ -164,24 +139,50 @@ std::shared_ptr<InMemoryMessageQueue> Scheduler::getFunctionQueue(
...
+
+    int newInFlight =
+      std::max<int>(thisHostResources.functionsinflight() - 1, 0);
+    thisHostResources.set_functionsinflight(newInFlight);
 }
 
-void Scheduler::notifyNodeFinished(const faabric::Message& msg)
+void Scheduler::notifyFaasletFinished(const faabric::Message& msg)
 {
     faabric::util::FullLock lock(mx);
```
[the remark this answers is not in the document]
Shillaker: Yes exactly, always start with the most granular, and only optimise if it's a performance problem. Even having a single lock can be a source of nasty bugs, so you should resist complexity where possible.

src/scheduler/Scheduler.cpp
Shillaker: Not using the term `nodes` to refer to Faaslets. Cores here really does mean processor cores

src/scheduler/Scheduler.cpp
Shillaker: Yes good point, I'll get rid of it. What I actually meant was [`std::set_difference`](https://en.cppreference.com/w/cpp/algorithm/set_difference), but it's a bit difficult to read and probably not worth it.

src/scheduler/Scheduler.cpp
Shillaker: See comment above, `cores` are processor cores, `nodes` have become `faaslets`.

src/scheduler/Scheduler.cpp
Shillaker: This function should need no lock as it's only ever called from functions that already have a lock, and it's `private`. There may be two instances of the same function executing on the same host, so locking does still matter.

tests/test/scheduler/test_function_client_server.cpp
Shillaker: See above

src/scheduler/Scheduler.cpp
csegarragonz: Well, afaict `std::set_difference` only works on [ordered sets](https://en.cppreference.com/w/cpp/algorithm/set_difference), that's why I suggested this other method.

src/scheduler/Scheduler.cpp
csegarragonz: I was confused

PR #63, opened 2021-03-29
Fix in the `MPI_MIN` operator when running a reduce

csegarragonz:
When runing a `reduce` with the `MPI_MIN` operator, if `sendRank == recvRank` yet `not isInPlace`, we did (simplified):
```cpp
// rankData  stores the received data, if sendRank != recvRank we'd call recv()
uint8_t* rankData = sendBuffer;
memset(recvBuffer, 0, bufferSize);

op_reduce(FAABRIC_OP_MIN, recvBuffer, rankData);
```

As `recvBuffer` is filled with `0`s, these will override the minimum, obtaining incorrect results. I extend the tests to consider this case as well.

Related to faasm/experiment-lammps#8

### DIFF
--- a/src/scheduler/MpiWorld.cpp
+++ b/src/scheduler/MpiWorld.cpp
@@ -674,7 +674,9 @@ void MpiWorld::reduce(int sendRank,
         // like the minimum, or product.
         // If we're receiving from ourselves and in-place, our work is
         // already done and the results are written in the recv buffer
-        memcpy(recvBuffer, sendBuffer, bufferSize);
+        if (!isInPlace) {
+            memcpy(recvBuffer, sendBuffer, bufferSize);
+        }
 
         uint8_t* rankData = new uint8_t[bufferSize];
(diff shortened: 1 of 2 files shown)


src/scheduler/MpiWorld.cpp
```diff
@@ -666,42 +666,36 @@ void MpiWorld::reduce(int sendRank,
...
+        // like the minimum, or product.
+        // If we're receiving from ourselves and in-place, our work is
+        // already done and the results are written in the recv buffer
         if (!isInPlace) {
-            memset(recvBuffer, 0, bufferSize);
+            memcpy(recvBuffer, sendBuffer, bufferSize);
         }
 
+        uint8_t* rankData = new uint8_t[bufferSize];
         for (int r = 0; r < size; r++) {
```
csegarragonz, line 681: re-shuffling the conditionals reduced a bit the tests we had to check

Shillaker (approved):

PR #64, opened 2021-03-30
Fix cartesian topology implementation

csegarragonz:
In this PR I correct some erroneous behaviour I introduced when implementing the cartesian topology methods in MPI. In particular I correct (and clarify) the implementation of `MPI_Cart_get` and provide an implementation for `MPI_Cart_shift`. Note that in this PR there's only the `faabric`-related changes. A companion one for `faasm` will be ready once LAMMPS execution is flushed out. I also improve the tests for this functionality.

`MPI_Cart_get` calls `MpiWorld::getCartesianRank` which, given a triple of dimensions `i x j x 1` (3D-grid), returns the coordinates in said grid of the caller rank. The grid is filled left to right, top to bottom (in this order).

`MPI_Cart_shift` calls `MpiWorld::shiftCartesianCoords` and returns, for the caller rank:
* The rank it would land if moving `disp` units along a `direction`.
* The rank that would land on it by moving `disp` units along a `direction`.

It's a bit of a tong-twister, but hopefuly the comments help understanding the code. It's easy to make mistakes with symbols, so I've taken care to add a couple of test cases, and have it be easily extensible (to include more test cases were it to be necessary).

Related to faasm/experiment-lammps#8

### DIFF
--- a/include/faabric/scheduler/MpiWorld.h
+++ b/include/faabric/scheduler/MpiWorld.h
@@ -52,6 +52,12 @@ class MpiWorld
 
     void getRankFromCoords(int* rank, int* coords);
 
+    void shiftCartesianCoords(int rank,
+                              int direction,
+                              int disp,
+                              int* source,
+                              int* destination);
+
     void send(int sendRank,
               int recvRank,
               const uint8_t* buffer,
--- a/include/faabric/scheduler/MpiWorld.h
+++ b/include/faabric/scheduler/MpiWorld.h
@@ -44,7 +44,11 @@ class MpiWorld
 
     void enqueueMessage(faabric::MPIMessage& msg);
 
-    void getCartesianRank(int rank, int* dims, int* periods, int* coords);
+    void getCartesianRank(int rank,
+                          int maxDims,
+                          const int* dims,
+                          int* periods,
+                          int* coords);
 
     void getRankFromCoords(int* rank, int* coords);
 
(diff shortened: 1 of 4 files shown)


include/faabric/scheduler/MpiWorld.h
```diff
@@ -214,6 +224,8 @@ class MpiWorld
       localQueueMap;
     std::unordered_map<int, std::thread> asyncThreadMap;
 
+    std::vector<int> cartProcsPerDim;
```
csegarragonz, line 227: This is a bit of state that would get embedded in the communicator, but if chosen to keep the communicator simple and store it here. It stores how many MPI processes are there along each dimension (i.e. in the `i x j x 1` grid).

It is safe to do so as all processes are invoked with the same parameters.

tests/test/scheduler/test_mpi_world.cpp
```diff
@@ -112,48 +112,94 @@ TEST_CASE("Test cartesian communicator", "[mpi]")
     cleanFaabric();
 
     faabric::Message msg = faabric::util::messageFactory(user, func);
-    // 5 processes create a 2x2 grid with one left MPI_UNDEFINED
```
csegarragonz, line 115: This was false in LAMMPS. In particular, my wrong assumption was that MPI inferred the size of the grid given the `worldSize`. However, I realized that the programmer provides it.

tests/test/scheduler/test_mpi_world.cpp
```diff
@@ -691,8 +737,9 @@ TEST_CASE("Test collective messaging locally and across hosts", "[mpi]")
 
         // Check the host that the root is on
         for (int rank : remoteWorldRanks) {
-            if (rank == remoteRankB)
+            if (rank == remoteRankB) {
```
csegarragonz, line 740: I am surprised that this wasn't brought up before by `clang-tidy` I'll note down to look into it.

PR #67, opened 2021-04-01
MPI Async Fixes

csegarragonz:
This PR solves a bug in the asynchronous API. As always it is a one-liner that ends up making it work, but I've polished some things in the process.

The main problem was that a shared data structure `asyncThreadMap` was concurrently edited by different threads. However, there really was no need to, as each thread R/W to different keys. Making the structure `thread_local` solved the issue.

Further, I've removed the `doISendRecv` method and provided separate implementations for `irecv` and `isend`, plus some small optimizations on the map usage.

Now LAMMPS' execution finishes. It does not scale at all and still sometimes randomly fails, but its quite robust and runs OK most of the times.

Closes faasm/experiment-lammps#8

### DIFF
--- a/include/faabric/scheduler/MpiWorld.h
+++ b/include/faabric/scheduler/MpiWorld.h
@@ -231,15 +231,6 @@ class MpiWorld
 
     void checkRankOnThisHost(int rank);
 
-    int doISendRecv(int sendRank,
-                    int recvRank,
-                    const uint8_t* sendBuffer,
-                    uint8_t* recvBuffer,
-                    faabric_datatype_t* dataType,
-                    int count,
-                    faabric::MPIMessage::MPIMessageType messageType =
-                      faabric::MPIMessage::NORMAL);
-
     void pushToState();
 };
 }
--- a/include/faabric/scheduler/MpiWorld.h
+++ b/include/faabric/scheduler/MpiWorld.h
@@ -222,7 +222,6 @@ class MpiWorld
 
     std::unordered_map<std::string, std::shared_ptr<InMemoryMpiQueue>>
       localQueueMap;
-    std::unordered_map<int, std::thread> asyncThreadMap;
 
     std::vector<int> cartProcsPerDim;
 
--- a/include/faabric/scheduler/MpiWorld.h
+++ b/include/faabric/scheduler/MpiWorld.h
@@ -98,7 +98,7 @@ class MpiWorld
               faabric::MPIMessage::MPIMessageType messageType =
                 faabric::MPIMessage::NORMAL);
 
-    void awaitAsyncRequest(int requestId);
+    void awaitAsyncRequest(int requestId, int myRank);
 
     void sendRecv(uint8_t* sendBuffer,
                   int sendcount,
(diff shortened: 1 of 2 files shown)


src/scheduler/MpiWorld.cpp
```diff
@@ -403,6 +389,55 @@ void MpiWorld::send(int sendRank,
     }
 }
 
+void MpiWorld::recv(int sendRank,
```
csegarragonz, line 392: Moved `recv` here to have `isend`, `irecv`, `send`, and `recv`'s implementations together, noisy diff however.

src/scheduler/MpiWorld.cpp
```diff
@@ -9,6 +9,8 @@
 #include <faabric/util/macros.h>
 #include <faabric/util/timing.h>
 
+static thread_local std::unordered_map<int, std::thread> asyncThreadMap;
```
csegarragonz, line 12: This line here basically solved the problem.

include/faabric/scheduler/MpiWorld.h
```diff
@@ -98,7 +98,7 @@ class MpiWorld
               faabric::MPIMessage::MPIMessageType messageType =
                 faabric::MPIMessage::NORMAL);
 
-    void awaitAsyncRequest(int requestId);
+    void awaitAsyncRequest(int requestId, int myRank);
```
csegarragonz, line 101: This is cheap, and helps with the logs. I can revert it though.

src/scheduler/MpiWorld.cpp
```diff
@@ -305,46 +307,30 @@ int MpiWorld::isend(int sendRank,
                     int count,
                     faabric::MPIMessage::MPIMessageType messageType)
 {
-    return doISendRecv(
-      sendRank, recvRank, buffer, nullptr, dataType, count, messageType);
+    int requestId = (int)faabric::util::generateGid();
```
csegarragonz, line 310: As discussed, removed the `doISendRecv` method and implemented separate `irecv` and `isend`.

src/scheduler/MpiWorld.cpp
```diff
@@ -305,46 +307,40 @@ int MpiWorld::isend(int sendRank,
...
+
+    asyncThreadMap.emplace(
```
csegarragonz, line 312: The previous version used `insert` instead of `emplace` both will silently fail if the key already exists, but the `[]` operator would silently override the value if the key was repeated.

To be extra cautious we'd check that the key is indeed not there, but I think trusting `util::generateGid()` is a safe assumption (for the moment).

src/scheduler/MpiWorld.cpp
```diff
@@ -648,81 +693,29 @@ void MpiWorld::allGather(int rank,
...
-
 void MpiWorld::awaitAsyncRequest(int requestId)
 {
     faabric::util::getLogger()->trace("MPI - await {}", requestId);
 
-    if (asyncThreadMap.count(requestId) == 0) {
-        throw std::runtime_error(
-          "Attempting to await unrecognised async request: " +
-          std::to_string(requestId));
+    auto it = asyncThreadMap.find(requestId);
```
csegarragonz, line 700: By using `find` instead of `count` we perform one lookup instead of two.

Shillaker (approved):
Awesome, I think we all learned something today :joy:

PR #68, opened 2021-04-01
Compose problems

csegarragonz:
The order is clearly specified [here](https://docs.docker.com/compose/env-file/):

> The project directory is specified by the order of precedence:
> * --project-directory flag
> * Folder of the first --file flag
> * Current directory

But has apparently been reverted [here](https://docs.docker.com/compose/release-notes/#1286), dated 23/3. Now:

> Bug fixes
> * Made --env-file relative to the current working directory. Environment file paths set with --env-file are now relative to the current working directory and override the default .env file located in the project directory.

This means, that it does not take into account the directory of the `--file` flag (so we must revert #58).

I add a check of the `docker-compose` version for future reference.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -125,6 +125,8 @@ jobs:
       redis:
         image: redis
     steps:
+      - name: "Compose version"
+        run: docker-compose --version
       # --- Code update ---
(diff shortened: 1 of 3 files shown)


Shillaker (approved):
Grim, good catch

PR #74, opened 2021-04-09
Thread Pool for MPI

csegarragonz:
In this PR I introduce a thread pool for asynchronous MPI work. Instead of creating a new thread for every different asynchronous task, we now have a fixed size thread pool per-MpiWorld (i.e. faabric instance/host). This alone, without any further optimization, reduces the runtime by over 25% (40s -> 27s when running LAMMPS with 2 processes). Unfortunately, it still does not scale and is far from native.

Further, profiling shows that no longer the `native_msr_write` kernel system call dominates _all_ the flame graph. Here are the [before](https://github.com/faasm/faabric/files/6289381/2-proc.zip) and [after](https://github.com/faasm/faabric/files/6289377/2-proc.zip) files (be sure to open them in a browser).

The pending work to do could be classified in two groups:
* Reducing the time spent in `cond_wait` (this could mean having more `cv`s or queues). Observe that it is not spent _actually_ waiting, but rather context switching/actually doing kerenl-side work.
* Reducing the copying, which becomes a problem as we increase the number of processes.


src/flat/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ add_custom_command(
     DEPENDS faabric.fbs
     WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
     COMMAND ${FLATBUFFERS_FLATC_EXECUTABLE}
-    ARGS --grpc --cpp faabric.fbs
+    ARGS --grpc --cpp --raw-binary faabric.fbs
```
csegarragonz, line 17: I am pretty sure this does no harm and allowed to compile outside the container. I am happy to unstage the changes though.

src/scheduler/MpiThreadPool.cpp
```diff
@@ -0,0 +1,98 @@
...
+    // dequeue timeout and won't be joinable, or none has and all are joinable.
+    // We wait untill all hit the timeout, saving an additional cond. variable.
+    for (auto& thread : threadPool) {
+        if (thread.joinable()) {
+            thread.join();
+        }
+    }
+}
+
+void MpiAsyncThreadPool::awaitAsyncRequest(int reqId)
```
csegarragonz, line 37: This implements the logic that before would have been in `faabric::awaitAsyncRequest`. I've kept both as I like the indirection: `faasm->faabric->threadPool_impl`

src/scheduler/MpiThreadPool.cpp
```diff
@@ -0,0 +1,98 @@
...
+{
+    faabric::util::UniqueLock lock(awakeMutex);
+
+    // Wait until the condition is signaled and the predicate is false.
+    // Note that the same lock protects concurrent accesses to asyncReqMap
+    awakeCV.wait(lock, [this, reqId] {
+        return this->asyncReqMap.find(reqId) != this->asyncReqMap.end();
```
csegarragonz, line 44: I wasn't aware of the predicate option in [`std::wait`](https://en.cppreference.com/w/cpp/thread/condition_variable/wait). Unless proven inefficient, it simplifies a lot the notify-all but wake-one logic.

include/faabric/scheduler/MpiThreadPool.h
```diff
@@ -0,0 +1,48 @@
...
+
+  protected:
+    void entrypoint(int i);
+
+  private:
+    std::vector<std::thread> threadPool;
+    std::mutex awakeMutex;
+    std::condition_variable awakeCV;
+
+    // TODO - use instead a data structure that is very easy to lookup. I am
```
csegarragonz, line 39: For the moment I've used a map (inherited from previous impl.) where:
* If `reqId` is a key it means that `reqId` has finished.
* If `reqId` is not a key, we must still wait.
Accesses to the map are protected by the `wait/notify` lock.

A map is not the right data structure for this as we don't use the value at all, and I may change this in the future. But this is not the current bottleneck.

src/scheduler/MpiThreadPool.cpp
```diff
@@ -0,0 +1,98 @@
...
+        return this->finishedReqs.find(reqId) != this->finishedReqs.end();
+    });
+
+    // Before giving up the lock, remove our request as it has already finished
+    auto it = finishedReqs.find(reqId);
+    if (it == finishedReqs.end()) {
+        throw std::runtime_error(
+          fmt::format("Error: unrecognized reqId {}", reqId));
+    }
+    finishedReqs.erase(it);
```
Shillaker, line 53: How come we can't just call `finishedReqs.erase(reqId)`?

Shillaker (approved):
Awesome, LGTM :+1:

src/scheduler/MpiThreadPool.cpp
csegarragonz: I double-check the iterator (for sanity) and given that I already have it, it's faster to erase from the iterator (as we don't have to look it up again).

PR #75, opened 2021-04-11
Minimize copies

csegarragonz:


include/faabric/util/queue.h
```diff
@@ -30,7 +30,7 @@ class Queue
     {
         UniqueLock lock(mx);
 
-        mq.push(value);
+        mq.emplace(std::move(value));
```
csegarragonz, line 33: AFAICT the new line will do the same as the first one in the worst case, but can be optimized if we pass it a certain type of value. This is necessary to have a queue of `unique_ptr`s.

include/faabric/util/queue.h
```diff
@@ -53,7 +53,7 @@ class Queue
             }
         }
 
-        T value = mq.front();
+        T value = std::move(mq.front());
```
csegarragonz, line 56: See previous comment.

src/scheduler/MpiWorld.cpp
```diff
@@ -11,16 +11,12 @@
 #include <faabric/util/macros.h>
 #include <faabric/util/timing.h>
 
-static thread_local std::unordered_map<int, std::thread> asyncThreadMap;
```
csegarragonz, line 14: This was not used anymore.

src/scheduler/MpiWorld.cpp
```diff
@@ -11,16 +11,12 @@
...
-
 namespace faabric::scheduler {
 MpiWorld::MpiWorld()
   : id(-1)
   , size(-1)
   , thisHost(faabric::util::getSystemConfig().endpointHost)
   , creationTime(faabric::util::startTimer())
-  , threadPool(std::make_shared<faabric::scheduler::MpiAsyncThreadPool>(
```
csegarragonz, line 22: Playing around, I realized that having more or less one worker thread per rank reduces a lot the overhead in managing threads. See the attached flame graphs in the PR description.

include/faabric/scheduler/MpiWorld.h
```diff
@@ -9,7 +9,7 @@
 #include <thread>
 
 namespace faabric::scheduler {
-typedef faabric::util::Queue<faabric::MPIMessage> InMemoryMpiQueue;
+typedef faabric::util::Queue<std::unique_ptr<faabric::MPIMessage>> InMemoryMpiQueue;
```
csegarragonz, line 12: This is not a game-changer, but an easy win and nicer to prevent memory leaks.

include/faabric/scheduler/MpiWorld.h
```diff
@@ -9,7 +9,8 @@
 #include <thread>
 
 namespace faabric::scheduler {
-typedef faabric::util::Queue<faabric::MPIMessage> InMemoryMpiQueue;
+typedef faabric::util::Queue<std::shared_ptr<faabric::MPIMessage>>
```
csegarragonz, line 12: Originally these were `std::unique_ptr` (which I preferred). This works generally allright except when trying to `peek` the queue. In that case we need a reference from the first element in the queue, but want to keep another reference in the queue. This particular use-case only made it necessary to switch to `shared_ptr`.

src/scheduler/MpiWorld.cpp
```diff
@@ -369,39 +367,39 @@ void MpiWorld::send(int sendRank,
...
-            getLocalQueue(sendRank, recvRank)->enqueue(m);
+            getLocalQueue(sendRank, recvRank)->enqueue(std::move(m));
         }
     } else {
         logger->trace("MPI - send remote {} -> {}", sendRank, recvRank);
 
         // TODO - avoid creating a client each time?
         scheduler::FunctionCallClient client(otherHost);
-        client.sendMPIMessage(m);
+        client.sendMPIMessage(*m);
```
Shillaker, line 402: Can we avoid a copy here too? If we change `sendMPIMessage` to take a `shared_ptr` then only dereference it at the last minute

src/scheduler/MpiWorld.cpp
```diff
@@ -1005,14 +1004,15 @@ void MpiWorld::allToAll(int rank,
 
 void MpiWorld::probe(int sendRank, int recvRank, MPI_Status* status)
 {
+    // TODO - probe is now broken as we need to re-implement queue's peek
```
Shillaker, line 1007: If this TODO is still true we should throw an exception and delete/ comment out the broken implementation

Shillaker (changes requested):
Very nice, just a couple of small points, potentially avoiding another copy. We should really switch this to flatbuffers, but this is definitely sufficient for now.

include/faabric/util/queue.h
Shillaker: Should we have an associated `std::move` around `doDequeue` in the `dequeue` method?

src/scheduler/MpiWorld.cpp
csegarragonz: I forgot to delete this comment, thanks

Shillaker:
@csegarragonz I just pushed some changes here, see Slack

include/faabric/util/queue.h
```diff
@@ -53,18 +53,31 @@ class Queue
...
-        }
+        T value = std::move(mq.front());
+        mq.pop();
+        emptyNotifier.notify_one();
 
         return value;
     }
 
-    T peek(long timeoutMs = 0) { return doDequeue(timeoutMs, false); }
+    T* peek(long timeoutMs = 0)
```
Shillaker, line 63: To make peek work when  the values in the queue aren't smart pointers, we need to return a pointer here to avoid copying.

csegarragonz: Nice

Shillaker (approved):

PR #76, opened 2021-04-12
Use promises

csegarragonz:
In this PR I update the thread pool introduced in #74 to:
* Properly shutdown (i.e. not with a timeout, rather with a shutdown message).
* Use promises instead of an additional data structure with the finished request. Types are longer, but complexity smaller.


include/faabric/scheduler/MpiThreadPool.h
```diff
@@ -11,8 +12,10 @@
 #include <faabric/util/queue.h>
 
 namespace faabric::scheduler {
-typedef faabric::util::Queue<std::pair<int, std::function<void(void)>>>
-  MpiReqQueue;
+// < ReqId, <func(), promise>>
+typedef std::pair<int, std::pair<std::function<void(void)>, std::promise<void>>>
```
csegarragonz, line 16: This is getting out of hand

Shillaker (changes requested):
I think `std::tuple` might make things a bit tidier, but I may be wrong

Shillaker (approved):

PR #78, opened 2021-04-14
Running multi-host MPI applications

csegarragonz:
This PR introduces the minimal changes needed to run multi-host MPI applications through the REST API. I also include a new config parameter that allows overriding the CPU count to fake single-machine multi-host docker deployments.

I also use this PR as a clean sheet to start GRPC-related optimizations.

### DIFF
--- a/src/scheduler/MpiWorld.cpp
+++ b/src/scheduler/MpiWorld.cpp
@@ -66,14 +66,15 @@ std::shared_ptr<state::StateKeyValue> MpiWorld::getRankHostState(int rank)
 
 int MpiWorld::getMpiThreadPoolSize()
 {
+    auto logger = faabric::util::getLogger();
     int usableCores = faabric::util::getUsableCores();
     int worldSize = size;
 
-    // Note - this will overprovision threads if worldSize > usableCores, and
-    // worldSize % usableCores != 0
-    faabric::util::getLogger()->warn(
-      "Over-provisioning threads in the MPI thread pool. To avoid this, set a "
-      "MPI world size multiple of the number of cores per machine.");
+    if ((worldSize > usableCores) && (worldSize % usableCores != 0)) {
+        logger->warn("Over-provisioning threads in the MPI thread pool.");
+        logger->warn("To avoid this, set an MPI world size multiple of the "
+                     "number of cores per machine.");
+    }
     return std::min<int>(worldSize, usableCores);
 }
 
--- a/src/scheduler/MpiWorld.cpp
+++ b/src/scheduler/MpiWorld.cpp
@@ -66,26 +66,12 @@ std::shared_ptr<state::StateKeyValue> MpiWorld::getRankHostState(int rank)
 
 int MpiWorld::getMpiThreadPoolSize()
 {
-    // We initialise as many threads as MPI ranks assigned to this host, as long
-    // as this number does not exceed the number of usable cores.
     int usableCores = faabric::util::getUsableCores();
-    int ranksInHost = size;
-    /* TODO improve
-    for (int i = 0; i < this->size; i++) {
-        if (getHostForRank(i) == thisHost) {
-            ranksInHost++;
-        }
-    }
-    */
+    int worldSize = size;
 
-    // TODO - does it make sense to pin each thread to a CPU core using affinity
-    if (ranksInHost > usableCores) {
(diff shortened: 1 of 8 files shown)


src/scheduler/MpiWorld.cpp
```diff
@@ -64,14 +64,25 @@ std::shared_ptr<state::StateKeyValue> MpiWorld::getRankHostState(int rank)
...
 }
 
+int MpiWorld::getMpiThreadPoolSize()
+{
+    int usableCores = faabric::util::getUsableCores();
+    int worldSize = size;
+
+    // Note - this may overprovision threads if worldSize > usableCores, and
+    // worldSize % usableCores != 0
+    return std::min<int>(worldSize, usableCores);
```
Shillaker, line 74: Could we log a warning here if it is overprovisioning? Sometimes it's good to grep the logs for warnings to see what might be going wrong.

src/util/environment.cpp
```diff
@@ -35,7 +36,14 @@ void unsetEnvVar(const std::string& varName)
...
 {
-    unsigned int nCores = std::thread::hardware_concurrency();
+    auto conf = faabric::util::getSystemConfig();
+    unsigned int nCores;
+
+    if (conf.overrideCpuCount == 0) {
+        nCores = std::thread::hardware_concurrency();
+    } else {
+        nCores = conf.overrideCpuCount;
+    }
```
Shillaker, line 46: Can you add a test for this? Mainly to check that overriding returns the overridden value rather than the default.

src/util/config.cpp
```diff
@@ -114,6 +115,7 @@ void SystemConfig::print()
 
     logger->info("--- Scheduling ---");
     logger->info("NO_SCHEDULER               {}", noScheduler);
+    logger->info("OVERRIDE_CPU_COUNT         {}", overrideCpuCount);
```
Shillaker, line 118: Could you add this field to the config test?

Shillaker (changes requested):

src/util/environment.cpp
csegarragonz: I completely forgot about `/util` tests. Sorry about that.

tests/test/util/test_config.cpp
```diff
@@ -58,6 +59,7 @@ TEST_CASE("Test overriding system config initialisation", "[util]")
     std::string redisPort = setEnvVar("REDIS_PORT", "1234");
 
     std::string noScheduler = setEnvVar("NO_SCHEDULER", "1");
+    std::string overrideCpuCount = setEnvVar("OVERRIDE_CPU_COUNT", "4");
```
Shillaker, line 62: Gah, sorry about this but you're missing the resetting of this env var at the bottom, so all tests after this will have this set to 4

tests/test/util/test_environment.cpp
```diff
@@ -35,4 +37,16 @@ TEST_CASE("Test setting environment variables", "[util]")
...
+TEST_CASE("Test overriding CPU count", "[util]")
+{
+    // Unset just in case it is set
+    unsetEnvVar("OVERRIDE_CPU_COUNT");
+
+    unsigned int expectedNumCores = std::thread::hardware_concurrency();
+    REQUIRE(getUsableCores() == expectedNumCores);
+
+    getSystemConfig().overrideCpuCount = 4;
+    REQUIRE(getUsableCores() == 4);
```
Shillaker, line 50: This again will leave the modifications hanging around for any tests that don't call `cleanFaabric`, before running. I'd recommend this instead:

```
cleanFaabric();

// Check default cores is same as available concurrency
auto &conf = getSystemConfig();
int defaultCores = getUsableCores();
REQUIRE(defaultCores == std::thread::hardware_concurrency());

// Check it can be overridden (using a number that's obviously always wrong)
conf.overrideCpuCount = 1234;
REQUIRE(getUsableCores() == 1234);

// Reset the conf
conf.reset();

// Check we're back
REQUIRE(getUsableCores() == defaultCores);
```

Shillaker (changes requested):

tests/test/util/test_config.cpp
csegarragonz: Sure, no problem. I am very sloppy today apparently :sweat_smile:

Shillaker (approved):

PR #79, opened 2021-04-16
gRPC asynchronous client-server

csegarragonz:
UPDATE: I mark this PR as draft until we decide on the adequate underlying transport for MPI messaging.

In this PR I introduce an asynchronous client-server pair for the inter-host MPI messaging layer. The current version runs a single-threaded server, and is very much inspired on [gRPC's tutorial](https://grpc.io/docs/languages/cpp/async/).

Preliminary results show a decent improvement as we increase the volume of inter-host communication.

I am already working in the multi-threaded async server, which should be included in #84. The version there changes quite a bit the server-side logic, but I prefer to merge them separately, as this one is tested and working fine.


src/scheduler/MpiWorld.cpp
```diff
@@ -425,11 +452,14 @@ void MpiWorld::send(int sendRank,
             getLocalQueue(sendRank, recvRank)->enqueue(std::move(m));
         }
     } else {
-        logger->trace("MPI - send remote {} -> {}", sendRank, recvRank);
+        logger->debug("MPI - send remote {} -> {}", sendRank, recvRank);
 
-        // TODO - avoid creating a client each time?
-        scheduler::FunctionCallClient client(otherHost);
-        client.sendMPIMessage(m);
+        if (messageType == faabric::MPIMessage::RMA_WRITE) {
```
csegarragonz, line 457: RMA exposes a purely synchronous API, whereas for other MPI calls, only active things like `MPI_Recv` will block. With the new asyncrhonous client-server, the way to replicate purely synchronous behaviour is by managing the lifecycle of the client: it's destructor acts as synchronisation point.

tests/test/scheduler/test_mpi_world.cpp
```diff
@@ -471,16 +471,6 @@ TEST_CASE("Test send across hosts", "[mpi]")
     remoteWorld.send(
       rankA, rankB, BYTES(messageData.data()), MPI_INT, messageData.size());
 
-    SECTION("Check queueing")
```
csegarragonz, line 474: Issuing a `faabric::MpiWorld::send` does no longer wait for completion. Thus, checking the queues right after we send a message will most likely fail.

The workaround is to force a purely-synchronous message. This behaviour, together with a test for correct enqueueing, is already provided in `[Test sending MPI message] [scheduler]`. 

I can also workaround a test for the queues in this file, but it wouldn't use the `MpiWorld` 's API.

include/faabric/scheduler/MpiWorld.h
```diff
@@ -224,15 +225,23 @@ class MpiWorld
 
     std::unordered_map<std::string, std::shared_ptr<InMemoryMpiQueue>>
       localQueueMap;
+
+    // Thread pool for asynchronous work
     std::shared_ptr<faabric::scheduler::MpiAsyncThreadPool> threadPool;
+    int getMpiThreadPoolSize();
```
Shillaker, line 231: The big question is, can we get rid of the MPI thread pool once we have this stuff in place?

src/scheduler/MpiWorld.cpp
Shillaker: Could we somehow abstract this by having a function called `sendSynchronousMPIMessage` that contains these two lines and a comment explaining why? That way we can change it in future.

I think perhaps we would create a separate RPC just for handling MPI RMA.

src/scheduler/MpiWorld.cpp
```diff
@@ -448,6 +477,9 @@ void MpiWorld::recv(int sendRank,
     std::shared_ptr<faabric::MPIMessage> m =
       getLocalQueue(sendRank, recvRank)->dequeue();
 
+    if (!m) {
+        throw std::runtime_error("Recevied a malformed message");
+    }
```
Shillaker, line 482: Does this ever happen? The message in this case isn't malformed it's null. I would just lose this if it never happens, or add a more detailed error log line if it does (e.g. Recv message 1 -> 2 is null)

include/faabric/rpc/RPCServer.h
```diff
@@ -27,5 +27,7 @@ class RPCServer
     std::thread servingThread;
 
     virtual void doStart(const std::string& serverAddr) = 0;
+
+    virtual void doStop() = 0;
```
Shillaker, line 31: It seems that every subclass calls `server->Shutdown()`, so rather than making this a pure virtual function, I would just make it virtual, provide a default implementation which calls `server->Shutdown()`, then override in the async one.

Shillaker (commented):
Couple of very minor comments, feel free to implement them here or in the next PR

csegarragonz:
Closing as #85 has been merged.

PR #81, opened 2021-04-21
Introduce delta snapshot utilities and zstd

kubasz:
This introduces a zstd dependency and adds configurable delta snapshots with multiple encodings.

The different encoding formats (with/without pages, compressed/uncompressed, xor/overwrite mode) are tested to work, but I have not determined their individual overheads and benefits so far.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13.0)
+cmake_minimum_required(VERSION 3.14.0)
 project(faabric)
 
 option(FAABRIC_WASM_BUILD "Build Faabric wasm library" OFF)
@@ -28,7 +28,7 @@ include(cmake/ExternalProjects.cmake)
 function(faabric_lib lib_name lib_deps)
     # "Normal" library used for linking internally
     add_library(${lib_name} ${lib_deps})
-    target_include_directories(${lib_name} 
+    target_include_directories(${lib_name}
         PUBLIC ${FAABRIC_INCLUDE_DIR}
         PUBLIC ${PISTACHE_INCLUDE_DIR}
         PUBLIC ${SPDLOG_INCLUDE_DIR}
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ include(cmake/ExternalProjects.cmake)
 function(faabric_lib lib_name lib_deps)
     # "Normal" library used for linking internally
     add_library(${lib_name} ${lib_deps})
-    target_include_directories(${lib_name}
```
csegarragonz, line 31: `clang-format` removes trailing whitespaces, but we have no linting for `CMake` files. Thus the unpleasant and misleading `diff`. I agree though that removing trailing whitespaces is nice. Should we "unofficially" enforce this ourselves @Shillaker ?

include/faabric/util/config.h
```diff
@@ -26,6 +26,7 @@ class SystemConfig
     std::string captureStdout;
     std::string stateMode;
     std::string wasmVm;
+    std::string deltaSnapshotEncoding;
```
csegarragonz, line 29: Could you add a very simple test for setting and unsetting this value in `tests/test/util/test_config.cpp`?

csegarragonz (commented):
Hey @kubasz , thanks for this! I am leaving the bulk of the review for Simon, but just pointed out one minor thing as I skimmed through the code.

CMakeLists.txt
Shillaker: Yes I think it's great that Jakub is cleaning up all this unnecessary whitespace :joy: . We'll have to do a big clear-up ourselves but this can definitely stay as a start.

include/faabric/util/config.h
Shillaker: Yes that's the only thing I'd add, but I'll be doing some work on this so happy to add it myself. Let's approve this and I'll add that part.

Shillaker (approved):

Shillaker:
LGTM :+1:

PR #82, opened 2021-04-22
gRPC Release Build + Bump Version

csegarragonz:
In this PR I change the build type for `gRPC` to make sure we use the `Release` one.


Shillaker (approved):
Nice, might be worth grepping for anywhere else we call CMake to ensure we're doing a release build there too.

Bigger question, do dependencies automatically pick up our `CMAKE_BUILD_TYPE` when they're added through `FetchContent` etc.?

PR #83, opened 2021-04-22
Unifying threads, functions and thread pooling

### LINKED ISSUE #65: Potential scheduler bug? nextMsgIdx not updated when scheduling to other hosts
I'm adapting my compute/storage division-aware scheduler patches for faasm-ndp to the new code, and I noticed that in the blocks of code responsible for scheduling off to other nodes when the current one is busy, the `nextMsgIndex` used for calculations later is not updated - it was updated in the for loop stepping above, maybe it was missed in a copy-paste?

https://github.com/faasm/faabric/blob/master/src/scheduler/Scheduler.cpp#L287-L391

```cpp
for (; nextMsgIdx < nLocally; nextMsgIdx++) { // for loop - nextMsgIdx updated
...
remainder -= nOnThisHost; // but nextMsgIdx not updated
```

From a code review standpoint it seems that there are multiple variables manually tracked: `nextMsgIdx`, `remainder` (which should just always be total minus nextMsgIdx), and the actual messages vector. The vector is not mutated, so that's not a problem, but I think replacing either `nextMsgIdx` or `remainder` with a formula based on the other one (you could put it in a capture-by-reference local lambda to avoid duplication) would eliminate this bug and also prevent making similar mistakes. Also marking locals like `nMessages`, `isThreads`, `cores` as `const` would make it clear these are not changing in the function.

```cpp
const auto remainder = [&](){ return nMessages - nextMsgIdx; };
if (remainder() > 0) {...}
```
or
```cpp
if (nextMsgIdx < nMessages) {...} // without remainder
```

Shillaker:
The purpose of this change is to unify the treatment of threads and functions into a single simple executor interface. This is necessary to support remote threading and scheduling of distributed threads, making them first class citizens alongside functions (in terms of scheduling decisions).

The key changes are:

- The scheduler now controls the creation and destruction of executors (rather than having a queue between the scheduler and a pool of executors)
- Rather than running each executor in its own thread, executors will lazily create their own worker threads when needed. For a simple function, this means the executor will create a single worker thread to execute that function. If that function then spawns more threads, the executor's thread pool will grow.
- To simplify the logic, there is no longer the concept of cold and warm executors. Executors only exist bound to a function or don't exist. This removes an extra layer of messaging and queueing.
- The interface required from `Executor` subclasses is much simpler, with each only needing to implement `executeTask`, which takes a bulk request and a set of indexes saying which messages from that bulk request it needs to execute. For a simple function this request will hold one message and the list of indexes will be `{0}`.
- Protobuf messages involved in function execution are not copied unless absolutely necessary (when receiving in the function call server).
- Faabric tests now include more testing of the `Executor` subclasses, including error handling, how they execute functions and threads etc. 
- I've remove the use of the word "faaslet" wherever I found it, as it's Faasm-specific. Wherever it was used is now "executor".
- Closes #65 by simplifying the scheduler logic


src/proto/faabric.proto
```diff
@@ -93,12 +96,10 @@ message MPIMessage {
...
-    string user = 1;
-    string function = 2;
-    string snapshotKey = 4;
-    int32 snapshotSize = 5;
-    int32 funcPtr = 6;
-    int32 returnValue = 7;
+    int32 id = 1;
+    int32 appId = 2;
+    int32 appIndex = 3;
+    string masterHost = 4;
```
Shillaker, line 102: I introduced these fields, but rearranged and renumbered the rest as they didn't make sense. According to the protobuf docs it's also most efficient to use numbers 1-15 for regularly used fields, so I made sure that's the case.

include/faabric/executor/FaabricMain.h
```diff
@@ -1,23 +1,31 @@
 #pragma once
 
-#include <faabric/executor/FaabricPool.h>
+#include <faabric/scheduler/ExecutorFactory.h>
```
Shillaker, line 3: This now has the stuff that used to be in `FaabricPool` as well (i.e. managing the different servers)

src/scheduler/Scheduler.cpp
```diff
@@ -193,31 +185,19 @@ void Scheduler::notifyFaasletFinished(const faabric::Message& msg)
...
-}
-
-Scheduler& getScheduler()
-{
-    // Note that this ref is shared between all faaslets on the given host
-    static Scheduler scheduler;
-    return scheduler;
-}
-
 std::vector<std::string> Scheduler::callFunctions(
```
Shillaker, line 188: The logic the `callFunctions` is the heart of the scheduler and was a bit sketchy, so it's undergone quite a bit tidy-up. Best to look at the full thing: https://github.com/faasm/faabric/blob/batch-execution/src/scheduler/Scheduler.cpp#L188

src/scheduler/Scheduler.cpp
```diff
@@ -500,71 +454,94 @@ int Scheduler::scheduleFunctionsOnHost(const std::string& host,
...
           "Scaling {} {}->{} faaslets", funcStr, nFaaslets, nFaaslets + 1);
+        doAddFaaslets(msg, difference);
+    }
+}
+
+void Scheduler::doAddFaaslets(const faabric::Message& msg, int count)
+{
+    const std::string funcStr = faabric::util::funcToString(msg, false);
 
-        // Increment faaslet count
```
Shillaker, line 552: Adding an executor is now a case of creating a new instance of the class, rather than sending a bind message to a pool

include/faabric/scheduler/DummyExecutor.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
```
Shillaker, line 1: `DummyExecutor` and the associated factory are needed to set a default in tests

mpi-native/examples/mpi_allgather.cpp
```diff
@@ -3,43 +3,14 @@
 
 #include <faabric/mpi-native/MpiExecutor.h>
 #include <faabric/util/logging.h>
+
 int main(int argc, char** argv)
 {
-    auto logger = faabric::util::getLogger();
```
Shillaker, line 8: This chunk of code seemed to be duplicated over all MPI native examples so I factored it out

src/scheduler/Scheduler.cpp
```diff
@@ -226,203 +226,179 @@ std::vector<std::string> Scheduler::callFunctions(
...
-            // Now we need to find hosts that aren't already registered for
-            // this function and add them
-            if (remainder > 0) {
-                std::unordered_set<std::string>& thisRegisteredHosts =
-                  registeredHosts[funcStr];
-
-                std::unordered_set<std::string> allHosts = getAvailableHosts();
+        if (offset < nMessages) {
+            // At this point we know we need to enlist unregistered hosts
+            std::unordered_set<std::string> allHosts = getAvailableHosts();
```
csegarragonz, line 317: If we were to use an ordered container like `std::set` instead of `std::unordered_set` we could use [`std::set_difference`](https://en.cppreference.com/w/cpp/algorithm/set_difference), which feels a bit like what we are doing.

src/scheduler/Scheduler.cpp
```diff
@@ -500,72 +478,69 @@ int Scheduler::scheduleFunctionsOnHost(const std::string& host,
 void Scheduler::callFunction(faabric::Message& msg, bool forceLocal)
 {
     // TODO - avoid this copy
```
csegarragonz, line 480: AFAICT we are not copying anymore? Could we remove the `TODO`?

csegarragonz (approved):
LGTM. We can go over the details offline.

src/scheduler/Scheduler.cpp
Shillaker: Yes that might make sense here. In general I always avoid the ordered collections as the performance of `std::map` is noticeably different to `std::unordered_map` when doing inserts. In this case the set will be static so that would be good.

src/scheduler/Scheduler.cpp
```diff
@@ -20,20 +21,19 @@ using namespace faabric::util;
...
 }
 
 Scheduler::Scheduler()
   : thisHost(faabric::util::getSystemConfig().endpointHost)
   , conf(faabric::util::getSystemConfig())
 {
-    bindQueue = std::make_shared<InMemoryMessageQueue>();
-
     // Set up the initial resources
     int cores = faabric::util::getUsableCores();
```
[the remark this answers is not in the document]
Shillaker: Not sure I agree, this number is the number of cores, so calling it cores makes sense. We're then setting the number of slots equal to the number of cores, but the two are different things (i.e. we could have more slots than cores if we found that overloading didn't make much of a performance difference).

tests/test/scheduler/test_executor.cpp
```diff
@@ -0,0 +1,308 @@
...
+    std::shared_ptr<BatchExecuteRequest> req =
+      faabric::util::batchExecFactory("dummy", "simple", 1);
+    uint32_t msgId = req->messages().at(0).id();
+
+    REQUIRE(req->messages_size() == 1);
+
+    executeWithTestExecutor(req, false);
+
+    auto& sch = faabric::scheduler::getScheduler();
+    faabric::Message result = sch.getFunctionResult(msgId, 1000);
```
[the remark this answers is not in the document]
Shillaker: Yes having magic numbers like this isn't great. They're there to make sure the tests time out if they go wrong, but to give the system time to perform the action we're testing in the background. The bound timeout is usually longer than we'd want a test to hang before failing, but perhaps I can find a nicer way to set these values.

src/scheduler/Scheduler.cpp
Shillaker: I'm pretty sure `*req->add_messages() = msg;` is doing a copy. If this function weren't so widely used I'd get rid of it altogether; all scheduling of functions should be done through `callFunctions`. Fortunately it's mostly just used in tests, and I'll port the important bits of Faasm (like the function chaining calls) to use `callFunctions`.

src/runner/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+set(LIB_FILES
+        "${FAABRIC_INCLUDE_DIR}/faabric/runner/FaabricMain.h"
+        FaabricMain.cpp
+        )
+
+faabric_lib(runner "${LIB_FILES}")
```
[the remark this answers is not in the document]
Shillaker: I'm not 100% sure whether it works with the `faabric_lib` function, if it was `add_library` I think it would be fine. I'll see if it's a straight swap, if not we can update all uses of `faabric_lib` as part of that issue.

PR #84, opened 2021-04-24
Async grpc

csegarragonz:
In this PR I introduce the multi-threaded version of the async server.

The client and server pair are implemented, but the system can not handle out-of-order MPI message responses, and our MPI runtime still relies on those.

Using a multi-threaded asynchronous transport layer breaks the inherent message ordering in MPI.


csegarragonz:
Closing after #85 has been merged.

PR #85, opened 2021-04-26
ZeroMQ as message transport layer

csegarragonz:
This PR introduces ZeroMQ as the underlying messging transport layer for Faabric. We move from an RPC oriented system, to a message-based one. As a consequence, we no longer need `gRPC` as a dependency, but still rely on `protobuf` and `flatbufferss` for message serialisation.

The abstractions I introduce are the following:

**1. Message endpoint:** socket-like object that can be `open`-ed, `close`-ed, and can `send` and `recv`.  It expects a `char*` + `size` for sending, and, delegates the -blocking- recv to a virtual `doRecv` method. The `MessageEndpointClient` class provides a client-like implementation, focused on `send`-ing, which overwrites the `recv` to block on synchronous calls. There are different sub-classes specialized for different means: `SendMessageEndpoint`, `RecvMessageEndpoint`, and `MessageEndpointClient` (a `SendMessageEndpoint` with blocking behaviour).

**2. Message endpoint server:** an asyncrhonous message server. The server can be `start`-ed and `stop`-ed. Synchronous behaviour may be achieved with additional calls to `sendResponse` from implementations. It forcefully adheres to a protocol of: message header + message body.

**3. Message and Message context:** lightweight wrappers arround 0MQ's `context_t` and `message_t`.

By removing `gRPC` from the dependencies, there was no real need for the `grpc-root` dockerfile, which has been re-shaped into the `faabric-base` one. Additionally, I've changed `protobuf`'s installation to rely on CMake. Even though I've tried to change the least the interfaces exposed by the components outside the transport layer, it's been inevitable to change several things here and there, hence the number of files changed.


include/faabric/scheduler/Scheduler.h
```diff
@@ -169,6 +172,11 @@ class Scheduler
 
     std::unordered_map<uint32_t, std::promise<int32_t>> threadResults;
 
+    std::unordered_map<std::string, faabric::scheduler::FunctionCallClient>
```
csegarragonz, line 175: A big design decision all around has been what to do with clients: persist them or open/close them.
* By persisting them we would expect a certain performance gain (assumption, not fact).
* By opening and closing them we have easier code to manage.
Message clients (aka 0MQ sockets) are _not_ thread safe and _must_ be opened and closed from the same thread.

I have followed the (always wrong) path of optimizing before measuring, having three different implementations:
* Thread-local, persistent, clients in `MpiWorld`. This handle _tons_ of messages, so it _theoretically_ made sense to try and optimize.
* Persistent clients in the scheduler.
* One-off clients in the state: only for the `inmemory` version.

I am open to re-considering any of this design choices.

src/scheduler/FunctionCallClient.cpp
```diff
@@ -123,29 +167,53 @@ void FunctionCallClient::sendMPIMessage(
...
+        // Serialise using protobuf
+        if (!msg->SerializeToArray(serialisedMsg, msgSize)) {
+            throw std::runtime_error("Error serialising message");
+        }
+        send(serialisedMsg, msgSize);
     }
 }
 
-faabric::HostResources FunctionCallClient::getResources(
-  const faabric::ResourceRequest& req)
```
csegarragonz, line 133: This request was empty, and we needed not to create it outside the function's scope.

src/scheduler/FunctionCallServer.cpp
```diff
@@ -1,110 +1,176 @@
 #include <faabric/scheduler/FunctionCallServer.h>
```
csegarragonz, line 1: This diff is painful to watch. You can examine the source [here](https://github.com/faasm/faabric/blob/zeromq/src/scheduler/FunctionCallServer.cpp).

src/state/DummyStateServer.cpp
```diff
@@ -49,38 +49,34 @@ void DummyStateServer::start()
...
-            if (!inMemKv->isMaster()) {
-                logger->error("Dummy state server not master for data");
-                throw std::runtime_error("Remote state server failed");
-            }
-
-            // Set the data
-            kv->set(dummyData.data());
-            logger->debug(
-              "Finished setting master for test {}/{}", kv->user, kv->key);
+    const std::shared_ptr<spdlog::logger>& logger = faabric::util::getLogger();
```
csegarragonz, line 52: We need not to start the server on another thread, as it runs in the background by default.

src/state/StateClient.cpp
```diff
@@ -1,138 +1,193 @@
...
+    char* serialisedMsg = new char[requestSize];
+    // Serialise using protobuf
+    if (!request.SerializeToArray(serialisedMsg, requestSize)) {
+        throw std::runtime_error("Error serialising message");
+    }
+    send(serialisedMsg, requestSize);
+}
 
-    stream->WritesDone();
+void StateClient::pushChunks(const std::vector<StateChunk>& chunks)
```
csegarragonz, line 79: Note that `pushChunks` and `pullChunks` used to be streaming RPCs, whereas now we send as many messages as chunks, and acknowledge them _one by one_.

src/state/StateClient.cpp
```diff
@@ -1,138 +1,193 @@
 #include <faabric/state/StateClient.h>
```
csegarragonz, line 1: As discussed at the description of the PR, the state API is blocking. Thus, even if messages expect no response, they block until receiving an empty response from the server.

src/state/StateServer.cpp
```diff
@@ -7,208 +7,277 @@
 #include <faabric/util/logging.h>
```
csegarragonz, line 7: Another grim diff, see the raw source [here](https://github.com/faasm/faabric/blob/zeromq/src/state/StateServer.cpp).

tests/test/scheduler/test_function_client_server.cpp
```diff
@@ -32,14 +32,17 @@ TEST_CASE("Test sending MPI message", "[scheduler]")
...
-    int worldSize = 2;
+    faabric::Message msg;
+    msg.set_user(user);
+    msg.set_function(func);
+    msg.set_mpiworldid(worldId);
+    msg.set_mpiworldsize(2);
+    faabric::util::messageFactory(user, func);
 
     scheduler::MpiWorldRegistry& registry = getMpiWorldRegistry();
-    scheduler::MpiWorld& localWorld = registry.createWorld(msg, worldId);
```
csegarragonz, line 40: Here we created the world _twice_, once through the registry, another one directly calling `create`.

tests/test/state/test_state_server.cpp
```diff
@@ -95,6 +94,7 @@ TEST_CASE("Test request/ response", "[state]")
         std::vector<StateChunk> chunks = { chunkA, chunkB, chunkC };
         std::vector<uint8_t> expected = { 0, 1, 2, 3, 4, 5, 0, 7 };
         client.pullChunks(chunks, actual.data());
+        REQUIRE(actual == expected);
```
csegarragonz, line 97: We weren't checking anything here before.

cmake/ExternalProjects.cmake
```diff
@@ -52,13 +63,13 @@ set_target_properties(pistache
 )
 
 # RapidJSON
-set(RAPIDJSON_BUILD_DOC OFF CACHE INTERNAL "")
```
csegarragonz, line 55: After double-checking, and inspecting the contents of `/build/faabric` after a clean sweep, this was actually not enforced. This other option through `CMAKE_CACHE_ARGS` seems to do it.

src/scheduler/SnapshotServer.cpp
```diff
@@ -1,40 +1,39 @@
-#include <faabric/scheduler/MpiWorldRegistry.h>
```
csegarragonz, line 1: You can check a cleaner diff [here](https://github.com/faasm/faabric/blob/zeromq/src/scheduler/SnapshotServer.cpp).

src/transport/MessageEndpoint.cpp
```diff
@@ -0,0 +1,119 @@
...
+    } else {
+        if (!this->socket->send(msg, zmq::send_flags::none)) {
+            throw std::runtime_error("Error sending message through socket");
+        }
+    }
+}
+
+// This method overload is used to send flatbuffers that deal with unsigned
+// chars, and are stack-allocated.
+void MessageEndpoint::send(uint8_t* serialisedMsg, size_t msgSize, bool more)
```
csegarragonz, line 81: The alternative was to define a template function, and check the argument type with something like:
```cpp
if constexpr(std::is_same_v<T, char>)
```
but in the end it added more complexity imo.

src/state/DummyStateServer.cpp
```diff
@@ -49,41 +49,37 @@ void DummyStateServer::start()
...
+    }
+
+    // Start the state server
+    // Note - by default the state server runs in a background thread
+    logger->debug("Running state server");
+    stateServer.start();
 
     // Give it time to start
-    usleep(500 * 1000);
+    usleep(1000 * 1000);
```
csegarragonz, line 82: Copied this from #86

tests/test/main.cpp
```diff
@@ -13,9 +13,9 @@ struct LogListener : Catch::TestEventListenerBase
     void testCaseStarting(Catch::TestCaseInfo const& testInfo) override
     {
         auto logger = faabric::util::getLogger();
-        logger->debug("---------------------------------------------");
-        logger->debug("TEST: {}", testInfo.name);
-        logger->debug("---------------------------------------------");
+        logger->info("---------------------------------------------");
```
csegarragonz, line 16: Copied from #86

src/transport/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+    MessageEndpointClient.cpp
+    MessageEndpointServer.cpp
+    ${HEADERS}
+)
+
+faabric_lib(transport "${LIB_FILES}")
+
+# ----------------------------------------------
+# Add dependency with ZeroMQ
+# ----------------------------------------------
```
Shillaker, line 25: Maybe don't need such a big comment for this one `target_link_libraries` :smile:

src/transport/MessageEndpoint.cpp
```diff
@@ -0,0 +1,113 @@
...
+        case faabric::transport::SocketType::PULL:
+            this->socket = std::make_unique<zmq::socket_t>(
+              context.get(), zmq::socket_type::pull);
+            break;
+        default:
+            throw std::runtime_error("Unrecognized socket type");
+    }
+
+    // Set linger period to zero. This avoids blocking if closing a socket with
+    // outstanding messages.
```
Shillaker, line 41: Is this a good thing? Shouldn't we block when closing to wait for the messages to be processed?

src/transport/MessageEndpoint.cpp
Shillaker: Having the `char*` version delete the data that's being passed in is quite unexpected IMO. If I'm passing data to a function via a pointer, I wouldn't expected it to be deleted. Do we really need to pass 0MQ a deallocation function to avoid it doing a copy? Unless you can't avoid it, I think it's safest to assume the caller is responsible for deleting a pointer to data they have created.

Also, couldn't we change the `char` version to be:

```
void MessageEndpoint::send(char* serialisedMsg, size_t msgSize, bool more)
{
    send(reinterpret_cast<uint8_t *>(serialisedMsg), msgSize, more);    
}
```

src/transport/MessageEndpointServer.cpp
```diff
@@ -0,0 +1,108 @@
...
+    // Finally join the serving thread
+    if (this->servingThread.joinable()) {
+        this->servingThread.join();
+    }
+}
+
+void MessageEndpointServer::recv(MessageEndpointClient& endpoint)
+{
+    // Receive header
+    if (!endpoint.socket) {
```
Shillaker, line 64: Generally I would avoid relying on an object's `operator !` if you are just checking for `nullptr`, and instead write `if(endpoint.socket == nullptr)`. As it stands, this condition could evaluate to `false` when the socket is _not_ a null pointer (depending on the implementation of `operator !`), thus leading to an invalid error message in this case.

src/transport/MessageEndpointServer.cpp
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    // Check the header was sent with ZMQ_SNDMORE flag
+    if (!header.more()) {
+        throw std::runtime_error("Header sent without SNDMORE flag");
+    }
+
+    // Receive body
+    zmq::message_t body;
+    if (!endpoint.socket->recv(body)) {
```
Shillaker, line 80: Probably getting into personal preference here but unless `recv` returns a `bool` I would be more explicit in the condition check, e.g. `if(endpoint.socket->recv(body) <= 0)` as it's easier to read, easier to diagnose when something goes wrong, and in this case I assume means number of bytes received?.

I would also try and include more information in these error messages as they're a bit vague, for example:

```
int bytesReceived = endpoint.socket->recv(body);
if(bytesReceived <= 0) {
    throw std::runtime_error(fmt::format("Error receiving message through socket, got {} bytes", bytesReceived));
}
```

tasks/container.py
```diff
@@ -59,11 +59,11 @@ def build(ctx, nocache=False, push=False):
 
 
 @task
-def build_grpc(ctx, nocache=False, push=False):
+def build_base(ctx, nocache=False, push=False):
     """
-    Build current base gRPC container
+    Build current native MPI container
```
Shillaker, line 64: Not sure this comment is right

tasks/container.py
```diff
@@ -83,11 +83,11 @@ def push(ctx):
 
 
 @task
-def push_grpc(ctx):
+def push_base(ctx):
     """
-    Push current version of gRPC container
+    Push current version of the native MPI container
```
Shillaker, line 88: Also not sure this comment is right

Shillaker (changes requested):
Wow, what a beast, great work!

Obviously the majority of the work is excellent and you've done a great job keeping all the interfaces relatively untouched (evidenced by the small number of changes in the tests). 

Lots of comments but mostly superficial stuff around structure of the interfaces etc.

include/faabric/scheduler/FunctionCallServer.h
```diff
@@ -1,44 +1,45 @@
...
     FunctionCallServer();
 
-    Status Flush(ServerContext* context,
-                 const faabric::Message* request,
-                 faabric::FunctionStatusResponse* response) override;
+    /* Stop the function call server
+     *
+     * Override the base stop method to do some implementation-specific cleanup.
+     */
+    void stop();
```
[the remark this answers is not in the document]
csegarragonz: It is not enforced (that I am aware) as this was indeed overriding. A quick google showed [modernize-use-override](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-override.html) for `clang-tidy`. Should we include it?

src/scheduler/FunctionCallClient.cpp
```diff
@@ -96,24 +93,62 @@ void clearMockRequests()
...
+
+        // Send the message body
+        call.set_returnhost(faabric::util::getSystemConfig().endpointHost);
+        size_t msgSize = call.ByteSizeLong();
+        char* serialisedMsg = new char[msgSize];
+        // Serialise using protobuf
+        if (!call.SerializeToArray(serialisedMsg, msgSize)) {
+            throw std::runtime_error("Error serialising message");
+        }
+        send(serialisedMsg, msgSize);
```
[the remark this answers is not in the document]
csegarragonz: Yes, the macro is looking great. Wrt the `delete` I deliberately omitted it. This is because I am setting the cleanup callback in 0MQ's send, for it to be zero copy. Note that this was taken from a [2010 tutorial](http://wiki.zeromq.org/blog:zero-copy). More on this on the other comments.

src/scheduler/FunctionCallServer.cpp
```diff
@@ -1,122 +1,157 @@
...
-    *response = sch.getThisHostResources();
+    faabric::ResponseRequest request;
+
+    // Deserialise message string
+    if (!request.ParseFromArray(msgData, size)) {
+        throw std::runtime_error("Error deserialising message");
+    }
+
+    // Clear out any cached state
+    faabric::state::getGlobalState().forceClearAll(false);
```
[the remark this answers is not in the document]
csegarragonz: Nope. This was already [there](https://github.com/faasm/faabric/blob/65b81c528a150afe2c486f3ba0fd35ad185037f3/src/scheduler/FunctionCallServer.cpp#L40). It's just that this diff is grim to read :rofl:

include/faabric/transport/MessageEndpoint.h
```diff
@@ -0,0 +1,58 @@
...
+    void close();
+
+    void send(char* serialisedMsg, size_t msgSize, bool more = false);
+
+    // Overload for sending flatbuffers
+    void send(uint8_t* serialisedMsg, size_t msgSize, bool more = false);
+
+    void recv();
+
+    std::unique_ptr<zmq::socket_t> socket;
```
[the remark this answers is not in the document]
csegarragonz: Currently, `MessageEndpointServer` has to access the underlying socket for server-specific send/recv. I also agree that having it public is not nice. I will set `MessageEndpointServer` as a [friend](https://en.cppreference.com/w/cpp/language/friend) class of `MessageEndpoint`. It used to inherit from it, thus they have always been tightlyt coupled.

include/faabric/scheduler/FunctionCallServer.h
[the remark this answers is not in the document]
Shillaker: Ah ok good spot, yes I think including that wouldn't hurt, IIRC most of our classes should conform to that.

include/faabric/transport/MessageEndpointServer.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include <thread>
+
+namespace faabric::transport {
+/* Server handling a long-running 0MQ socket
+ *
+ * This abstract class implements a server-like loop functionality and will
+ * always run in the background. Note that message endpoints (i.e. 0MQ sockets)
+ * are _not_ thread safe, must be open-ed and close-ed from the _same_ thread,
+ * and thus should preferably live in the thread's local address space.
```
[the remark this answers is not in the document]
csegarragonz: I agree with the previous comment on storing the `tid` as a variable and checking against that, but not so much on this one. In particular, the message endpoint server uses no thread-local storage. Given that the server socket runs in a background thread, I did not want this running socket to be a member of the server class, rather to be defined, opened, and closed, from the serving thread.

src/transport/MessageEndpointServer.cpp
csegarragonz: After checking the sources, it is a bit more involved than that. In particular, [`zmq::recv`](https://github.com/zeromq/cppzmq/blob/b65dde8725b2eb6d174a10e9e57891de554fd175/zmq.hpp#L1960) returns an optional value: either `numBytes` if it goes right, or an empty object `{}` otherwise. The C++ binding (cppzmq, the link I share before), already does the actual checks. IMO this is fine like it is.

src/transport/MessageEndpoint.cpp
csegarragonz: Yes, I think you are right :+1:

include/faabric/transport/macros.h
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+#define PARSE_MSG(T, data, size)                                               \
+    T msg;                                                                     \
+    if (!msg.ParseFromArray(data, size)) {                                     \
+        throw std::runtime_error("Error deserialising message");               \
+    }
+
+#define PARSE_RESPONSE(T, data, size)                                          \
+    T response;                                                                \
```
csegarragonz, line 42: I maybe could have included the variable name in the macro, but using variables defined _inside_ the macro makes the code hard to read, so I decided to have two different macros.

src/mpi_native/mpi_native.cpp
```diff
@@ -235,7 +235,6 @@ int MPI_Bcast(void* buffer,
         world.broadcast(
           rank, (uint8_t*)buffer, datatype, count, faabric::MPIMessage::NORMAL);
     } else {
-        auto fptr = &faabric::scheduler::MpiWorld::recv;
```
csegarragonz, line 238: This variable was unused, and adding the compiler checks to `faabric` raised the issue.

src/scheduler/FunctionCallClient.cpp
```diff
@@ -123,29 +127,35 @@ void FunctionCallClient::sendMPIMessage(
...
-                  stub->GetResources(&context, req, &response));
+        request.set_returnhost(faabric::util::getSystemConfig().endpointHost);
+
+        SEND_MESSAGE(faabric::scheduler::FunctionCalls::GetResources, request);
+
+        // Receive message
+        faabric::transport::Message msg =
+          awaitResponse(faabric::util::getSystemConfig().endpointHost,
+                        FUNCTION_CALL_PORT + REPLY_PORT_OFFSET);
+        // Deserialise message string
```
csegarragonz, line 155: I don't use `PARSE_RESPONSE` here because the response variable is defined in a larger scope (both for mock and non-mock mode). Alternatively, I could define and return from within both branches,

src/transport/MessageEndpoint.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// TODO - use optional size parameter
+Message MessageEndpoint::recv(int size)
+{
+    assert(tid == std::this_thread::get_id());
+    assert(this->socket != nullptr);
+    assert(size >= 0);
+
+    // Pre-allocate buffer to avoid copying data
```
csegarragonz, line 80: Even though we still don't use it, this is how we would have 0MQ receive in a pre-allocated buffer, avoiding one check. We would just need to call `recv` with a buffer size up front.

tests/test/scheduler/test_function_client_server.cpp
```diff
@@ -16,6 +16,13 @@
 
 using namespace scheduler;
 
+static void tearDown(FunctionCallClient& cli, FunctionCallServer& server)
```
csegarragonz, line 19: I have decided to be verbose, and add a `tearDown` function per file. Hopefuly this aids when moving to text fixtures as discussed off line.

src/scheduler/SnapshotServer.cpp
```diff
@@ -1,40 +1,47 @@
...
-Status SnapshotServer::PushSnapshot(
-  ServerContext* context,
-  const flatbuffers::grpc::Message<SnapshotPushRequest>* request,
-  flatbuffers::grpc::Message<SnapshotPushResponse>* response)
+void SnapshotServer::recvPushSnapshot(faabric::transport::Message& msg)
 {
-    const SnapshotPushRequest* r = request->GetRoot();
+    // Persist message to ensure contents are not cleared
+    // Note that these are currently not free-d, see TODO in next function
+    msg.persist();
```
csegarragonz, line 35: The memory `mmap`-ed for the snapshot push request needs _not_ to be freed when the `transport::Messge` goes out of scope, so we mark it as persistent. We though need to make sure we `free` the underlying pointer aftterwards.

tests/test/scheduler/test_mpi_context.cpp
```diff
@@ -7,6 +7,15 @@
 
 using namespace faabric::scheduler;
 
+static void tearDown(std::vector<MpiWorld*> worlds)
```
Shillaker, line 10: Looks like this is only ever used with one world, might be able to simplify

tests/test/scheduler/test_function_client_server.cpp
```diff
@@ -273,10 +287,10 @@ TEST_CASE("Test unregister request", "[scheduler]")
     reqB.set_host(otherHost);
     *reqB.mutable_function() = msg;
     cli.unregister(reqB);
+    usleep(1000 * 100);
```
Shillaker, line 290: The minimum I would do is 500ms just to be safe in slow environments (e.g. GHA)

tests/test/transport/test_message_endpoint_client.cpp
```diff
@@ -0,0 +1,229 @@
...
+        if ((i % (numMessages / 10)) == 0) {
+            std::string expectedMsg = baseMsg + std::to_string(i);
+            REQUIRE(recvMsg.size() == expectedMsg.size());
+            std::string actualMsg(recvMsg.data(), recvMsg.size());
+            REQUIRE(actualMsg == expectedMsg);
+        }
+    }
+
+    // Wait for the sender thread to finish
+    senderThread.join();
```
Shillaker, line 166: `joinable`

include/faabric/transport/MessageEndpoint.h
```diff
@@ -0,0 +1,86 @@
...
+
+  protected:
+    const std::string host;
+    const int port;
+    std::thread::id tid;
+};
+
+/* Send and Recv Message Endpoints */
+
+class SendMessageEndpoint : public MessageEndpoint
```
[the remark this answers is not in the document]
csegarragonz: `MessageEndpointClient` inherits from `SendMessageEndpoint` and maybe at some point we derive another subclass from `RecvMessageEndpoint`, so I'd rather not make them `final`.

src/transport/Message.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+// Empty message signals shutdown
+Message::Message()
+  : msg(nullptr)
+{}
+
+Message::~Message()
+{
+    if (!_persist) {
+        free(reinterpret_cast<void*>(msg));
```
csegarragonz, line 31: Note that if the message goes out of scope, but is marked as persistent, we will have a dangling pointer that we need to free ourselves. While risky, this is necessary for the snapshot server.

src/transport/MessageEndpoint.cpp
```diff
@@ -0,0 +1,303 @@
+#include <faabric/transport/MessageEndpoint.h>
```
csegarragonz, line 1: This has become the only file where we interact with ZeroMQ. After tedious debugging, I've wrapped all zeromq calls in `try/catch` blocks as otherwise exceptions were lost sometimes.

As a consequence, this file is now quite verbose, but we have hid the complexity from the others.

src/transport/MessageEndpoint.cpp
```diff
@@ -0,0 +1,303 @@
...
+    assert(this->socket != nullptr);
+
+    const auto& logger = faabric::util::getLogger();
+
+    if (more) {
+        try {
+            auto res = this->socket->send(zmq::buffer(serialisedMsg, msgSize),
+                                          zmq::send_flags::sndmore);
+            if (res != msgSize) {
+                throw std::runtime_error(
```
csegarragonz, line 97: I had problems _catching_ exceptions and then re-throwing `std::runtime_error`s. My takeaway is that throwing _new_ exceptions from within a `catch` block can be risky, but otherwise should be fine.

Instead of throwing new ones, I've decided to log and re-throw.

cmake/ExternalProjects.cmake
```diff
@@ -107,6 +116,33 @@ FetchContent_MakeAvailable(zstd_ext)
 target_include_directories(libzstd_static INTERFACE $<BUILD_INTERFACE:${zstd_ext_SOURCE_DIR}/lib>)
 add_library(zstd::libzstd_static ALIAS libzstd_static)
 
+# ZeroMQ
+set(ZEROMQ_LIBRARY ${CMAKE_INSTALL_PREFIX}/lib/libzmq.so)
+ExternalProject_Add(libzeromq_ext
+    GIT_REPOSITORY "https://github.com/zeromq/libzmq.git"
+    GIT_TAG "v4.3.4"
+    CMAKE_CACHE_ARGS "-DCMAKE_INSTALL_PREFIX:STRING=${CMAKE_INSTALL_PREFIX}"
```
csegarragonz, line 124: `libzmq` defaults to [Release](https://github.com/zeromq/libzmq/blob/78ea4ee7878779fb72d3551fb0d79b7b0a5ecd22/CMakeLists.txt#L841) build.

src/transport/MessageEndpoint.cpp
Shillaker: I generally go for a detailed logging statement then a generic "why" in the exception message. Here and on line 112 you've got a detailed "why" message and no log line (i.e. using `fmt::format` in the exception constructor). In threaded code the exception message can get lost or be nested in a stack trace, whereas if we always log before throwing an exception, you can scroll up to the line just before it to see what happened.

src/transport/MessageEndpoint.cpp
```diff
@@ -0,0 +1,302 @@
...
+                              msgSize));
+            }
+        } catch (zmq::error_t& e) {
+            logger->error("Error sending message: {}", e.what());
+            throw;
+        }
+    }
+}
+
+// TODO - use optional size parameter
```
Shillaker, line 125: Why would we want an optional size parameter?

src/transport/MessageEndpoint.cpp
```diff
@@ -0,0 +1,302 @@
...
+        Message msg(size);
+
+        try {
+            auto res = this->socket->recv(zmq::buffer(msg.udata(), msg.size()));
+            if (res.has_value() && (res->size != res->untruncated_size)) {
+                throw std::runtime_error(
+                  fmt::format("Received more bytes than buffer can hold. "
+                              "Received: {}, capacity {}",
+                              res->untruncated_size,
+                              res->size));
```
Shillaker, line 145: Same point about log line + exception vs. just exception.

src/transport/MessageEndpoint.cpp
```diff
@@ -0,0 +1,302 @@
...
+    return Message(msg);
+}
+
+void MessageEndpoint::close(bool bind)
+{
+    if (this->socket != nullptr) {
+        const auto& logger = faabric::util::getLogger();
+
+        if (tid != std::this_thread::get_id()) {
+            logger->error("Closing socket from a different thread");
```
Shillaker, line 189: Having an error line with no exception or failure seems strange. Either it's something we expect, which isn't necessarily a failure but should be obvious when investigating issues (in which case it would be a `warn`), or it's a failure, in which case we'd put more information in this `error` log (e.g. which thread IDs it is), then throw an exception.

tests/test/scheduler/test_function_client_server.cpp
Shillaker: I think we'd have a `FunctionCallClientServerFixture` that would set up both the client and server, then perform this tear down in its destructor. This should be relatively easy to configure similar to the [`DistTestsFixture`](https://github.com/faasm/faabric/blob/master/tests/dist/fixtures.h). This would save a couple of lines of repeated setup code in the tests in this file.

tests/test/transport/test_message_endpoint_client.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    // Close receiving endpoints
+    dst.close();
+    dstResponse.close();
+
+    // Close message context
+    context.close();
+}
+
+TEST_CASE("Test send/recv many messages", "[transport]")
```
Shillaker, line 125: These tests could also have a `MessageContextFixture` which sets up and tears down the message context. Might be overkill as it's only a couple of lines in each place, but you get the picture.

Shillaker (changes requested):
Couple of minor questions/ tidy-ups, otherwise bulk of it looks good

src/transport/MessageEndpoint.cpp
csegarragonz: I've updated the comment, the optional size parameter is aimed at receiving into a pre-allocated buffer. I left a `TODO` as it's currently not used anywhere, but changed it to an informative comment.

include/faabric/transport/MessageEndpoint.h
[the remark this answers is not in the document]
Shillaker: Ah ok, I didn't realise `MessageEndpointClient` inherited from this. Looking at it now, the `MessageEndpointClient` class looks a little thin, it's got a pass-through constructor and just one extra method. It's probably not worth doing in this PR but we could just put that method in this class and get rid of `MessageEndpointClient` (AFAICT).  

In general I like `final`-by-default if subclassing. It's just a style thing, but I think it improves readability as you know you've hit the bottom of the class hierarchy and you will think a bit more before subclassing (at which point you can of course remove the `final` if you want).

Shillaker (approved):
Nice, just one small ignor-able point re. `final` subclasses. 

Let's merge this beast!

PR #86, opened 2021-05-05
Executor fixes

Shillaker:
After testing out the new Executor set-up in anger, added the following changes:

- Give Faabric responsibility for calling `restore` and `reset` as part of the `Executor` lifecycle (`restore` gets called to restore a snapshot, and `reset` when a function has finished).
- In the scheduler, avoid use of maps with an integer index where we know the keys will just be a range 0 to N (instead use an array).
- Have the scheduler "claim" `Executor`s when they're executing, rather than switching them between lists of `warm` and `executing` executors.
- Add more logging to the `Executor` lifecycle (mostly `trace`-level to avoid bloat in normal circumstances).
- Add more detailed tests to check overloading, chaining, executing different functions and restore counts.
- Avoid the use of `REQUIRE` in more than one thread in Catch2 tests (as it's [not safe](https://github.com/catchorg/Catch2/blob/devel/docs/limitations.md#thread-safe-assertions)).
- Fix small bug thread pool index allocation, where threads must be executed by the thread pool index `>= 1`, leaving `0` free to execute the spawning function (thus avoiding deadlocks).

I've also sprinkled a few `assert`s around the place. I've tried to stick to the rule of checking inputs to "external" interfaces, or to sense-check a bit of thorny logic (e.g. in the scheduler). Hopefully these should catch slip ups in the future.

### DIFF
--- a/include/faabric/scheduler/Scheduler.h
+++ b/include/faabric/scheduler/Scheduler.h
@@ -70,6 +70,8 @@ class Executor
       std::pair<int, std::shared_ptr<faabric::BatchExecuteRequest>>>>
       threadQueues;
 
+    void threadPoolThread(int threadPoolIdx);
+
     void shutdownThreadPoolThread(int threadPoolIdx);
 };
 
--- a/include/faabric/scheduler/Scheduler.h
+++ b/include/faabric/scheduler/Scheduler.h
@@ -43,6 +43,12 @@ class Executor
       int msgIdx,
       std::shared_ptr<faabric::BatchExecuteRequest> req);
 
+    bool tryClaim();
+
+    void claim();
+
+    void releaseClaim();
+
   protected:
     virtual void restore(const faabric::Message& msg);
 
@@ -53,16 +59,15 @@ class Executor
   private:
     std::string lastSnapshot;
 
+    std::atomic<bool> claimed = false;
+
     std::atomic<int> executingTaskCount = 0;
 
     std::mutex threadsMutex;
     uint32_t threadPoolSize = 0;
-    std::unordered_map<int, std::thread> threadPoolThreads;
-
-    std::unordered_map<
-      int,
-      faabric::util::Queue<
-        std::pair<int, std::shared_ptr<faabric::BatchExecuteRequest>>>>
+    std::vector<std::shared_ptr<std::thread>> threadPoolThreads;
+    std::vector<faabric::util::Queue<
+      std::pair<int, std::shared_ptr<faabric::BatchExecuteRequest>>>>
       threadQueues;
 
     void shutdownThreadPoolThread(int threadPoolIdx);
@@ -162,10 +167,7 @@ class Scheduler
     faabric::util::SystemConfig& conf;
 
     std::unordered_map<std::string, std::vector<std::shared_ptr<Executor>>>
-      executingExecutors;
-
-    std::unordered_map<std::string, std::vector<std::shared_ptr<Executor>>>
-      warmExecutors;
+      executors;
 
     std::shared_mutex mx;
 
--- a/include/faabric/scheduler/Scheduler.h
+++ b/include/faabric/scheduler/Scheduler.h
@@ -120,8 +120,6 @@ class Scheduler
 
     void vacateSlot();
 
-    void notifyExecutorFinished(Executor* exec, const faabric::Message& msg);
-
     void notifyExecutorShutdown(Executor* exec, const faabric::Message& msg);
 
     std::string getThisHost();
--- a/include/faabric/scheduler/Scheduler.h
+++ b/include/faabric/scheduler/Scheduler.h
@@ -36,7 +36,7 @@ class Executor
 
     virtual void flush();
 
-    virtual void reset();
+    virtual void reset(const faabric::Message& msg);
 
     virtual int32_t executeTask(
       int threadPoolIdx,
--- a/include/faabric/scheduler/Scheduler.h
+++ b/include/faabric/scheduler/Scheduler.h
@@ -46,9 +46,9 @@ class Executor
 
     virtual void postFinish();
 
-  private:
     faabric::Message boundMessage;
 
+  private:
     std::string lastSnapshot;
 
     std::atomic<int> executingTaskCount = 0;
--- a/include/faabric/scheduler/Scheduler.h
+++ b/include/faabric/scheduler/Scheduler.h
@@ -23,6 +23,8 @@ Scheduler& getScheduler();
 class Executor
 {
   public:
+    std::string id;
+
     explicit Executor(const faabric::Message& msg);
 
     virtual ~Executor();
@@ -34,7 +36,7 @@ class Executor
 
     virtual void flush();
 
-    std::string id;
+    virtual void reset();
 
     virtual int32_t executeTask(
       int threadPoolIdx,
--- a/include/faabric/scheduler/Scheduler.h
+++ b/include/faabric/scheduler/Scheduler.h
@@ -36,12 +36,12 @@ class Executor
 
     std::string id;
 
-  protected:
     virtual int32_t executeTask(
       int threadPoolIdx,
       int msgIdx,
       std::shared_ptr<faabric::BatchExecuteRequest> req);
 
+  protected:
     virtual void restore(const faabric::Message& msg);
 
     virtual void postFinish();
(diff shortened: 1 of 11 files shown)


include/faabric/scheduler/Scheduler.h
```diff
@@ -34,33 +36,38 @@ class Executor
...
     uint32_t threadPoolSize = 0;
-    std::unordered_map<int, std::thread> threadPoolThreads;
-
-    std::unordered_map<
-      int,
-      faabric::util::Queue<
-        std::pair<int, std::shared_ptr<faabric::BatchExecuteRequest>>>>
+    std::vector<std::shared_ptr<std::thread>> threadPoolThreads;
+    std::vector<faabric::util::Queue<
+      std::pair<int, std::shared_ptr<faabric::BatchExecuteRequest>>>>
```
Shillaker, line 70: Avoided the use of maps where possible

include/faabric/scheduler/Scheduler.h
```diff
@@ -160,10 +167,7 @@ class Scheduler
     faabric::util::SystemConfig& conf;
 
     std::unordered_map<std::string, std::vector<std::shared_ptr<Executor>>>
-      executingExecutors;
-
-    std::unordered_map<std::string, std::vector<std::shared_ptr<Executor>>>
-      warmExecutors;
+      executors;
```
Shillaker, line 170: Have a single list of executors

src/runner/FaabricMain.cpp
```diff
@@ -42,9 +42,6 @@ void FaabricMain::startRunner()
     auto& sch = faabric::scheduler::getScheduler();
     sch.addHostToGlobalSet();
 
-    faabric::util::SystemConfig& conf = faabric::util::getSystemConfig();
-    conf.print();
-
```
Shillaker, line 47: This was really bloating the tests, so we can now call `print` in specific runners if we feel it necessary (e.g. in the Faasm worker)

include/faabric/util/queue.h
```diff
@@ -14,14 +14,6 @@ class QueueTimeoutException : public faabric::util::FaabricException
     {}
 };
 
-class ExecutorFinishedException : public faabric::util::FaabricException
-{
-  public:
-    explicit ExecutorFinishedException(std::string message)
-      : FaabricException(std::move(message))
-    {}
-};
```
Shillaker, line 23: This exception type is no longer handled in the scheduler, and only used in `mpi_native`. We can probably find a better way to do this when we come to resurrecting that part of the codebase.

src/scheduler/Executor.cpp
```diff
@@ -10,45 +10,60 @@
...
+    logger->debug("Executor {} shutting down", id);
 
-        killReq->add_messages()->set_type(faabric::Message::KILL);
-        queuePair.second.enqueue(std::make_pair(0, killReq));
-    }
+    // Shut down thread pools and wait
+    for (int i = 0; i < threadPoolThreads.size(); i++) {
+        if (threadPoolThreads.at(i) == nullptr) {
+            continue;
+        }
```
Shillaker, line 42: This bit of code was previously adding a message to every queue regardless of whether there was a thread waiting at the end of it.

src/scheduler/Executor.cpp
```diff
@@ -10,45 +10,60 @@
...
+    // Reset variables
+    boundMessage.Clear();
+    executingTaskCount = 0;
+
+    lastSnapshot = "";
+
+    claimed = false;
+
+    threadPoolThreads.clear();
+    threadQueues.clear();
```
Shillaker, line 66: Previously this method wasn't cleaning up "properly", i.e. it was leaving certain things set that were messing up repeated execution with the same executor.

src/scheduler/Executor.cpp
```diff
@@ -79,99 +119,116 @@ void Executor::executeTasks(std::vector<int> msgIdxs,
...
+            threadPoolIdx = msg.appindex() % threadPoolSize;
+        }
 
         // Enqueue the task
+        logger->trace(
+          "Assigning app index {} to thread {}", msg.appindex(), threadPoolIdx);
         threadQueues[threadPoolIdx].enqueue(std::make_pair(msgIdx, req));
 
         // Lazily create the thread
-        if (threadPoolThreads.count(threadPoolIdx) == 0) {
```
Shillaker, line 88: Not much has changed here, but I moved the guts of the thread to its own function below. This makes it much easier to read and avoids this deeply indented code block.

src/scheduler/Executor.cpp
```diff
@@ -79,99 +119,116 @@ void Executor::executeTasks(std::vector<int> msgIdxs,
...
+            sch.setThreadResult(msg, returnValue);
+        } else {
+            sch.setFunctionResult(msg);
+        }
+
+        // Notify the scheduler, note that we have to release the claim _after_
+        // resetting, once the executor is ready to be reused
+        if (oldTaskCount == 1) {
+            reset(msg);
+            releaseClaim();
```
Shillaker, line 221: This is the only slightly modified part, where we have to release the claim to this executor.

src/scheduler/Scheduler.cpp
```diff
@@ -134,61 +128,23 @@ void Scheduler::vacateSlot()
     thisHostResources.set_usedslots(thisHostResources.usedslots() - 1);
 }
 
-void Scheduler::notifyExecutorFinished(Executor* exec,
```
Shillaker, line 137: This function was basically just doing the logic of switching executors between the two lists, so can go.

src/state/DummyStateServer.cpp
```diff
@@ -83,7 +83,7 @@ void DummyStateServer::start()
     });
 
     // Give it time to start
-    usleep(500 * 1000);
+    usleep(1000 * 1000);
```
Shillaker, line 86: I found some of the state server tests were failing because of this short timeout.

csegarragonz (approved):

src/scheduler/Executor.cpp
```diff
@@ -79,99 +119,116 @@ void Executor::executeTasks(std::vector<int> msgIdxs,
...
     for (int msgIdx : msgIdxs) {
         const faabric::Message& msg = req->messages().at(msgIdx);
-        int threadPoolIdx = msg.appindex() % threadPoolSize;
+
+        // If executing threads, we must always keep thread pool index zero
+        // free, as this may be executing the function that spawned them
+        int threadPoolIdx;
+        if (isThreads) {
+            assert(threadPoolSize > 1);
+            threadPoolIdx = (msg.appindex() % (threadPoolSize - 1)) + 1;
```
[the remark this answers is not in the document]
Shillaker: As discussed offline, each executor will only ever be handling one function, and the threads in this thread pool are only for threads spawned by that function. We rely on the threading implementation to set the `appindex` properly, but there are definitely some potential bugs in here once things get overloaded (although the scheduler will do its best to avoid this).

src/scheduler/Executor.cpp
```diff
@@ -79,99 +119,116 @@ void Executor::executeTasks(std::vector<int> msgIdxs,
...
+
+void Executor::threadPoolThread(int threadPoolIdx)
+{
+    auto logger = faabric::util::getLogger();
+    logger->debug("Thread pool thread {}:{} starting up", id, threadPoolIdx);
+
+    auto& sch = faabric::scheduler::getScheduler();
+    auto& conf = faabric::util::getSystemConfig();
+
+    for (;;) {
```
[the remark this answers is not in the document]
Shillaker: Good point, this used to be a way around some linters which would complain about `while(true)`, so I can't say I have any particular preference (provided it's not a problem with `clang-tidy`). I'll have a look around and see what our current standard is.

src/scheduler/Executor.cpp
```diff
@@ -79,99 +119,116 @@ void Executor::executeTasks(std::vector<int> msgIdxs,
...
+        logger->trace("Thread starting loop {}:{}", id, threadPoolIdx);
+        std::pair<int, std::shared_ptr<faabric::BatchExecuteRequest>> task;
+
+        try {
+            task = threadQueues[threadPoolIdx].dequeue(conf.boundTimeout);
```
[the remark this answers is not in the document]
Shillaker: Yes, this is important for consistent use of threads in applications that care about the thread index (e.g. OpenMP). OpenMP applications need to be able to set a fixed limit on the number of threads, which should lead to queueing when overloaded. If we have a work pool then our system may not replicate the expected behaviour in these overloaded situations (i.e. the work pool may grow beyond the limit set by the framework). In an OpenMP application this limit can also change at runtime, which would be a bit gnarly to manage with a work pool approach as we'd have to support dynamically adding and removing threads.

PR #88, opened 2021-05-13
Distributed tests

Shillaker:
"Proper" distributed tests that will run on multiple containers via `docker-compose`.

- `dist-test` directory holding `docker-compose` file and scripts to run in the containers.
- Support for running locally and developing, and in CI through the same scriptse.
- Separate folder of specifically distributed tests at `tests/dist`.
- Added three tests: one to check the list of hosts is as expected, one to do remote thread execution, one to do remote function chaining.
- Added docs on how to use, starting the `docs/development.md` file (which can be added to with a lot of the `README` content).

There's quite a lot of scaffolding around the tests themselves. The idea is:

- They always use the `DistTestExecutor`.
- Each test can register its own callback function to execute messages (see the callbacks registered for `threads/simple` and `funcs/simple`.
- These callbacks don't all have to live in a single file, they can be put alongside the tests for that functionality (e.g. `tests/dist/scheduler/functions.cpp`
- The `DistTestFixture` handles the set-up and tear-down of the distributed test environment. This pattern can be repeated in future in other areas of the tests as discussed.

Finally note that there's a special container in the `docker-compose.yml` just for doing the build (`builder`). This is to get the benefits of the build cached in the base container, but avoid having to run it twice, once in the master and once in the worker.


.github/workflows/tests.yml
```diff
@@ -117,18 +130,11 @@ jobs:
   mpi_native:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
-    env:
-      HOST_TYPE: ci
-      REDIS_QUEUE_HOST: redis
-      REDIS_STATE_HOST: redis
-    services:
-      redis:
-        image: redis
```
Shillaker, line 126: This `redis` service wasn't used as everything for the MPI native tests runs through docker-compose (including another Redis instance)

docs/development.md
```diff
@@ -0,0 +1,33 @@
+# Developing Faabric
```
Shillaker, line 1: We can add most of the `README` content into this `development.md` file as it's quite verbose now. Then the README can just be a short bit of blurb on what the project is with a few links.

csegarragonz (approved):
LGTM. Nice to have these in.

PR #90, opened 2021-05-19
Shutdown issue fix

Shillaker:
Three problems fixed in here:

- When thread pool thread `0` finished in the executor, the reference to it was being removed, so it couldn't be joined, and hence when running quickly, would cause `terminate caused without an active exception`-like errors.
- When an executor was removed from the scheduler, the same thing was happening; the executor thread pool would finish, the reference would be removed, then the scheduler couldn't clear it up.
- The executor removal code wasn't actually working, calling `remove_if` but not calling `erase`.

Because all of this shutdown happens _inside_ one of the worker threads of the executor, we can't fully kill the executor without a deadlock (as it must also await the worker thread itself). As a result, I've added a bit of a hack which is to have a list of `deadExecutors` and `deadThreads` which will also get tidied up at the end.

### DIFF
--- a/src/scheduler/Executor.cpp
+++ b/src/scheduler/Executor.cpp
@@ -53,6 +53,13 @@ void Executor::finish()
         }
     }
 
+    // Await dead threads
+    for (auto dt : deadThreads) {
+        if (dt->joinable()) {
+            dt->join();
+        }
+    }
+
     // Hook
     this->postFinish();
 
@@ -153,6 +160,8 @@ void Executor::threadPoolThread(int threadPoolIdx)
     auto& sch = faabric::scheduler::getScheduler();
     auto& conf = faabric::util::getSystemConfig();
 
+    bool selfShutdown = false;
+
     for (;;) {
         logger->trace("Thread starting loop {}:{}", id, threadPoolIdx);
         std::pair<int, std::shared_ptr<faabric::BatchExecuteRequest>> task;
@@ -162,7 +171,11 @@ void Executor::threadPoolThread(int threadPoolIdx)
         } catch (faabric::util::QueueTimeoutException& ex) {
             // If the thread has had no messages, it needs to
             // remove itself
-            shutdownThreadPoolThread(threadPoolIdx);
+            logger->trace("Thread {}:{} got no messages in timeout {}ms",
+                          id,
+                          threadPoolIdx,
+                          conf.boundTimeout);
+            selfShutdown = true;
             break;
         }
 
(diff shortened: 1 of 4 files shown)


src/scheduler/Executor.cpp
```diff
@@ -223,19 +237,33 @@ void Executor::threadPoolThread(int threadPoolIdx)
...
+        std::shared_ptr<std::thread> thisThread =
+          threadPoolThreads.at(threadPoolIdx);
+        deadThreads.emplace_back(thisThread);
+
+        // Set this thread to nullptr
+        threadPoolThreads.at(threadPoolIdx) = nullptr;
+
+        // See if any threads are still running
+        bool isFinished = true;
+        for (auto t : threadPoolThreads) {
```
Shillaker, line 256: I would love not to iterate through this vector checking equality, but all the stdlib options look pretty verbose, and this isn't on the critical path performance-wise so I'd rather stick with readability and debuggability).

src/scheduler/Executor.cpp
```diff
@@ -223,19 +237,33 @@ void Executor::threadPoolThread(int threadPoolIdx)
...
+            if (t != nullptr) {
+                isFinished = false;
+                break;
+            }
+        }
```
Shillaker, line 261: I tried a few different stdlib options for doing this check, including `find_if`, `find_not_if` etc. but they all came out way more verbose and unreadable than this, so given that this isn't on the critical path performance-wise I'm inclined to leave the simple loop.

src/scheduler/Scheduler.cpp
```diff
@@ -137,14 +142,29 @@ void Scheduler::notifyExecutorShutdown(Executor* exec,
...
     std::vector<std::shared_ptr<Executor>>& thisExecutors = executors[funcStr];
-    std::remove_if(thisExecutors.begin(),
-                   thisExecutors.end(),
-                   [&execId](const auto& p) { return p->id == execId; });
+    for (int i = 0; i < thisExecutors.size(); i++) {
+        if (thisExecutors.at(i)->id == exec->id) {
+            execIdx = i;
+            break;
+        }
+    }
```
Shillaker, line 153: Ditto re. trying to use the stdlib for this, it came out really ugly and I think this is more readable (and editable).

src/scheduler/Executor.cpp
```diff
@@ -53,6 +53,13 @@ void Executor::finish()
         }
     }
 
+    // Await dead threads
+    for (auto dt : deadThreads) {
```
csegarragonz, line 57: Just in case, should we use `auto&` here?

csegarragonz (approved):

src/scheduler/Executor.cpp
Shillaker: Given that it's a vector of pointers I think it's ok.

PR #96, opened 2021-05-21
Dirty page checking and snapshot diffs

Shillaker:
The aim of this PR is:

- Return diffs from the workers back to the master after executing a batch of threads. This means tracking dirty pages on worker hosts when executing a batch of threads, then returning the resulting memory diff back to the master when finished.
- Push diffs from the master host to workers when executing subsequent batches of threads (i.e. push the whole snapshot to the host when executing the first batch, then only push diffs after that). This means tracking dirty pages on the master _between_ executing batches of threads.

Changes:

- Add pushing diffs between hosts. This is either pushing diffs on their own (master -> worker), or pushing diffs along with a thread result (worker -> master). See notes below on this.
- Add dirty page checking based on [soft-dirty PTEs](https://www.kernel.org/doc/html/v5.4/admin-guide/mm/soft-dirty.html).
- Add `snapshot()` method to executor subclasses so that Faabric can request a pointer to their memory (which it then uses to work out dirty pages).
- Convert both snapshot push operations (full and diffs) to await a response synchronous, as the master host needs to know the snapshot data has been pushed and set up before sending function messages.
- Add test fixtures where possible to reduce verbosity of tests.
- Made `logger` an instance variable where possible to avoid calling `getLogger` all over the place.

Because the diffs from child threads have to be applied before other threads can continue, we have to make sure that they are applied _before_ the thread result is returned (as the thread result is the thing that other threads wait on). Therefore we have to piggyback the diffs on the existing thread result message.

I've ported the thread result message to use flatbuffers which is where all the other snapshot stuff lives. For now I've put it in with the existing `SnapshotClient` and `SnapshotServer` in an attempt to minimise the number of changes in this PR, but eventually we'll need to commit to either protobuf or flatbuffers. Flatbuffers seem potentially more efficient, but the awkwardness of passing them around and accessing the data within is a major downside.

### DIFF
--- a/dist-test/run.sh
+++ b/dist-test/run.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -e
-
 export PROJ_ROOT=$(dirname $(dirname $(readlink -f $0)))
 pushd ${PROJ_ROOT}/dist-test >> /dev/null
 
@@ -10,6 +8,8 @@ if [[ -z "${FAABRIC_CLI_IMAGE}" ]]; then
     VERSION=$(cat ../VERSION)
     export FAABRIC_CLI_IMAGE=faasm/faabric:${VERSION}
 fi
+
+RETURN_VAL=0
 
 if [ "$1" == "local" ]; then
     INNER_SHELL=${SHELL:-"/bin/bash"}
@@ -32,9 +32,16 @@ else
         run \
         master \
         /build/faabric/static/bin/faabric_dist_tests
+    RETURN_VAL=$?
 
-    docker-compose \
-        stop
+    echo "-------------------------------------------"
+    echo "                WORKER LOGS                "
+    echo "-------------------------------------------"
+    docker-compose logs worker
+
+    docker-compose stop
 fi
 
 popd >> /dev/null
+
+exit $RETURN_VAL
--- a/src/proto/faabric.proto
+++ b/src/proto/faabric.proto
@@ -50,11 +50,6 @@ message UnregisterRequest {
     Message function = 2;
 }
 
-message ThreadResultRequest {
-    int32 messageId = 1;
-    int32 returnValue = 2;
-}
-
 message FunctionStatusResponse {
     enum FunctionStatus {
         OK = 0;
--- a/src/proto/faabric.proto
+++ b/src/proto/faabric.proto
@@ -27,7 +27,6 @@ message BatchExecuteRequest {
 
     // Shared snapshot used for threads
     string snapshotKey = 3;
-    int32 snapshotSize = 4;
 
     repeated Message messages = 5;
 
@@ -109,7 +108,6 @@ message Message {
     int32 returnValue = 11;
 
     string snapshotKey = 12;
-    int32 snapshotSize = 13;
 
     int64 timestamp = 14;
     string resultKey = 15;
--- a/src/scheduler/Scheduler.cpp
+++ b/src/scheduler/Scheduler.cpp
@@ -588,15 +588,22 @@ SnapshotClient& Scheduler::getSnapshotClient(const std::string& otherHost)
     auto it = snapshotClients.find(key);
     if (it == snapshotClients.end()) {
         faabric::util::UniqueLock lock(snapshotClientsMutex);
-        auto _it = snapshotClients.try_emplace(key, otherHost);
-        if (!_it.second) {
-            logger->error("Could not add snapshot client key {} for host {}",
-                          key,
-                          otherHost);
-            throw std::runtime_error("Error inserting snapshot client");
+
+        auto it = snapshotClients.find(key);
+        if (it == snapshotClients.end()) {
+
+            auto _it = snapshotClients.try_emplace(key, otherHost);
+            if (!_it.second) {
+                logger->error(
+                  "Could not add snapshot client key {} for host {}",
+                  key,
+                  otherHost);
+                throw std::runtime_error("Error inserting snapshot client");
+            }
+            it = _it.first;
         }
-        it = _it.first;
     }
+
     return it->second;
 }
 
--- a/src/scheduler/Scheduler.cpp
+++ b/src/scheduler/Scheduler.cpp
@@ -579,7 +579,7 @@ FunctionCallClient& Scheduler::getFunctionCallClient(
 SnapshotClient& Scheduler::getSnapshotClient(const std::string& otherHost)
 {
     // Note, our keys here have to include the tid as the clients can only be
-    // used within the same thread
+    // used within the same thread.
     std::thread::id tid = std::this_thread::get_id();
     std::stringstream ss;
     ss << otherHost << "_" << tid;
@@ -587,6 +587,7 @@ SnapshotClient& Scheduler::getSnapshotClient(const std::string& otherHost)
 
     auto it = snapshotClients.find(key);
     if (it == snapshotClients.end()) {
+        faabric::util::UniqueLock lock(snapshotClientsMutex);
         auto _it = snapshotClients.try_emplace(key, otherHost);
         if (!_it.second) {
             logger->error("Could not add snapshot client key {} for host {}",
--- a/src/scheduler/Scheduler.cpp
+++ b/src/scheduler/Scheduler.cpp
@@ -276,6 +276,10 @@ std::vector<std::string> Scheduler::callFunctions(
             // Do the snapshot diff pushing
             if (!snapshotDiffs.empty()) {
                 for (const auto& h : thisRegisteredHosts) {
+                    logger->debug("Pushing {} snapshot diffs for {} to {}",
+                                  snapshotDiffs.size(),
+                                  funcStr,
+                                  h);
                     SnapshotClient& c = getSnapshotClient(h);
                     c.pushSnapshotDiffs(snapshotKey, snapshotDiffs);
                 }
@@ -284,6 +288,8 @@ std::vector<std::string> Scheduler::callFunctions(
             // Now reset the dirty page tracking, as we want the next batch of
             // diffs to contain everything from now on (including the updates
             // sent back from all the threads)
+            logger->debug("Restting dirty tracking after pushing diffs {}",
+                          funcStr);
             faabric::util::resetDirtyTracking();
         }
 
@@ -525,7 +531,7 @@ int Scheduler::scheduleFunctionsOnHost(
     }
 
     logger->debug(
-      "Sending {}/{} {} to {}", available, nMessages, funcStr, host);
+      "Sending {}/{} {} to {}", nOnThisHost, nMessages, funcStr, host);
 
     getFunctionCallClient(host).executeFunctions(hostRequest);
 
@@ -723,29 +729,19 @@ void Scheduler::setThreadResult(
     bool isMaster = msg.masterhost() == conf.endpointHost;
 
     if (isMaster) {
-        setThreadResult(msg.id(), returnValue);
+        setThreadResultLocally(msg.id(), returnValue);
     } else {
         SnapshotClient& c = getSnapshotClient(msg.masterhost());
 
         if (diffs.empty()) {
-            logger->debug("Sending thread result {} for {} to {}",
-                          returnValue,
-                          msg.id(),
-                          msg.masterhost());
             c.pushThreadResult(msg.id(), returnValue);
-
         } else {
-            logger->debug("Sending thread result {} for {} to {} plus {} diffs",
-                          returnValue,
-                          msg.id(),
-                          msg.masterhost(),
-                          diffs.size());
             c.pushThreadResult(msg.id(), returnValue, msg.snapshotkey(), diffs);
         }
     }
 }
 
-void Scheduler::setThreadResult(uint32_t msgId, int32_t returnValue)
+void Scheduler::setThreadResultLocally(uint32_t msgId, int32_t returnValue)
 {
     logger->debug("Setting result for thread {} to {}", msgId, returnValue);
     threadResults[msgId].set_value(returnValue);
--- a/src/scheduler/Scheduler.cpp
+++ b/src/scheduler/Scheduler.cpp
@@ -1,4 +1,3 @@
-#include "faabric/util/memory.h"
 #include <faabric/proto/faabric.pb.h>
 #include <faabric/redis/Redis.h>
 #include <faabric/scheduler/ExecutorFactory.h>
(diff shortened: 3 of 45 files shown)


src/scheduler/Scheduler.cpp
```diff
@@ -549,6 +582,29 @@ FunctionCallClient& Scheduler::getFunctionCallClient(
     return it->second;
 }
 
+SnapshotClient& Scheduler::getSnapshotClient(const std::string& otherHost)
+{
+    // Note, our keys here have to include the tid as the clients can only be
+    // used within the same thread
```
Shillaker, line 588: @csegarragonz I don't know how this isn't failing for the function call clients, as the `functionCallClients` map is shared by all threads and keyed on the other hostname, so I would imagine the same client is getting used by different threads. However, it seems there aren't any assertions failing, so I guess that isn't true...?

tests/utils/fixtures.h
```diff
@@ -0,0 +1,54 @@
+#pragma once
+
+#include "faabric/util/memory.h"
+#include <faabric/redis/Redis.h>
+#include <faabric/scheduler/Scheduler.h>
+#include <faabric/util/testing.h>
+
+namespace tests {
+class BaseTestFixture
```
Shillaker, line 9: This `BaseTestFixture` is here to capture the stuff I found we were repeating in all the existing features. We have to be careful not to make this a dumping ground, and only included the tidy-up that's actually going to get repeated elsewhere.

src/util/snapshot.cpp
```diff
@@ -0,0 +1,30 @@
...
+        std::vector<faabric::util::SnapshotDiff> empty;
+        return empty;
+    }
+
+    // Get dirty pages
+    int nPages = getRequiredHostPages(size);
+    std::vector<bool> dirtyFlags = faabric::util::getDirtyPages(data, nPages);
+
+    // Convert to snapshot diffs
+    // TODO - reduce number of diffs by merging adjacent dirty pages
```
csegarragonz, line 18: Could we tackle the `TODO` in this PR? I feel it should not be very complex to implement. I was thinking of something like:
```cpp
auto next1 = std::find(dirtyFlags.begin(), dirtyFlags.end(), 1);
assert(next != dirtyFlags.end());
auto next0 = [&](){ return std::find(next1, dirtyFlags.end(), 0); };
while (next0() != dirtyFlags.end()) {
    // May have to +/- 1
    auto size = next0() - next1;
    uint32_t offset = next1 * faabric::util::HOST_PAGE_SIZE;
    diffs.emplace_back(offset, data + offset,  size * faabric::util::HOST_PAGE_SIZE;
    next1 = std::find(next0, dirtyFlags.end(), 1);
}
// handle the last chunk (i.e. when the vector ends with ones)
```
there's provably a nicer way, and the code above will be full of bugs, but just an idea.

tests/utils/fixtures.h
csegarragonz: I personally doubt that this is useful/desirable. The main point of fixtures was to remove the dumping ground `cleanFaabric()` had turned into. However, I feel we have just moved the burden of it to `BaseTestFixture`. 

For instance, why would the transport tests `redis.flushAll()`? I think there's value in knowing what you are modifying in your tests, and resetting just that.

src/scheduler/Scheduler.cpp
csegarragonz: It is not failing as we do the assertions, but I'm indeed surprised that it is not. Is your approach thread-safe either? I see that different threads would only ever access different keys, but I am wondering if this won't leave the unerlying data structure in an undefined state.

I guess I assumed the scheduler was single-threaded. In MPI we use TLS for the clients, but it's painful to clean up (as it must be explicitely cleaned, from the same thread which opened it). Happy to discuss this offline.

csegarragonz (changes requested):
Left some comments, overall LGTM, and the dirty page tracking looks great and neat. I know you started this _before_ we agreed on the new PR+commit structure, but for the future, 15 hundred lines added with four commits is very hard to review in a per-commit fashion (which btw you can do using `n` and `p` in the browser).

Wrt the PR size/topics I feel that all the fixture changes had really nothing to do with dirty page checking, so this could definately be in a different PR.

src/util/snapshot.cpp
Shillaker: I think this may be premature optimisation as I'm not sure how often this would benefit (and will add yet more complexity to something that's already quite complicated). I'll revisit this TODO once I've done some benchmarking.

src/scheduler/Scheduler.cpp
```diff
@@ -246,9 +260,31 @@ std::vector<std::string> Scheduler::callFunctions(
...
+            snapshotDiffs = snapshotData.getDirtyPages();
+
+            // Do the snapshot diff pushing
+            if (!snapshotDiffs.empty()) {
+                for (const auto& h : thisRegisteredHosts) {
+                    logger->debug("Pushing {} snapshot diffs for {} to {}",
+                                  snapshotDiffs.size(),
+                                  funcStr,
+                                  h);
+                    SnapshotClient& c = getSnapshotClient(h);
```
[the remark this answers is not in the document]
Shillaker: This doesn't seem like a big deal to me, although they are both functions to get clients they're not actually related in any other way, so I'm not sure consistency matters here. Are you saying there should be a rule that if we call a "get something" on one line, then it's only used to call a single method on the next line, that the "get something" should be inlined with the method call on the next line? There's probably not much consistency on that throughout the codebase.

src/scheduler/Scheduler.cpp
Shillaker: The scheduler is not single-threaded, it's a singleton object that gets called by lots of threads, so thread safety is still very important. The confusion may have arisen because the scheduler used to manage its own thread pool but now doesn't actually spawn any threads (the executors do that), this does not mean it's only ever run in a single thread though.

You're right on the map accesses, there should be a lock around this as the map itself isn't thread safe.

src/scheduler/SnapshotClient.cpp
```diff
@@ -48,37 +82,92 @@ void SnapshotClient::sendHeader(faabric::scheduler::SnapshotCalls call)
...
 
     if (faabric::util::isMockMode()) {
-        snapshotPushes.emplace_back(host, req);
+        faabric::util::UniqueLock lock(mockMutex);
+        snapshotPushes.emplace_back(host, data);
     } else {
         // Send the header first
         sendHeader(faabric::scheduler::SnapshotCalls::PushSnapshot);
 
+        faabric::util::SystemConfig& conf = faabric::util::getSystemConfig();
```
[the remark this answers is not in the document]
Shillaker: Yep good point on the `const`, but even without `const`, requesting a reference avoids a copy.

src/scheduler/SnapshotClient.cpp
```diff
@@ -48,37 +82,92 @@ void SnapshotClient::sendHeader(faabric::scheduler::SnapshotCalls call)
...
         auto keyOffset = mb.CreateString(key);
-        auto dataOffset = mb.CreateVector<uint8_t>(req.data, req.size);
-        auto requestOffset =
-          CreateSnapshotPushRequest(mb, keyOffset, dataOffset);
+        auto dataOffset = mb.CreateVector<uint8_t>(data.data, data.size);
+        auto requestOffset = CreateSnapshotPushRequest(
+          mb, returnHostOffset, keyOffset, dataOffset);
 
+        // Send the data
         mb.Finish(requestOffset);
```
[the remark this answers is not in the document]
Shillaker: Yep good spot.

src/scheduler/SnapshotClient.cpp
```diff
@@ -96,4 +185,54 @@ void SnapshotClient::deleteSnapshot(const std::string& key)
...
+        std::vector<flatbuffers::Offset<SnapshotDiffChunk>> diffsFbVector;
+        for (const auto& d : diffs) {
+            auto dataOffset = mb.CreateVector<uint8_t>(d.data, d.size);
+            auto chunk = CreateSnapshotDiffChunk(mb, d.offset, dataOffset);
+            diffsFbVector.push_back(chunk);
+        }
+
+        auto keyOffset = mb.CreateString(snapshotKey);
+        auto diffsOffset = mb.CreateVector(diffsFbVector);
+        auto requestOffset = CreateThreadResultRequest(
```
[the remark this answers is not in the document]
Shillaker: It's definitely close and I did try it, but it didn't add much and the method took quite a few parameters.

src/scheduler/SnapshotServer.cpp
```diff
@@ -50,8 +60,74 @@ void SnapshotServer::recvPushSnapshot(faabric::transport::Message& msg)
...
+
+void SnapshotServer::recvThreadResult(faabric::transport::Message& msg)
+{
+    const ThreadResultRequest* r =
+      flatbuffers::GetMutableRoot<ThreadResultRequest>(msg.udata());
+
+    // Apply snapshot diffs *first* (these must be applied before other threads
+    // can continue)
+    if (r->chunks()->size() > 0) {
+        faabric::util::getLogger()->info("Receiving {} diffs to snapshot {}",
```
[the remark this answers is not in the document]
Shillaker: Yes good point, no idea why these are `info`s

src/util/memory.cpp
```diff
@@ -57,4 +74,73 @@ AlignedChunk getPageAlignedChunk(long offset, long length)
...
+    if (fd == nullptr) {
+        throw std::runtime_error("Could not open clear_refs");
+    }
+
+    // Write 4 to the file to track from now on
+    // https://www.kernel.org/doc/html/v5.4/admin-guide/mm/soft-dirty.html
+    char value[] = "4";
+    size_t nWritten = fwrite(value, sizeof(char), 1, fd);
+    if (nWritten != 1) {
+        throw std::runtime_error("Failed to write to clear_refs");
```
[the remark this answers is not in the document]
Shillaker: Both the exceptions in here will only happen if there's a permissions error on the file itself, so it will only happen if we've misconfigured an environment, and this exception will be triggered from both the main thread and child threads. A log line could be useful though.

tests/utils/fixtures.h
Shillaker: Yes, I did think that when creating this, the thing is with Redis and the scheduler for example, the snippets of clean-up would have to be sprinkled around so many places that it makes sense to avoid repetition. 

The solution I think should be a fixture per feature and have multiple inheritance in the tests, i.e. we have a `RedisFixture`, a `StateFixture`, a `MockClientsFixture` etc. I'll see if that works here...

Shillaker:
> Left some comments, overall LGTM, and the dirty page tracking looks great and neat. I know you started this _before_ we agreed on the new PR+commit structure, but for the future, 15 hundred lines added with four commits is very hard to review in a per-commit fashion (which btw you can do using `n` and `p` in the browser).

Yes let's not be too strict on this one, I think we can aim for the `n`/`p` reviewing in an ideal world but not worry too much if we don't achieve it every time. In this case it was a choice between 50 commits with things like "fixing tests" repeated over and over and one big one, neither of which is useful. I could potentially have rebased into something slightly more meaningful, but I think you have to start the work in that mindset for it to really work.
 
> Wrt the PR size/topics I feel that all the fixture changes had really nothing to do with dirty page checking, so this could definately be in a different PR.

Yes, again in an ideal world, but let's not let great be the enemy of the good. "nothing to do" is a little strong, it avoided repeating code in the new tests I was writing.

dist-test/run.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -e
-
```
Shillaker, line 4: To handle failures in the worker and still be able to print logs we have to switch off `set -e`

src/proto/faabric.proto
```diff
@@ -114,7 +108,6 @@ message Message {
     int32 returnValue = 11;
 
     string snapshotKey = 12;
-    int32 snapshotSize = 13;
```
Shillaker, line 117: Neither this nor the `snapshotSize` variable above were actually used. I've deleted them but not updated the numbers of the subsequent fields to save on the diff size.

src/scheduler/Scheduler.cpp
```diff
@@ -535,18 +561,54 @@ std::vector<faabric::Message> Scheduler::getRecordedMessagesLocal()
...
+            snapshotClients.emplace(key, otherHost);
         }
-        it = _it.first;
     }
-    return it->second;
+
+    {
+        faabric::util::SharedLock lock(snapshotClientsMx);
+        return snapshotClients.at(key);
+    }
```
Shillaker, line 611: These were the changes to `getSnapshotClient` and `getFunctionCallClient` that address the issues of cross-thread clients and race conditions.

src/scheduler/SnapshotClient.cpp
```diff
@@ -48,37 +82,92 @@ void SnapshotClient::sendHeader(faabric::scheduler::SnapshotCalls call)
...
+          mb, returnHostOffset, keyOffset, dataOffset);
 
+        // Send the data
         mb.Finish(requestOffset);
         uint8_t* msg = mb.GetBufferPointer();
         int size = mb.GetSize();
         send(msg, size);
+
+        // Await a response as this call must be synchronous
+        awaitResponse(SNAPSHOT_PORT + REPLY_PORT_OFFSET);
```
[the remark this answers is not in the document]
Shillaker: Yeah i think doing it automatically inside the method makes sense. The only downside is that it adds a bit of magic that may not be obvious to the caller, but I think it's worth it. Let's do in another PR though to try and keep a lid on the size of this one.

src/util/logging.cpp
```diff
@@ -51,10 +54,16 @@ static void initLogging(const std::string& name)
...
+        if (loggers.find(name) == loggers.end()) {
+            initLogging(name);
+        }
     }
 
-    return loggers[name];
+    {
+        faabric::util::SharedLock lock(loggerMx);
+        return loggers[name];
+    }
```
Shillaker, line 67: Same race condition issue mentioned previously. Given the proposed move to using spdlog macros we don't need to worry too much about the performance hit this will have (which will probably be negligible anyway unless called in a very tight loop)

tests/dist/DistTestExecutor.h
```diff
@@ -27,6 +28,14 @@ class DistTestExecutor final : public faabric::scheduler::Executor
       int threadPoolIdx,
       int msgIdx,
       std::shared_ptr<faabric::BatchExecuteRequest> req) override;
+
+    faabric::util::SnapshotData snapshot() override;
+
+    uint8_t* snapshotMemory = nullptr;
+    size_t snapshotSize = 0;
```
Shillaker, line 35: Originally I hadn't set defaults on the `snapshotMemory` and `snapshotSize`, which worked on my machine as the size was getting set to zero, but in Github Actions it was getting set to a non-zero value which was breaking everything. The take-away is that we must always set default values on class members. I looked for a clang-tidy rule to enforce this but couldn't find it.

csegarragonz (approved):

PR #102, opened 2021-06-04
Introduce MPI-specific p2p message endpoint

csegarragonz:
It seems that there is a lot of contention in MPI applications that stress faabric. Profiling seems to indicate that this may be due to handling many threads all around the place. In this PR I attempt to alleviate some of this, by introducing p2p (rank to rank) communication between MPI processes.

Ideally, this will reduce the number of sleep-awake pattern by:
* Not relying on the blocking, main `recv` loop (running in a different socket).
* Use specific `(host, port)` pairs per rank.
* Not wait on the `localQueue` (plus accquire a lock to fetch it, which in turn acquires another lock in the actual queue, which already relies on the internal zeromq queueing).


csegarragonz:
Superseeded by #105

PR #103, opened 2021-06-04
State-less MPI rank management

csegarragonz:
In this PR I remove the dependency with `state` for the rank-host map bookkeeping. This was something I had in my to-do list as it really wasn't needed. Even though this does not have a direct impact on performance, it opens the door to rank-to-rank communication, I elaborate.

Before, upon `MPI_Init` all ranks would halt at a barrier. Non-zero ranks would `recv` a broadcast message from the root rank (0), and bind to their local queue (not knowing which host the message came from). Only when trying to send a response afterwards, they'd query `redis` for the host where the root rank was placed. If we were to query `redis` before the first `recv` (to, for instance, `recv` on a per-rank port) we would be racing the root rank and its update of the rank-host map.

Given the socket-like tools introduced with zeromq, we can just send the map around to the hosts that need it. This is greatly facilitated by the fact that the scheduler returns the list of hosts where it has scheduled the calls.

The only part of the API that now needs the state are the RMA calls, which I haven't worked with at all, but quite keen to do so when they are needed.

RIP `registerRank` :skull:


include/faabric/scheduler/MpiContext.h
```diff
@@ -9,7 +9,7 @@ class MpiContext
   public:
     MpiContext();
 
-    void createWorld(const faabric::Message& msg);
+    void createWorld(faabric::Message& msg);
```
csegarragonz, line 12: See `MpiContext.cpp` for an explanation of this.

src/scheduler/MpiContext.cpp
```diff
@@ -25,6 +25,9 @@ void MpiContext::createWorld(const faabric::Message& msg)
     worldId = (int)faabric::util::generateGid();
     logger->debug("Initialising world {}", worldId);
 
+    // Update the original message to contain the world ID
+    msg.set_mpiworldid(worldId);
```
csegarragonz, line 29: This is needed so all `getExecutingWorld` from the master rank in `faasm`, call `MpiWorldRegistry::getOrInitialiseWorld` with the updated `worldId`.

include/faabric/transport/MpiMessageEndpoint.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
```
csegarragonz, line 1: Currently, the implementation of these is so simple, that they are alright with just being abstract methods. However, I thought they belonged in `transport` rather than `scheduler`.

src/proto/faabric.proto
```diff
@@ -84,6 +84,10 @@ message MPIMessage {
     bytes buffer = 8;
 }
 
+message MpiHostRankMsg {
```
csegarragonz, line 87: Note that `protobuf` [guarantees in-order delivery for `repeated` fields](https://developers.google.com/protocol-buffers/docs/proto3#specifying_field_rules).

tests/test/scheduler/test_function_client_server.cpp
```diff
@@ -53,32 +53,51 @@ class ClientServerFixture
 
 TEST_CASE_METHOD(ClientServerFixture, "Test sending MPI message", "[scheduler]")
 {
-    // Create an MPI world on this host and one on a "remote" host
-    std::string otherHost = "192.168.9.2";
+    auto& sch = faabric::scheduler::getScheduler();
```
csegarragonz, line 56: Removing `registerRank` and having everything be handled by the bootstrapping process has the downside that we now need to do a better job at faking the MPI environment in tests to ensure the expected rank allocation.

tests/test/scheduler/test_mpi_context.cpp
```diff
@@ -131,8 +132,9 @@ TEST_CASE("Check joining world", "[mpi]")
 
     // Check rank is registered to this host
     MpiWorldRegistry& reg = getMpiWorldRegistry();
-    MpiWorld& world = reg.getOrInitialiseWorld(msgB, worldId);
+    MpiWorld& world = reg.getOrInitialiseWorld(msgB);
     const std::string actualHost = world.getHostForRank(1);
+    REQUIRE(actualHost == expectedHost);
```
csegarragonz, line 137: We weren't checking anything here.

src/scheduler/MpiWorld.cpp
```diff
@@ -133,9 +100,26 @@ void MpiWorld::create(const faabric::Message& call, int newId, int newSize)
         msg.set_ismpi(true);
         msg.set_mpiworldid(id);
         msg.set_mpirank(i + 1);
+        msg.set_mpiworldsize(size);
     }
 
-    sch.callFunctions(req);
+    // The scheduler returns at which host each message is scheduled
```
Shillaker, line 106: The comments over the next few lines are a little hard to parse as they don't necessarily correspond to the line below and the code that builds the list of recipients is interleaved with establishing the list of ranks. 

Also remember you don't need to add markdown syntax to comments (i.e. don't need the backticks around variables), and in general it's best to avoid including variable names in comments (e.g. `rankHostMap`) as they may not get changed if someone changes the variable name.

Finally does the line setting `mutable_hosts` avoid a copy ( `*hostRankMsg.mutable_hosts() = { executedAt.begin(), executedAt.end() };`)? If not I think we usually use the syntax of `hostRankMsg.set_hosts(executedAt)` (which is a bit clearer). 

To restructure I'd suggest:

```
// Send the init messages (note that message i corresponds to rank i+1)
std::vector<std::string> executedAt = sch.callFunctions(req);

// Prepend this host for rank 0
executedAt.insert(executedAt.begin(), thisHost);

// Register host to rank mapping on this host
setHostForRank(executedAt);

// Set up list of hosts to broadcast to (excluding this host)
std::set<std::string> hosts(executedAt.begin(), executedAt.end());
hosts.erase(thisHost);

// Do the broadcast
faabric::MpiHostRankMsg ranksMsg;
ranksMsg.set_hosts(executedAt);
for (const auto& h : hosts) {
    faabric::transport::sendMpiHostRankMsg(h, hostRankMsg);
}
```

src/scheduler/MpiWorld.cpp
```diff
@@ -193,77 +171,36 @@ void MpiWorld::closeThreadLocalClients()
...
+    // _same_ host we have created one (note that this would never happen in
+    // reality). If so, we skip the rank-host map broadcasting.
+    if (!forceLocal) {
+        // Block until we receive
+        faabric::MpiHostRankMsg hostRankMsg =
+          faabric::transport::recvMpiHostRankMsg();
+        std::vector<std::string> rankHostVec(hostRankMsg.hosts_size());
+        for (int i = 0; i < rankHostVec.size(); i++) {
+            rankHostVec.at(i) = hostRankMsg.hosts(i);
+        }
```
Shillaker, line 194: Can we just use the copy constructor here and avoid the loop, i.e. 

```
std::vector<std::string> rankHostVec = hostRankMsg.hosts()
```

Or even better, can't we just pass `hostRankMsg.hosts()` straight into `setHostForRank`?

If the copying is necessary, might be good to add a comment explaining why.

src/scheduler/MpiWorld.cpp
```diff
@@ -272,6 +209,22 @@ std::string MpiWorld::getHostForRank(int rank)
...
+{
+    assert(rankHostVec.size() == size);
+    faabric::util::FullLock lock(worldMutex);
+    for (int i = 0; i < size; i++) {
+        auto it = rankHostMap.try_emplace(i, rankHostVec.at(i));
+        if (!it.second) {
+            logger->error("Error emplacing in rankHostMap: {} -> {}",
+                          i + 1,
+                          rankHostVec.at(i));
+            throw std::runtime_error("Error emplacing in rankHostMap");
```
Shillaker, line 223: I'm not sure I understand the use of this pattern with `try_emplace` instead of just `emplace`. What situation is it guarding against? It's adding 5 lines of error handling code, and even when it fails the message isn't particularly clear. Much like avoiding variable names in comments, it's not good to include them in error messages, as this will be even more confusing if the variable name changes.

tests/test/scheduler/test_function_client_server.cpp
```diff
@@ -53,32 +53,50 @@ class ClientServerFixture
...
+
+    // Force the scheduler to initialise a world in the remote host by setting
+    // a worldSize bigger than the slots available locally
```
Shillaker, line 59: Again, very small nitpicky thing but having variable names in comments not ideal. Here it can just be "world size" rather than "worldSize"

tests/test/scheduler/test_function_client_server.cpp
Shillaker: Yes this is a bit of a hassle but doesn't look _too_ bad. Ideally we'd move these sort of tests to "proper" distributed tests, but this will do for now.

tests/test/scheduler/test_mpi_world.cpp
```diff
@@ -1051,16 +1058,13 @@ template void doReduceTest<double>(scheduler::MpiWorld& world,
 
 TEST_CASE("Test reduce", "[mpi]")
 {
+    cleanFaabric();
```
Shillaker, line 1061: Would be nice to add a fixture to this file as we also have a custom `tearDown` method. One for another PR probably.

Shillaker (changes requested):
This is great, the use of state for MPI has always been very fiddly. 

A few style/ naming/ structure comments but nothing fundamental.

include/faabric/transport/common.h
```diff
@@ -8,3 +8,5 @@
 #define MPI_MESSAGE_PORT 8005
 #define SNAPSHOT_PORT 8006
 #define REPLY_PORT_OFFSET 100
+
+#define MPI_PORT 8800
```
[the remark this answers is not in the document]
csegarragonz: After our offline discussion, I'll leave the port like this as it will change in coming PRs.

src/scheduler/MpiWorld.cpp
csegarragonz: Ok to everything besides the `set_hosts()` bit. Hosts is a `repeated` field (somewhat like a vector) and, using the `set_hosts` syntax, we'd have to add hosts one by one. This is actually how it's done for `Message` in `BatchExecRequest` ([`/faabric/src/util/func.cpp`](https://github.com/faasm/faabric/blob/2be5d88e57f1cdcdc6ec62758e7f67aab4c460d4/src/util/func.cpp#L64).

src/scheduler/MpiWorld.cpp
csegarragonz: `try_emplace` might be an overkill here indeed. It does the same as `emplace` (and has the same return value), but the the element is not [constructed if the key is already there](https://en.cppreference.com/w/cpp/container/map/try_emplace) (again, not important with `string`s and `int`s). In short, `emplace` and `try_emplace` guard against the same behaviour: trying to insert a key that already exists.

The error message was not clear at all.

src/scheduler/MpiWorld.cpp
Shillaker: Ah I see, sorry I was getting confused by our `inputdata` and `outputdata` fields. They are also vector-like and you can pass another vector into `set_inputdata` and `set_outputdata`, but it's because they are `bytes` fields, and not standard `repeated` fields.

src/scheduler/MpiWorld.cpp
Shillaker: Yes I understand what `try_emplace` does, but what situation is it guarding against _here_? It seems like this will only error if we're setting a value for a rank that has already been set. AFAICT this won't happen as it's only called once in MPI initialisation right? Even if it did somehow happen (e.g. if someone changes the code in future), it's probably not worthy of an error, at most a warning to say that it's getting reset. Either way I think just using `emplace` is fine and much clearer.

Shillaker (approved):
Nice one 👍

PR #104, opened 2021-06-07
Marginal gains in send/recv fast-path

csegarragonz:
In this PR I revist the `send/recv` fast-path and try to prune unnecessary complexity. The most notable changes are:
1. Switch from `if`-based undefined behaviour to assertions. This is risky and controversial, as the checks will be removed in `Release` builds. I say we use with care, but feels acceptable in this situation.
2. Switch from lock-protected maps to lock-free vectors by initialising resources in advance, rather than lazily. Resource initialisation in both cases is thread safe as it is already protected by the locking in the world registry.

In my rudimentary stress test, I can see a noticeable increase in throughput (5-10%).

Lastly, removing the `if`+exception scheme, some tests which relied on `REQUIRE_THROWS` are now failing. It remains to be discussed what to do with them.

### DIFF
--- a/include/faabric/scheduler/MpiWorld.h
+++ b/include/faabric/scheduler/MpiWorld.h
@@ -195,6 +195,10 @@ class MpiWorld
 
     std::shared_ptr<InMemoryMpiQueue> getLocalQueue(int sendRank, int recvRank);
 
+    void setLocalQueues();
+
+    int getKeyForRanks(int sendRank, int recvRank);
+
     long getLocalQueueSize(int sendRank, int recvRank);
 
     void overrideHost(const std::string& newHost);
@@ -220,12 +224,11 @@ class MpiWorld
     std::string function;
 
     std::shared_ptr<state::StateKeyValue> stateKV;
-    std::unordered_map<int, std::string> rankHostMap;
+    std::vector<std::string> rankHostMap;
 
     std::unordered_map<std::string, uint8_t*> windowPointerMap;
 
-    std::unordered_map<std::string, std::shared_ptr<InMemoryMpiQueue>>
-      localQueueMap;
+    std::vector<std::shared_ptr<InMemoryMpiQueue>> localQueueMap;
 
     std::shared_ptr<faabric::scheduler::MpiAsyncThreadPool> threadPool;
     int getMpiThreadPoolSize();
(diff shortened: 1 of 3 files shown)


src/scheduler/MpiWorld.cpp
```diff
@@ -1227,21 +1213,6 @@ long MpiWorld::getLocalQueueSize(int sendRank, int recvRank)
     return queue->size();
 }
 
-void MpiWorld::checkRankOnThisHost(int rank)
```
csegarragonz, line 1230: Even if we decide to keep the `if` statements, we can delete this method and keep the check in `getLocalQueue` through `getHostForRank`.

include/faabric/scheduler/MpiWorld.h
```diff
@@ -220,12 +224,11 @@ class MpiWorld
     std::string function;
 
     std::shared_ptr<state::StateKeyValue> stateKV;
-    std::unordered_map<int, std::string> rankHostMap;
+    std::vector<std::string> rankHostMap;
```
csegarragonz, line 227: I could use `std::array` instad of `std::vector`, but the changes would then be more involved, as I would have to change the way we initialise the `MpiWorld` (arrays need to know their size at compile time).

tests/test/scheduler/test_mpi_world.cpp
```diff
@@ -301,6 +301,7 @@ TEST_CASE("Test send and recv on same host", "[mpi]")
         REQUIRE(status.bytesSize == messageData.size() * sizeof(int));
     }
 
+    /*
```
csegarragonz, line 304: By switching to `assert`-based errors, some tests that rely on `REQUIRE_THROWS` are not passing. I leave them commented here to remember to discuss about this. 

I've done some reading and it seems as though we could do [some dirty hacks](https://stackoverflow.com/a/38551114/10662021) to have assertions fails throw exceptions during testing, but not convinced.

include/faabric/scheduler/MpiWorld.h
```diff
@@ -195,6 +195,10 @@ class MpiWorld
 
     std::shared_ptr<InMemoryMpiQueue> getLocalQueue(int sendRank, int recvRank);
 
+    void setLocalQueues();
+
+    int getKeyForRanks(int sendRank, int recvRank);
```
Shillaker, line 200: Using the word "key" is a little confusing as it's not a map any more. Perhaps `getArrayIndexForRanks` or something.

This should also be `private` as it's an internal method. At the moment everything on this class is `public` which should be fixed as well (but in another PR).

Shillaker (changes requested):

Shillaker (approved):

PR #105, opened 2021-06-08
Queue-less remote send/recv 

csegarragonz:
In this PR I introduce rank-to-rank messaging bypassing the central message endpoint server.

Sending and receiving messages is in the hot path for most workloads we run, thus we want to introduce the least contention. A low-hanging-fruit is allowing ranks to open messaging endpoints themselves.

As discussed offline, this exposes a bug not covered in the tests, for which I provide a solution in #111.


src/transport/MpiMessageEndpoint.cpp
```diff
@@ -27,4 +27,50 @@ void sendMpiHostRankMsg(const std::string& hostIn,
...
+MpiMessageEndpoint::MpiMessageEndpoint(const std::string& hostIn, int portIn)
+  : sendMessageEndpoint(hostIn, portIn)
+  , recvMessageEndpoint(portIn)
+{}
+
+void MpiMessageEndpoint::sendMpiMessage(
+  const std::shared_ptr<faabric::MPIMessage>& msg)
+{
+    // TODO - is this lazy init very expensive?
+    if (sendMessageEndpoint.socket == nullptr) {
```
csegarragonz, line 40: In normal circumstances, I would open both `sendMessageEndpoint` and `recvMessageEndpoint` in the constructor. However, this is very problematic for testing. That's because:
1. `mpiMessageEndpoints` are `static thread local`, thus we need to create different worlds in separate threads. 
2. If both `send` and `recv` sockets are opened in the constructor, both `recv` sockets will try to bind to the same adress + port, yielding an error (even though we are only using one, in the simplest scenario).

src/scheduler/MpiWorld.cpp
```diff
@@ -25,19 +19,84 @@ MpiWorld::MpiWorld()
...
 {
-    auto it = functionCallClients.find(otherHost);
-    if (it == functionCallClients.end()) {
-        // The second argument is forwarded to the client's constructor
-        auto _it = functionCallClients.try_emplace(otherHost, otherHost);
-        if (!_it.second) {
-            throw std::runtime_error("Error inserting remote endpoint");
+    // Assert the ranks are sane
+    assert(0 <= sendRank && sendRank < size);
+    assert(0 <= recvRank && recvRank < size);
```
Shillaker, line 29: What cases is this guarding against? Could they happen at runtime? AFAICT anything related to invalid MPI ranks can be caused by the application itself, hence must be an exception that will trigger at runtime and provide a useful error message.

src/scheduler/MpiWorld.cpp
```diff
@@ -25,19 +19,84 @@ MpiWorld::MpiWorld()
...
+
+    // Initialise the endpoint vector if not initialised
+    if (mpiMessageEndpoints.size() == 0) {
+        for (int i = 0; i < size * size; i++) {
+            mpiMessageEndpoints.emplace_back(nullptr);
         }
```
Shillaker, line 35: I can see you've ported these changes into the other PR which I reviewed accidentally (https://github.com/faasm/faabric/pull/111), so the comments apply to both. 

Relevant comments from the other PR:

a) Rather than spread it around, I think lazy initialisation for the rank should all happen in one place.
b) I think you can do `mpiMessageEndpoints = std::vector(size * size, nullptr)` and save a few lines.

src/scheduler/MpiWorld.cpp
```diff
@@ -25,19 +19,84 @@ MpiWorld::MpiWorld()
...
+    // Get the index for the rank-host pair
+    int index = getIndexForRanks(sendRank, recvRank);
+    assert(index >= 0 && index < size * size);
+
+    // Lazily initialise send endpoints
+    if (mpiMessageEndpoints[index] == nullptr) {
+        // Get host for recv rank
+        std::string host = getHostForRank(recvRank);
+        assert(!host.empty());
+        assert(host != thisHost);
```
Shillaker, line 47: Same comment re. assertions and invalid ranks.

src/transport/MpiMessageEndpoint.cpp
Shillaker: If the implementation relies on TLS then unfortunately the tests should exercise that TLS. As you say, this means we'd need to create worlds in different threads. 

Re. the ports and testing, the solution is probably to make the base port configurable, then change the value in tests. In this case, each world could take a port offset as a constructor argument, set to the value we currently use by default, but overridden in tests. 

Making this configurable arguable makes these classes more flexible and reusable so I'd say it's a +1 for testing!

src/transport/MpiMessageEndpoint.cpp
```diff
@@ -27,4 +27,49 @@ void sendMpiHostRankMsg(const std::string& hostIn,
...
+    }
+
+    size_t msgSize = msg->ByteSizeLong();
+    {
+        uint8_t sMsg[msgSize];
+        if (!msg->SerializeToArray(sMsg, msgSize)) {
+            throw std::runtime_error("Error serialising message");
+        }
+        sendMessageEndpoint.send(sMsg, msgSize, false);
+    }
```
Shillaker, line 51: Why is this nested in curly braces?

include/faabric/transport/MpiMessageEndpoint.h
```diff
@@ -6,8 +6,31 @@
...
 faabric::MpiHostsToRanksMessage recvMpiHostRankMsg();
 
 void sendMpiHostRankMsg(const std::string& hostIn,
                         const faabric::MpiHostsToRanksMessage msg);
+
+/* This class abstracts the notion of a communication channel between two MPI
+ * ranks. There will always be one rank local to this host, and one remote.
+ * Note that the port is unique per (user, function, sendRank, recvRank) tuple.
+ */
+class MpiMessageEndpoint
```
Shillaker, line 21: It would be nice to add some tests for this class. It contains non-trivial logic and is used quite deep in the stack so it's good to have a simple test that makes debugging issues easier.

Shillaker (changes requested):

src/scheduler/MpiWorld.cpp
csegarragonz: I think the first `assert` can go as we actually check this in `getHostForRank` but the latter has to do with logic particular to this method. I.e. if we send to a remote rank, the sending rank, must actually be in a remote host.

src/scheduler/MpiWorld.cpp
csegarragonz: b) `mpiMessageEndpoints` is a vector of `std::unique_ptr` so it has a deleted `=` operator. Thus I need to emplace `nullptr`s back. Note that this also answers your concern in #111 wrt `mpiMessageEndpoint[index` being actually initialised to `nullptr`.

tests/test/scheduler/test_remote_mpi_worlds.cpp
```diff
@@ -83,45 +96,69 @@ TEST_CASE_METHOD(RemoteMpiTestFixture, "Test send across hosts", "[mpi]")
...
     // Destroy worlds
+    senderThread.join();
     localWorld.destroy();
-    remoteWorld.destroy();
-
-    server.stop();
 }
 
 TEST_CASE_METHOD(RemoteMpiTestFixture,
-                 "Test collective messaging across hosts",
```
csegarragonz, line 93: These tests weren't using two different threads (not sure why they were passing). Given that I had to change them, I also separated them into smaller tests, easier to identify which fails.

src/transport/MpiMessageEndpoint.cpp
csegarragonz: As discussed offline, we'll first try with binding to a specific (protocol, address, port) tuple, and will leave port management until the multi-tenancy PR.

tests/utils/fixtures.h
```diff
@@ -117,6 +118,19 @@ class ConfTestFixture
     faabric::util::SystemConfig& conf;
 };
 
+class MessageContextFixture : public SchedulerTestFixture
```
csegarragonz, line 121: Had to move this here to use it in the `MpiMessageEndpoint` tests.

src/scheduler/MpiWorld.cpp
Shillaker: Ah that's annoying, makes sense though.

Shillaker (approved):

PR #107, opened 2021-06-08
Switch to spdlog macros

Shillaker:
The aim here is to switch to using spdlog macros (`SPDLOG_DEBUG`, `SPDLOG_INFO` etc.) for two reasons:

- Replace our current logging implementation which is a point of contention in a multi-threaded environment.
- Allow us to noop trace and debug statements in release builds.

How:

- Global find and replace on all the logging statements.
- Set the `SPDLOG_ACTIVE_LEVEL` macro depending on whether it's a release build. This sets the minimum possible logging level (i.e. if it's set to info (`2` in this case), this means the debug and trace macros are nooped).
- Reintroduce the `initLogging` function which calls `spdlog::set_level` to set the active logging level for that execution. This selects which log statements get printed out at runtime and uses the `faabric::util::SystemConfig::logLevel` property as before.

This means, in a debug build you can still change `LOG_LEVEL` to `debug` or `trace` without recompiling, because we have `SPDLOG_ACTIVE_LEVEL` set to `0`, thus allowing all levels of logging. However, in a release build, even if you set `LOG_LEVEL` to `debug`, the debug statements won't get printed as the macros are nooped.

I've added the relevant comments in the code, including some warnings if you've chosen a level that won't get printed.

[spdlog docs](https://github.com/gabime/spdlog#usage-samples) for reference.

The only files with non-trivial changes are:

- `include/util/logging.h` - sets `SPDLOG_ACTIVE_LEVEL`, i.e. the minimum possible level.
- `src/util/logging.cpp` - contains the implementation of `initLogging` which sets the current logging level and log format.


csegarragonz (approved):
:+1:

PR #108, opened 2021-06-09
Test fixtures for MPI

csegarragonz:
In this PR I implement three test fixtures for MPI-related tests:
1. `MpiBaseTestFixture`: which defines the usual variables but _does not_ initialise a world.
2. `MpiTestFixture`: inherits from 1, and initialises a world.
3. `RemoteMpiTestFixture`: useful for testing remote messaging locally, it sets the resources in each host as desired, but for this reason can't initialise the worlds transparently.

I have also added a separate file to hold the remote messaging tests. I am aware this is a great example of something that can be tested using `dist-tests` but I think there's still value in mocking for single-host tests (?).

Next step will be to add more remote messaging tests, as with #108 they will use a different mechanism than local messages.

### DIFF
--- a/tests/test/scheduler/test_mpi_context.cpp
+++ b/tests/test/scheduler/test_mpi_context.cpp
@@ -1,28 +1,16 @@
-#include "faabric_utils.h"
 #include <catch.hpp>
 
 #include <faabric/scheduler/MpiContext.h>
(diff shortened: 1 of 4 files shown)


Shillaker (approved):
Nice, much cleaner.

PR #110, opened 2021-06-10
Set default MPI world size when initialising from message

Shillaker:
The default world size wasn't getting set via the `initialiseFromMsg` function, which is an alternative route for creating a world other than the `create` method. 

This only causes a problem the first time an application results in `getLocalQueue` getting called, at which point there are no local queues initialised (because the world size is zero). This was only covered by an assertion, so I switched it to an exception for a clearer error message.

I realise this could also happen if the world size requested in the message doesn't match that used by the application (e.g. if I set world size to 5 then try to send something to rank 10). This is difficult to test for ahead of time, so we just need to make sure there are execptions with meaningful messages thrown when it causes a problem.

### DIFF
--- a/src/scheduler/MpiWorld.cpp
+++ b/src/scheduler/MpiWorld.cpp
@@ -176,9 +176,9 @@ void MpiWorld::initialiseFromMsg(const faabric::Message& msg, bool forceLocal)
     int msgWorldSize = msg.mpiworldsize();
     if (msgWorldSize <= 0) {
         size = conf.defaultMpiWorldSize;
-        logger->warn("Defaulting to default world size ({}) for {}",
-                     msgWorldSize,
-                     faabric::util::funcToString(msg, false));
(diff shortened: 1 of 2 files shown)


csegarragonz (changes requested):
> The default world size wasn't getting set via the initialiseFromMsg function, which is an alternative route for creating a world other than the create method.

As mentioned offline, I disagree that `initialiseFromMsg` is an alternative route to create a world (from scratch). Note that things like the host-to-rank vector are not set unless the world is created through `create` (actually through `createWorld` in `MpiWorldRegistry`). Further, creating through the registry is vital to avoid race conditions and ensuring all worlds are ready to go, as in the registry is where we hold the locks for world creation.

Is there any use case that creates a world through `initialiseFromMsg`? (Without the world having been created elsewhere before that is).

Happy to discuss this offline if you prefere, just :calling:

PR #111, opened 2021-06-10
Different async mechanism: MPI message buffer

csegarragonz:
## Problem

Current thread pool is shared among all ranks. Thus, any thread in the pool may perform send/recv in lieu of any other rank in the host. This has several problems:
  1. Increases the number of thread-local endpoint arrays.
  2. Throws `EADDRINUSE` errors when the main thread and a pool-worker thread try to recv from the same rank (into the same rank as well).
  3. In general, it breaks the point-to-point thread-local semantics desirable  for MPI.
  4. Lastly, `send/recv` are very lightweight methods. One could argue that  running them in separate threads is an overkill.

In [line 327 here](https://github.com/faasm/faabric/pull/111/files#diff-dcdea8519fa1d9e1ded61a65520a51debeb6ca5e1d40a78a8dd52b00a99d3a0a) I introduce a test that breaks the current semantics.

## Solution

Remove the thread pool. Keep track of async messages that we expected to receive per `(sendRank, recvRank)` and perform batch reads in blocking calls.

~~Rely on ZeroMQ's asyncrhonous API. In particular, on the `poll` call, that blocks until it has received at least one message (but does not actually call recv). I.e. it polls the underlying receive buffers.~~
After having thought through it, I think we don't need polling (which is used to multiplex input sockets) but just `recv`-ing enough times according to our protocol.

**We assume** that the MPI application expects to `recv` messages in the order it issues `recv/irecv`s. This is, if the application issues `irecv`, `irecv`, and `recv`, then the 3rd message entering our network buffer will correspond to the `recv`, and not to any of the `irecv` even in this case:

```cpp
/* SENDER */
send(msg1);
send(msg2);
send(msg3);

/* RECEIVER */
int id1 = irecv();
int id2 = irecv();

// third message in buffer
msg_t msg3 = recv();

// second message in buffer
msg_t msg2 = await(id2)
// first message in buffer
msg_t msg1 = await(id1)
```

This is the case for OpenMPI tag-less (tested), note that we don't support tags.

## Detail on the implementation

We introduce a thread local unacknowledged message buffer (UMB). There is one per rank-to-rank channel local to the host. Each host has `localRanks * worldSize` such channels.

It stores messages that have been received, but not claimed, or claimed but not received.

The struct prototype is:
```cpp
struct UMB {
    // List of claims
    std::list<int> ids;
    // List of messages
    std::list<msg_t> msgs;

    // How many claims have no matching message
    int pendingIds() {
        return std::max<int>(0, ids.size() - msgs.size());
    }
};
```

There are three methods that modify UMB: `recv -> msg`, `irecv -> id`, `await(id) -> msg`. All are executed by the same thread so we need no locks. I include the pseudo-code for each:

1. `irecv`: receive a message asynchronously.
```cpp
int irecv()
{
    int id = generateId();
    UMB.ids.push_back(id);
}
```

2. `recv`: recv a message synchronously (block until we receive it)
```cpp
msg_t recv()
{
    return recvBatchReturnLast(pendingIds() + 1);
}
```

3. `await(id)`: wait until `irecv` with `id` has finished (i.e. msg has arrived)
```cpp
msg_t await(id)
{
    // Await comes after irecv
    auto index = UMB.ids.find(id);
    assert(index != UMB.ids.end());

    // We assume that awaits can happen in a different order than `irecv`s,
    // that's why we can't use queues in the UMB in the first place, as here
    // we may want to pop an element in the middle of the list
    assert(UMB.msgs.size > index);
    auto it = std::advance(UMB.msgs.begin(), index);
    if (*it != nullptr) {
        msg_t toReturn = *it;
        // Remove ids and message from lists
        UMB.msgs.erase(it);
        UMB.ids.erase(index);
        return toReturn;
    } else {
        // Receive enough messages to get to our index
        return recvBatchReturnLast(index);
    }
}
```

Lastly, here's the auxiliary method to poll until a certain number of messages have arrived our inbound network sockets (ZeroMQ's):

```cpp
msg_t recvBatchReturnLast(int numMsgToRecv)
{
    // First we receive all messages for which there is an id but no msg
    // i.e. `irecv`'s that happened before our `recv/await`.
    for (int i = 0; i < numMsgToRecv - 1; i++) {
        if (UMB.isLocal) {
            msg_t msg = localQueue.dequeue();
        } else {
            msg_t msg = socket.recv();
        }
        UMB.msg.push_back(msg);
    }

    // This is the message we are interested in returning
    msg_t toReturn = socket.recv;
    // Note that there may be other messages ready to be acknowledged in the
    // underlying buffer, but adding them to the UMB may add messages that 
    // correspond to standard `recv` and not `irecv`.

    return msg;
}
```

## Possible pitfalls

1. Is this slowing our fast path down?
2. Can a slow `recv` be stuck behind a `irecv` that never appears?
    - No. According to our assumptions, all `irecv` sends need to be done before the `recv` send, if `recv` happens after those `irecv`.


Shillaker:
This looks great, very well thought out. I also like the iteration to avoid 0MQ's async API, the simpler we make it the better. 

Re. slowing down the fast path, it's probably much of a muchness, given that this implementation is much nicer (and actually works), I wouldn't let that stop us. We can do some benchmarking once it's done as well.

src/scheduler/MpiWorld.cpp
```diff
@@ -1,20 +1,21 @@
...
 #include <faabric/util/macros.h>
-#include <faabric/util/timing.h>
 
-static thread_local std::unordered_map<int, std::future<void>> futureMap;
-static thread_local std::unordered_map<std::string,
-                                       faabric::scheduler::FunctionCallClient>
-  functionCallClients;
+/* Each MPI rank runs in a separate thread, however they interact with faabric
+ * as a library. Thus, we use thread_local storage to guarantee that each rank
+ * sees its own version of these data structures.
```
Shillaker, line 10: I don't quite understand this comment, what do you mean "they interact with faabric as a library"?

src/scheduler/MpiWorld.cpp
```diff
@@ -113,13 +185,37 @@ void MpiWorld::create(const faabric::Message& call, int newId, int newSize)
 
 void MpiWorld::destroy()
 {
-    // Destroy once per host
-    if (!isDestroyed.test_and_set()) {
-        shutdownThreadPool();
+    // Destroy once per thread the rank-specific data structures
+    // Remote message endpoints
+    if (mpiMessageEndpoints.size() > 0) {
```
Shillaker, line 190: I think this and other `foo.size() > 0` is a clang-tidy warning about using `empty` when checking for emptiness as it's a bit more readable (i.e. `if(!mpiMessageEndpoints.size().empty())`)

src/scheduler/MpiWorld.cpp
```diff
@@ -113,13 +185,37 @@ void MpiWorld::create(const faabric::Message& call, int newId, int newSize)
...
+        }
+        mpiMessageEndpoints.clear();
+    }
+
+    // Unacked message buffers
+    if (unackedMessageBuffers.size() > 0) {
+        for (auto& umb : unackedMessageBuffers) {
+            if (umb != nullptr) {
+                assert(umb->ids.empty() && umb->msgs.empty() &&
+                       umb->args.empty());
```
Shillaker, line 204: There are a lot of assertions in this PR and it's not clear what role they're all performing. Let's discuss offline.

src/scheduler/MpiWorld.cpp
```diff
@@ -113,13 +185,37 @@ void MpiWorld::create(const faabric::Message& call, int newId, int newSize)
...
+            }
+        }
+        unackedMessageBuffers.clear();
+    }
 
-        // Note - we are deliberately not deleting the KV in the global state
-        // TODO - find a way to do this only from the master client
+    // Request to rank map
+    assert(reqIdToRanks.empty());
+    reqIdToRanks.clear();
```
Shillaker, line 215: Why would you need to clear after you've asserted that it's empty?

src/scheduler/MpiWorld.cpp
```diff
@@ -345,23 +413,14 @@ int MpiWorld::isend(int sendRank,
                     faabric::MPIMessage::MPIMessageType messageType)
 {
     int requestId = (int)faabric::util::generateGid();
+    auto it = reqIdToRanks.try_emplace(
+      requestId, MPI_IS_ISEND_REQUEST, MPI_IS_ISEND_REQUEST);
+    if (!it.second) {
+        SPDLOG_ERROR("Request ID {} is already present in map", requestId);
+        throw std::runtime_error("Request ID already in map");
+    }
```
Shillaker, line 421: I don't understand what this check is guarding against, surely this would only trigger if `generateGid` wasn't working properly? We've talked about `try_emplace` before and I think this form of check is quite verbose, so we should use it sparingly. When we do, we should add a comment as to why.

src/scheduler/MpiWorld.cpp
```diff
@@ -714,17 +787,65 @@ void MpiWorld::awaitAsyncRequest(int requestId)
...
+        SPDLOG_ERROR("Asynchronous request id not recognized: {}", requestId);
+        throw std::runtime_error("Unrecognized async request id");
+    }
+    int sendRank = it->second.first;
+    int recvRank = it->second.second;
+    reqIdToRanks.erase(it);
+
+    // If awaiting an isend request, return immediately as our transport layer
+    // is asynchronous, thus request is already handled
+    if (sendRank == MPI_IS_ISEND_REQUEST && recvRank == MPI_IS_ISEND_REQUEST) {
```
Shillaker, line 802: Putting this `MPI_IS_ISEND_REQUEST` constant in as a marker for isend requests is a bit hard to read; it's overloading the `reqIdToRanks` map with two jobs. Instead it might be easier to understand and reason about it we had a separate `std::set` to track the IDs of isend requests, then this method would start with:

```
if(iSendRequests.find(requestId) == iSendRequests.end()) {
    return;
}
```

I also think debugging might get confusing as the value of the constant is `-1`, so you might have to take a few steps to realise what's going on to get a rank of `-1`.

src/scheduler/MpiWorld.cpp
```diff
@@ -714,17 +787,65 @@ void MpiWorld::awaitAsyncRequest(int requestId)
...
 
-    // This call blocks until requestId has finished.
-    it->second.wait();
-    futureMap.erase(it);
+    std::shared_ptr<faabric::scheduler::UnackedMessageBuffer> umb =
+      getUnackedMessageBuffer(sendRank, recvRank);
+
+    // The request id must be in the UMB, as an irecv must happen before an
+    // await. It is very likely that if the assert were to fail, the previous
+    // check would have also failed, thus why we assert here.
```
Shillaker, line 811: The assertions in this method and several others in this PR seem to be guarding against something that might fail at runtime, e.g. a message not getting delivered due to some networking issue or an issue with the running application logic sending and receiving from the wrong ranks. 

This comment refers to "a previous check" but it's hard to tell what that is, and what error condition it's actually describing. Perhaps something like "If the message we are awaiting isn't present in the unacked message buffer, then it either hasn't been sent, or hasn't been delivered". If it hasn't been sent then this could be an issue with the MPI application we're executing, or it could be a network failure, or it could be the logic of our implementation. The first two cases warrant an exception with a meaninful error message.

src/scheduler/MpiWorld.cpp
```diff
@@ -1104,6 +1209,70 @@ void MpiWorld::initLocalQueues()
...
+    // (batchSize - 1) were `irecv`-ed before ours.
+    std::shared_ptr<faabric::MPIMessage> m;
+    auto it = firstNullMsg;
+    if (isLocal) {
+        // First receive messages that happened before us
+        for (int i = 0; i < batchSize - 1; i++) {
+            SPDLOG_TRACE("MPI - pending recv {} -> {}", sendRank, recvRank);
+            auto _m = getLocalQueue(sendRank, recvRank)->dequeue();
+
+            assert(_m != nullptr);
```
Shillaker, line 1241: This assertion should go inside `getLocalQueue`.

src/scheduler/MpiWorld.cpp
```diff
@@ -374,28 +433,32 @@ int MpiWorld::irecv(int sendRank,
...
-                                count,
-                                nullptr,
-                                messageType),
-                      std::move(resultPromise)));
-
-    // Place the promise in a map to wait for it later
-    futureMap.emplace(std::make_pair(requestId, std::move(resultFuture)));
+    getUnackedMessageBuffer(sendRank, recvRank)->ids.push_back(requestId);
+    getUnackedMessageBuffer(sendRank, recvRank)->args.push_back(args);
+    getUnackedMessageBuffer(sendRank, recvRank)->msgs.push_back(nullptr);
```
Shillaker, line 448: Given that we've got a separate object to represent an `UnackedMessageBuffer` it seems like we could wrap these three lines into a method on that object, e.g. `addMessage(requestId, args, nullptr)`.

src/scheduler/MpiWorld.cpp
```diff
@@ -714,17 +787,65 @@ void MpiWorld::awaitAsyncRequest(int requestId)
...
+
+    // Get the corresponding message
+    assert(idDistance < umb->msgs.size());
+    auto msgIndex = umb->msgs.begin();
+    std::advance(msgIndex, idDistance);
+
+    // Get the corresponding request arguments
+    assert(idDistance < umb->args.size());
+    auto argsIndex = umb->args.begin();
+    std::advance(argsIndex, idDistance);
```
Shillaker, line 824: This logic using `advance` and `distance` is quite hard to read but seems like it's all operations on the message buffer, therefore could this not be a method on that object? E.g. `umb.ackMessage(requestId)`?

src/scheduler/MpiWorld.cpp
```diff
@@ -714,17 +787,65 @@ void MpiWorld::awaitAsyncRequest(int requestId)
...
+           argsIndex->buffer,
+           argsIndex->dataType,
+           argsIndex->count,
+           MPI_STATUS_IGNORE,
+           argsIndex->messageType);
+
+    // Remove the acknowledged indexes from the UMB
+    umb->ids.erase(idIndex);
+    umb->msgs.erase(msgIndex);
+    umb->args.erase(argsIndex);
```
Shillaker, line 848: Feels like these three lines could also be captured in a method on the buffer object.

include/faabric/scheduler/MpiWorld.h
```diff
@@ -3,19 +3,41 @@
...
 #include <atomic>
-#include <thread>
+#include <list>
 
 namespace faabric::scheduler {
 typedef faabric::util::Queue<std::shared_ptr<faabric::MPIMessage>>
   InMemoryMpiQueue;
 
-std::string getWorldStateKey(int worldId);
+/* The untracked message buffer (UMB) keeps track of the asyncrhonous
```
Shillaker, line 18: Typo, "untracked" vs "unacked". The object doesn't seem to have any properties that are unique to the messages being unacked, could we just call it a `MPIMessageBuffer` and put it in a separate file with some tests (see my comments below about encapsulating the logic from elsewhere in this file into this class).

include/faabric/scheduler/MpiWorld.h
```diff
@@ -3,19 +3,41 @@
...
+{
+    struct Arguments
+    {
+        int sendRank;
+        int recvRank;
+        uint8_t* buffer;
+        faabric_datatype_t* dataType;
+        int count;
+        faabric::MPIMessage::MPIMessageType messageType;
+    };
```
Shillaker, line 35: If we put the shared pointer to the message into this struct, could we get rid of one of the `std::lists` in this class?

src/scheduler/MpiWorld.cpp
```diff
@@ -25,40 +30,109 @@ MpiWorld::MpiWorld()
...
 {
-    auto it = functionCallClients.find(otherHost);
-    if (it == functionCallClients.end()) {
-        // The second argument is forwarded to the client's constructor
-        auto _it = functionCallClients.try_emplace(otherHost, otherHost);
-        if (!_it.second) {
-            throw std::runtime_error("Error inserting remote endpoint");
+    // Assert the ranks are sane
+    assert(0 <= sendRank && sendRank < size);
+    assert(0 <= recvRank && recvRank < size);
```
Shillaker, line 40: This assertion could fail because of an error in the application we're executing, so should probably be an error with a useful message.

src/scheduler/MpiWorld.cpp
```diff
@@ -25,40 +30,109 @@ MpiWorld::MpiWorld()
...
+
+    // Initialise the endpoint vector if not initialised
```
Shillaker, line 42: It looks like there's a few places where we're doing lazy initialisation in this class that could be captured in a single method, e.g. `initialiseAsyncForRank()`. This would be invoked call at the start of each `isend`/ `irecv` calls and do nothing if the initialisation had already happened, e.g. 

```
static thread_local asyncInitialised = false;

void initialiseAsyncForRank() {
    if(asyncInitialised) {
        return;
    }
    
    mpiMessageEndpoints = std::list(size * size, nullptr);
    
    // Other init
    
    asyncInitialised = true;
}
```

src/scheduler/MpiWorld.cpp
```diff
@@ -25,40 +30,109 @@ MpiWorld::MpiWorld()
...
+    if (mpiMessageEndpoints[index] == nullptr) {
+        // Get host for recv rank
+        std::string host = getHostForRank(recvRank);
+        assert(!host.empty());
+        assert(host != thisHost);
+
+        // Get port for send-recv pair
+        int port = getMpiPort(sendRank, recvRank);
+        // TODO - finer-grained checking when multi-tenant port scheme
+        assert(port > 0);
```
Shillaker, line 63: This assertion should live in `getMpiPort` if we want to keep it, but it seems like it will only fail if the ranks are invalid, which is probably something we should check higher up the stack with an exception and error message.

Shillaker (changes requested):

Shillaker:
Ah sorry I got confused and reviewed this when you had actually requested a review on the other one 🤦

src/scheduler/MpiWorld.cpp
```diff
@@ -25,40 +30,109 @@ MpiWorld::MpiWorld()
...
-        it = _it.first;
     }
-    return it->second;
+
+    // Get the index for the rank-host pair
+    int index = getIndexForRanks(sendRank, recvRank);
+    assert(index >= 0 && index < size * size);
+
+    // Lazily initialise send endpoints
+    if (mpiMessageEndpoints[index] == nullptr) {
```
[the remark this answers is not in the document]
csegarragonz: (Currently going through the comments in #105, but this applies as well)

And yes, this will definately be uninitialised as we explicitely `emplace_back` a nullptr in line 45.

src/scheduler/MpiWorld.cpp
csegarragonz: But `getLocalQueue` does not interact with the message at all right? We need to call `dequeue()` on it. Granted that we only `enqueue()` items ourselves (in local `send()`, not in remote ones anymore), and we assert this before enqueueing, thus I thought an `assert` went well.

src/scheduler/MpiWorld.cpp
csegarragonz: You are right. As we discussed `try_emplace` and `emplace` are mostly interchangable. The main problem here is that what I check would only trigger if `generateGid` wasn't working properly, and this is not the right place to check it.

src/scheduler/MpiWorld.cpp
```diff
@@ -25,40 +30,109 @@ MpiWorld::MpiWorld()
...
+// local nature, we expect it to be quite sparse (i.e. filled with nullptr).
+std::shared_ptr<faabric::scheduler::UnackedMessageBuffer>
+MpiWorld::getUnackedMessageBuffer(int sendRank, int recvRank)
+{
+    // Lazy initialise all empty slots
+    if (unackedMessageBuffers.size() == 0) {
+        for (int i = 0; i < size * size; i++) {
+            unackedMessageBuffers.emplace_back(nullptr);
+        }
+    }
```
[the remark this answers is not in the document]
csegarragonz: In this case yes as we are using shared pointers. However, I will use `resize` as I think it fits better.
This is, `unackedMessageBuffers.resize(size * size, nullptr)`.

include/faabric/scheduler/MpiWorld.h
csegarragonz: Yes, in fact after some deep refactoring I realized we could just as well use one single list (which makes iterator management much easier).

src/scheduler/MpiWorld.cpp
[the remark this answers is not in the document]
csegarragonz: Really the comment is too verbose and poorly phrased, will rephrase.

tests/test/scheduler/test_remote_mpi_worlds.cpp
```diff
@@ -366,4 +362,187 @@ TEST_CASE_METHOD(RemoteCollectiveTestFixture,
     senderThread.join();
     localWorld.destroy();
 }
+
+TEST_CASE_METHOD(RemoteMpiTestFixture,
+                 "Test sending sync and async message to same host",
```
csegarragonz, line 367: This is the test that exposed the flaws in the thread pool (and would make the current master fail).

tests/test/scheduler/test_mpi_message_buffer.cpp
```diff
@@ -0,0 +1,119 @@
...
+    if (overrideRequestId != -1) {
+        requestId = overrideRequestId;
+    } else {
+        requestId = static_cast<int>(faabric::util::generateGid());
+    }
+
+    MpiMessageBuffer::Arguments args = {
+        requestId, nullptr, 0, 1,
+        nullptr,   MPI_INT, 0, faabric::MPIMessage::NORMAL
+    };
```
Shillaker, line 22: This is a good example of why this sort of struct initialisation is not great, it's hard to tell what each of these numbers and `nullptr`s refer to.

include/faabric/scheduler/MpiWorld.h
```diff
@@ -221,6 +214,24 @@ class MpiWorld
                                                               int recvRank);
     void closeMpiMessageEndpoints();
 
+    // Support for asyncrhonous communications
+    std::shared_ptr<faabric::scheduler::MpiMessageBuffer>
```
Shillaker, line 218: Could perhaps save a few characters by removing the `faabric::scheduler::` prefix when in the `faabric::scheduler` namespace already.

src/scheduler/MpiMessageBuffer.cpp
```diff
@@ -0,0 +1,69 @@
+#include <faabric/scheduler/MpiMessageBuffer.h>
+#include <faabric/util/logging.h>
+
+namespace faabric::scheduler {
+typedef std::list<MpiMessageBuffer::Arguments>::iterator ArgListIterator;
```
Shillaker, line 5: As mentioned re. the naming of `Arguments`, I think the name of this `ArgListIterator` is a little ambiguous. Given that it's an `MpiMessageBuffer` it would perhaps be clearer if it was something like `MpiMessageIterator`

src/scheduler/MpiWorld.cpp
```diff
@@ -170,13 +177,43 @@ void MpiWorld::create(const faabric::Message& call, int newId, int newSize)
...
+                    SPDLOG_ERROR("Destroying the MPI world with outstanding {}"
+                                 " messages in the message buffer",
+                                 umb->size());
+                    throw std::runtime_error(
+                      "Destroying world with a non-empty MPI message buffer");
+                }
+            }
+        }
+        unackedMessageBuffers.clear();
+    }
```
Shillaker, line 205: Would be good to add a couple of small tests to make sure these `destroy` checks work, e.g. call `isend` a few times then call `destroy` and `REQUIRE_THROWS`

Shillaker (changes requested):
Nice, this is looking good, just a few style/ structure changes.

src/scheduler/MpiWorld.cpp
csegarragonz: Thanks for pointing out, as I had actually missed checking if `iSendRequest` was empty at destruction time.

include/faabric/scheduler/MpiWorld.h
```diff
@@ -212,15 +205,36 @@ class MpiWorld
     void initLocalQueues();
 
     // Rank-to-rank sockets for remote messaging
-    void initRemoteMpiEndpoint(int sendRank, int recvRank);
-    int getMpiPort(int sendRank, int recvRank);
+    std::vector<int> basePorts;
```
csegarragonz, line 208: I have changed these to accomodate for the new port offset per world.

src/scheduler/MpiWorld.cpp
```diff
@@ -261,16 +262,51 @@ std::string MpiWorld::getHostForRank(int rank)
     return host;
 }
 
+// Returns a pair (sendPort, recvPort)
```
csegarragonz, line 265: Also worth checking this which has changed since last review.

csegarragonz: (basically [this commit](https://github.com/faasm/faabric/pull/111/commits/88bcb7d6a6c88f8118301d8948fd9b39e40f3acc))

Shillaker (approved):
Nice, LGTM.

PR #112, opened 2021-06-14
Change flush to execute once per host

Shillaker:
Flushing is a one-off host-wide task that needs to be called _after_ all the executors have terminated. However, flushing was previously implemented as an instance method on the `Executor`, meaning it was unnecessarily called repeatedly and wouldn't actually get called if there were no executors present. Therefore I've moved it to a method on the `ExecutorFactory` (which is a singleton), so it now gets called once per host.

### DIFF
--- a/src/scheduler/FunctionCallServer.cpp
+++ b/src/scheduler/FunctionCallServer.cpp
@@ -66,9 +66,6 @@ void FunctionCallServer::recvFlush(faabric::transport::Message& body)
 
     // Clear the scheduler
     scheduler.flushLocally();
-
-    // Reset the scheduler
-    scheduler.reset();
 }
 
 void FunctionCallServer::recvExecuteFunctions(faabric::transport::Message& body)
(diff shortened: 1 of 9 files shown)


src/scheduler/FunctionCallServer.cpp
```diff
@@ -66,9 +66,6 @@ void FunctionCallServer::recvFlush(faabric::transport::Message& body)
 
     // Clear the scheduler
     scheduler.flushLocally();
-
-    // Reset the scheduler
-    scheduler.reset();
```
Shillaker, line 71: Scheduler `reset` gets called in `flushLocally`

csegarragonz (approved):
LGTM. However, the dummy tests look indeed a bit dumb. Maybe not in this PR, but is there a way we could actually test the issue reported in faasm/faasm#381 ? Maybe in the distributed tests?

I've also ran into the same issue myself after the issue was closed.

Shillaker:
> 
> 
> LGTM. However, the dummy tests look indeed a bit dumb. Maybe not in this PR, but is there a way we could actually test the issue reported in [faasm/faasm#381](https://github.com/faasm/faasm/issues/381) ? Maybe in the distributed tests?
> 
> I've also ran into the same issue myself after the issue was closed.

All the complexity around flushing is in Faasm so that's where the tests should be. It's a good point re. that issue and it's something that's been hanging around for a while. A more rigorous distributed test in Faasm would be good.

PR #114, opened 2021-06-16
Fix MPI world creation of size 1

csegarragonz:
Running MPI applications with world size 1 would not work as we'd call `callFunctions` with an empty protobuf message, and the library would raise an error.

Adding a simple check, and a test that (surprisingly) failed before and now passes.

### DIFF
--- a/tests/test/scheduler/test_mpi_world.cpp
+++ b/tests/test/scheduler/test_mpi_world.cpp
@@ -43,6 +43,19 @@ TEST_CASE_METHOD(MpiBaseTestFixture, "Test world creation", "[mpi]")
     world.destroy();
 }
 
+TEST_CASE_METHOD(MpiBaseTestFixture, "Test creating world of size 1", "[mpi]")
+{
+    // Create the world
+    MpiWorld world;
+    int worldSize = 1;
+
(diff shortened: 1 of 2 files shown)


tests/test/scheduler/test_mpi_world.cpp
```diff
@@ -43,6 +43,19 @@ TEST_CASE_METHOD(MpiBaseTestFixture, "Test world creation", "[mpi]")
     world.destroy();
 }
 
+TEST_CASE_METHOD(MpiBaseTestFixture, "Test creating world of size 1", "[mpi]")
```
Shillaker, line 46: I think this test could be a bit more specific in checking the effect we're trying to avoid (i.e. initialising the world but not sending messages). It would look a lot like the test above where we check a valid world initialisation.

```
// Create a world of size 1
MpiWorld world;
world.create(msg, worldId, 1);

// Check details are set up properly
REQUIRE(world.getSize() == 1);
REQUIRE(world.getId() == worldId);
REQUIRE(world.getUser() == user);
REQUIRE(world.getFunction() == func);

// Check no messages are sent
REQUIRE(sch.getRecordedMessagesAll().empty());
```

Shillaker (changes requested):

Shillaker (approved):

PR #115, opened 2021-06-17
Remove local queue clearing

csegarragonz:
As discussed offline, clearing the local queues is not necessary when destroying the world, and together with the barrier in faasm could be causing some gnarly race conditions.


Shillaker (approved):
LGTM, happy to add the barrier removal here too, but equally happy with doing them separately.

PR #119, opened 2021-06-17
Add send/ recv timeouts to message endpoints

Shillaker:
At the moment we don't set any timeouts in our use of 0MQ, which means we get an infinite hang if something goes wrong. Instead we would like things to time out, so this PR sets the send timeout and recv timeout on every 0MQ socket by default.

I actually can't work out in which scenario the send timeout would be triggered, hence there's no test for it and it could be overkill. However, for now I'd like to keep it in case it catches any of the issues we've been seeing.

### DIFF
--- a/include/faabric/transport/MessageEndpoint.h
+++ b/include/faabric/transport/MessageEndpoint.h
@@ -13,6 +13,11 @@
 
 #define ANY_HOST "0.0.0.0"
 
+// These timeouts should be long enough to permit sending and receiving large
+// messages, but short enough not to hang around when something has gone wrong.
+#define DEFAULT_RECV_TIMEOUT_MS 20000
+#define DEFAULT_SEND_TIMEOUT_MS 20000
+
 namespace faabric::transport {
 enum class SocketType
 {
@@ -59,11 +64,21 @@ class MessageEndpoint
 
     int getPort();
 
+    void setRecvTimeoutMs(int value);
+
+    void setSendTimeoutMs(int value);
+
   protected:
     const std::string host;
     const int port;
     std::thread::id tid;
     int id;
+
+  private:
+    int recvTimeoutMs = DEFAULT_RECV_TIMEOUT_MS;
+    int sendTimeoutMs = DEFAULT_SEND_TIMEOUT_MS;
+
+    void validateTimeout(int value);
 };
 
 /* Send and Recv Message Endpoints */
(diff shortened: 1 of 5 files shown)


include/faabric/transport/MessageEndpoint.h
```diff
@@ -59,11 +65,20 @@ class MessageEndpoint
...
   protected:
     const std::string host;
     const int port;
     std::thread::id tid;
     int id;
+
+    int recvTimeoutMs = DEFAULT_RECV_TIMEOUT_MS;
+    int sendTimeoutMs = DEFAULT_SEND_TIMEOUT_MS;
+
+    void validateTimeout(int value);
```
Shillaker, line 81: These have to be `protected` because of the subclassing that goes on.

src/transport/MessageEndpoint.cpp
```diff
@@ -144,10 +155,11 @@ Message MessageEndpoint::recv(int size)
...
                 SPDLOG_TRACE("Shutting endpoint down after receiving ETERM");
                 return Message();
-            } else {
-                SPDLOG_ERROR("Error receiving message: {}", e.what());
-                throw;
             }
+
+            // Print default message and rethrow
+            SPDLOG_ERROR("Error receiving message: {} ({})", e.num(), e.what());
+            throw;
```
Shillaker, line 162: This message was missing the error code before.

tests/test/transport/test_message_server.cpp
```diff
@@ -163,4 +186,56 @@ TEST_CASE("Test multiple clients talking to one server", "[transport]")
 
     server.stop();
 }
+
+TEST_CASE("Test client timeout on requests to valid server", "[transport]")
```
Shillaker, line 190: Not sure if this is the right place to put this test. It requires a custom server, but is really testing functionality in the client.

csegarragonz (approved):
LGTM. Just one minor pint re. infinite timeouts.

src/transport/MessageEndpoint.cpp
```diff
@@ -253,6 +265,31 @@ int MessageEndpoint::getPort()
     return port;
 }
 
+void MessageEndpoint::validateTimeout(int value)
+{
+    if (value <= 0) {
```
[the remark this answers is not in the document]
Shillaker: We should not support infinite timeouts anywhere. I can't think of a case where it's justified.

PR #120, opened 2021-06-18
Fix race conditions around executor reuse

Shillaker:
Once an executor has finished execution it needs to do three things:

- Set the function result - this notifies anything waiting on that task's completion.
- Vacate a slot in the scheduler - the scheduler keeps track of how many threads or processes are running concurrently via these slots.
- Release its claim - the scheduler sets a boolean flag called a claim on each executor to keep track of whether it's available for reuse.

The previous flow of events was:

1. Set the function result
2. Vacate the scheduler slot
3. Release the executor claim

If something is waiting on the task result, it can receive the result and submit another instance of the same task _between_ steps 1 and 2 or steps 2 and 3. 

Both cause problems, if it's before 2 then the scheduler will wrongly assume one of its slots is still occupied. If it's before 3, 
the scheduler knows the right number of slots, but still can't reuse the executor.

In fact, the sequence of events should be reversed:

1. Release the executor claim 
2. Vacate the scheduler slot 
3. Set the function result

At the time of writing the initial race condition is reproduced fairly reliably in our Github Actions build; the test case for doing so is also included in this PR.

### DIFF
--- a/tests/dist/fixtures.h
+++ b/tests/dist/fixtures.h
@@ -14,16 +14,11 @@ class DistTestsFixture
   , public ConfTestFixture
   , public SnapshotTestFixture
 {
-  protected:
-    std::set<std::string> otherHosts;
-
   public:
     DistTestsFixture()
     {
         // Get other hosts
         std::string thisHost = conf.endpointHost;
-        otherHosts = sch.getAvailableHosts();
-        otherHosts.erase(thisHost);
 
         // Set up executor
         std::shared_ptr<tests::DistTestExecutorFactory> fac =
--- a/tests/test/scheduler/test_scheduler.cpp
+++ b/tests/test/scheduler/test_scheduler.cpp
@@ -70,6 +70,27 @@ class SlowExecutorFixture
     };
 };
 
+class DummyExecutorFixture
+  : public RedisTestFixture
+  , public SchedulerTestFixture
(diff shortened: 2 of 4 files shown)


tests/dist/fixtures.h
```diff
@@ -14,16 +14,11 @@ class DistTestsFixture
   , public ConfTestFixture
   , public SnapshotTestFixture
 {
-  protected:
-    std::set<std::string> otherHosts;
```
Shillaker, line 18: I noticed in passing that this isn't actually used anywhere so I removed it as part of this PR. Happy to move elsewhere but it's just 5 lines in this one file.

tests/test/scheduler/test_scheduler.cpp
```diff
@@ -783,4 +804,30 @@ TEST_CASE_METHOD(SlowExecutorFixture,
       std::get<3>(actualTuple);
     REQUIRE(actualDiffs.size() == diffs.size());
 }
+
+TEST_CASE_METHOD(DummyExecutorFixture, "Test executor reuse", "[scheduler]")
```
Shillaker, line 808: This is the sense check I mention in the comment. It was the first thing I wanted to double check when looking into this issue.

csegarragonz (approved):

PR #121, opened 2021-06-18
Fix hanging in transport layer

Shillaker:
This PR attempts to fix the issues causing the application to hang indefinitely. This was apparently due to us making some mistakes with the lifecycle of ZeroMQ sockets in our transport layer, and also the use of fragile timeouts in our tests.

AFAICT we must ensure that:

- The ZeroMQ context outlives the lifetime of all sockets (i.e. sockets must be constructed and destructed after the context has been opened, but before the context is closed). Note that it is insufficient just to _close_ sockets before the context is closed, their destructors must have been called. If this isn't the case, the context shutdown will hang.
- Any two local sockets performing communication must not be destructed before the other socket has completed its side of the communcation (e.g. we cannot destruct a push socket before the pull socket has finished its recv call).
- We do not rely on the destructors of any global variables, as this violates the first point (a global destructor will be called after we've closed the context). If we do have any sockets (or pointers to them) existing in a global scope, we must force their destructors to be called when we're finished with them.
- We avoid sleeps to reason about threaded operations wherever possible, instead using latches or other synchronisation primitives. If we absolutely must have a sleep, we need to wrap it in a retry to tolerate the underlying operations being several times slower in different environments (i.e. slow CI servers).

Previously we were creating multiple contexts which made it difficult to reason about these rules, so this PR switches to using only one. We were using ephemeral sockets which would occasionally break the second rule, i.e. be destructed before their communication was complete. Finally we had the ZeroMQ contexts living in a global scope, as well as sockets related to MPI living in a thread-local global scope; these could not be removed, but we now make sure they are cleared up properly.

The changes in this PR are:

- Switch to using [cppzmq](https://github.com/zeromq/cppzmq) RAII semantics for managing socket and context shutdown. Previously we were manually closing down sockets, but the library will handle this automatically.
- Use REQ/REP sockets where we need synchronous client/ server behaviour, and PULL/PUSH where we just need to send and forget. Previously we were only using PULL/PUSH and rolling our own synchronous response mechanism which necessitated an ephemeral PULL socket on the client. This caused port conflicts when clients were running concurrently.
- Change `MessageEndpointServer` and `MessageEndpointClient` to have two sockets each, one for async, one for sync.
- Go from two types of endpoint (`SendMessageEndpoint`/ `RecvMessageEndpoint`) to four (`SyncSendMessageEndpoint`, `AsyncSendMessageEndpoint`, `SyncRecvMessageEndpoint`, `AsyncRecvMessageEndpoint`).
- Subclasses of `MessageEndpointServer` now have to implement `doAsyncRecv` and `doSyncRecv` accordingly. `doSyncRecv` returns the response message. Subclasses of `MessageEndpointClient` can call `syncSend` and `asyncSend`, with `syncSend` taking a pointer to a response which gets populated.
- Use a single, global message context only ever opened and closed once and used by all threads. Previously we were creating and destroying contexts regularly and using this as a signal for when things were shut down. 
- Use an explicit shutdown message to close a server. This makes it easier to trace where and when servers get closed.
- Remove `MessageContext` wrapper object and just use the basic `zmq::context_t`. Without the need to use contexts for closing things down, we just need a shared pointer to the global context. 
- Add a linger to every socket, this avoids one source of infinite hangs when sockets are still in a valid state with outstanding messages.
- Remove the `MessageEndpoint::open` method and merge the opening of the socket into the constructor (it was always called straight afterwards anyway).
- Remove the need for "persistent" `Message`s by copying the message body in the snapshot server. This simplifies the `Message` class and means we can use a stdlib `vector` rather than a raw pointer, which removes complexity around message lifecycle. We can introduce a zero-copy recv approach for the snapshot server in a later PR.
- Standardise error checking around zmq operations in the `MessageEndpoint` subclasses.
- Remove the `DummyStateServer` used in tests, moved to a simpler test fixture in a single test file. This removes another moving part in the tests (and makes debugging the state tests much easier).
- Remove sleeps where possible, replacing with a latch. Wrap any remaining sleeps in at least one retry, including the new `REQUIRE_RETRY` macro used in tests.
- Change the existing `Barrier` class to a simpler `Latch` shared between threads via a `shared_ptr`. 

MPI-specific:

- Move sending and receiving of the rank-to-host mapping into the `MpiWorld` class and avoid ephemeral sockets. 
- Split the broadcasting of rank-to-host messages into a separate function. Previously this was rolled into the world creation, and would still send messages in tests even when mock mode was on. The amount of messaging done implicitly in the world creation was getting difficult to manage, so splitting this out also makes it clearer.
- Refactor the MPI "remote" server tests, which were hanging regularly and using slightly confusing terminology around "remote" vs. "local", when in fact both worlds are running locally (one is just in a different thread).

We now manage the zmq context in a global `shared_ptr` which works fine until it comes to destruction. It seems that cppzmq doesn't like being destructed so late in the application lifecycle and prints an error message along the lines of:

```
Invalid argument ... (.../mutex.hpp:123)
```

The solution to this is to explicitly close the context from the `main` method. This means we'll have to add the opening and closing of the zmq context to any `main` method using Faabric for now. I've added a check in `getGlobalMessageContext` which errors if it hasn't been initialised, which should prompt us into adding it where necessary.


include/faabric/transport/MessageEndpoint.h
```diff
@@ -72,13 +72,18 @@ class MessageEndpoint
   protected:
     const std::string host;
     const int port;
+    const std::string address;
```
Shillaker, line 75: This address can be worked out at construction time so I made it a `const` member

include/faabric/transport/MessageEndpointServer.h
```diff
@@ -67,6 +67,8 @@ class MessageEndpointServer
   private:
     const int port;
 
+    std::unique_ptr<RecvMessageEndpoint> endpoint = nullptr;
```
Shillaker, line 70: This fell out of the refactor away from using the message context to shut down. Now the endpoint server can manage its own endpoint internally.

src/transport/MessageEndpoint.cpp
```diff
@@ -1,13 +1,36 @@
...
+              "Got ZeroMQ ETERM for {} on address {}", label, address);        \
+        } else {                                                               \
+            SPDLOG_ERROR("Caught ZeroMQ error for {} on address {}: {} ({})",  \
+                         label,                                                \
+                         address,                                              \
+                         e.num(),                                              \
+                         e.what());                                            \
+            throw;                                                             \
+        }                                                                      \
+    }
```
Shillaker, line 23: This standardises the error checking we do around all ZMQ calls

src/transport/MessageEndpoint.cpp
```diff
@@ -88,171 +95,163 @@ void MessageEndpoint::open(faabric::transport::MessageContext& context,
...
-                SPDLOG_ERROR("Sent different bytes than expected (sent "
-                             "{}, expected {})",
-                             res.value_or(0),
-                             msgSize);
-                throw std::runtime_error("Error sending message");
-            }
-        } catch (zmq::error_t& e) {
-            SPDLOG_ERROR("Error sending message: {}", e.what());
-            throw;
-        }
```
Shillaker, line 122: This `if`/ `else` was essentially doing the same `send` operation in both branches, just with different `send_flags`, so I merged it into one.

src/transport/MessageEndpoint.cpp
```diff
@@ -88,171 +95,163 @@ void MessageEndpoint::open(faabric::transport::MessageContext& context,
...
-            if (e.num() == ZMQ_ETERM) {
-                // Return empty message to signify termination
-                SPDLOG_TRACE("Shutting endpoint down after receiving ETERM");
-                return Message();
-            }
-
-            // Print default message and rethrow
-            SPDLOG_ERROR("Error receiving message: {} ({})", e.num(), e.what());
-            throw;
-        }
```
Shillaker, line 163: To make things simpler to parse I split `recv` up into two methods based on whether a `size` is provided.

include/faabric/transport/Message.h
```diff
@@ -18,22 +18,19 @@ class Message
 
     Message();
 
-    ~Message();
-
     char* data();
 
     uint8_t* udata();
 
+    std::vector<uint8_t> dataCopy();
```
Shillaker, line 25: Convenience method for getting a copy of the data.

include/faabric/transport/MessageEndpoint.h
```diff
@@ -19,12 +18,12 @@
...
+// which also determines how long the context will hang for when closing if
+// things haven't yet completed (usually only when there's an error).
+#define LINGER_MS 1000
+
 namespace faabric::transport {
-enum class SocketType
-{
-    PUSH,
-    PULL
-};
```
Shillaker, line 27: We don't need this enum now that we're managing all socket types within this file.

include/faabric/transport/MessageEndpointServer.h
```diff
@@ -22,27 +20,14 @@ class MessageEndpointServer
...
-    /* Start and stop the server
-     *
-     * Generic methods to start and stop a message endpoint server. They take
-     * a, thread-safe, 0MQ context as an argument. The stop method will block
-     * until _all_ sockets within the context have been closed. Sockets blocking
-     * on a `recv` will be interrupted with ETERM upon context closure.
-     */
-    void start(faabric::transport::MessageContext& context);
-
-    void stop(faabric::transport::MessageContext& context);
```
Shillaker, line 34: This is just removing the `start` and `stop` methods which took a context as argument.

include/faabric/transport/MessageEndpointServer.h
```diff
@@ -53,17 +38,6 @@ class MessageEndpointServer
     virtual void doRecv(faabric::transport::Message& header,
                         faabric::transport::Message& body) = 0;
 
-    /* Send response to the client
-     *
-     * Send a one-off response to a client identified by host:port pair.
-     * Together with a blocking recv at the client side, this
-     * method can be used to achieve synchronous client-server communication.
-     */
-    void sendResponse(uint8_t* serialisedMsg,
```
Shillaker, line 62: `sendResponse` moved to `RecvMessageEndpoint`.

src/transport/MessageEndpoint.cpp
```diff
@@ -1,257 +1,78 @@
...
-}
+    // Create the socket
+    CATCH_ZMQ_ERR(socket =
+                    zmq::socket_t(*getGlobalMessageContext(), socketType),
+                  "socket_create")
 
-void MessageEndpoint::open(faabric::transport::MessageContext& context,
-                           faabric::transport::SocketType sockType,
-                           bool bind)
-{
```
Shillaker, line 26: Opening has been moved to the constructor.

src/transport/MessageEndpoint.cpp
```diff
@@ -1,257 +1,78 @@
...
+                         port,
+                         timeoutMs);
+            CATCH_ZMQ_ERR(socket.bind(address), "bind")
+            break;
         }
-
-        std::string address =
-          "tcp://" + this->host + ":" + std::to_string(this->port);
-
-        // We duplicate the call to close() because when unbinding, we want to
```
Shillaker, line 202: All the closing logic has gone, and is now handled by cppzmq.

tests/test/state/test_state.cpp
```diff
@@ -387,100 +487,58 @@ TEST_CASE("Test push partial with mask", "[state]")
...
-        localKv->pull();
-        localKv->get(actual.data());
-        REQUIRE(actual == newValues);
-    } else {
-        // Check local unchanged without another pull
-        localKv->get(actual.data());
-        REQUIRE(actual == values);
-    }
-
-    server.stop();
```
Shillaker, line 434: Moved `checkPulling` to a function on the fixture.

src/transport/MessageEndpoint.cpp
```diff
@@ -265,70 +198,124 @@ int MessageEndpoint::getPort()
...
 }
 
-void MessageEndpoint::setRecvTimeoutMs(int value)
+void AsyncSendMessageEndpoint::sendShutdown()
 {
-    validateTimeout(value);
-    recvTimeoutMs = value;
+    int header = -1;
+    uint8_t headerBytes = static_cast<uint8_t>(header);
+    doSend(pushSocket, &headerBytes, sizeof(headerBytes), false);
```
Shillaker, line 223: Shutdown is now indicated by sending a header of -1 with "more" set to false.

src/state/StateKeyValue.cpp
```diff
@@ -559,7 +559,7 @@ uint32_t StateKeyValue::waitOnRedisRemoteLock(const std::string& redisKey)
             break;
         }
 
-        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        SLEEP_MS(1);
```
Shillaker, line 562: I was dealing with so many sleeps and writing `usleep(123 * 1000)` that I made a macro.

tests/test/scheduler/test_mpi_world.cpp
```diff
@@ -63,22 +63,6 @@ TEST_CASE_METHOD(MpiBaseTestFixture, "Test creating world of size 1", "[mpi]")
     world.destroy();
 }
 
-TEST_CASE_METHOD(MpiTestFixture, "Test world loading from msg", "[mpi]")
```
Shillaker, line 66: This was the only thing requiring `initialiseFromMsg` to take an extra boolean, and given that the function is implicitly tested all over the place, I thought it wasn't worth it.

include/faabric/transport/MessageEndpointServer.h
```diff
@@ -1,72 +1,49 @@
...
 
-    /* Common start and stop entrypoint
-     *
-     * Call the generic methods with the default global message context.
-     */
-    void start();
+    void setAsyncLatch();
 
-    virtual void stop();
+    void awaitAsyncLatch();
```
Shillaker, line 25: These `setAsyncLatch` and `awaitAsyncLatch` functions are to support testing. Tests frequently check the side effects of sending an async message, then have to add some arbitrary sleep until waiting to check that it's done. This allows us to avoid the sleep and the inherent flakiness of that approach.

csegarragonz (changes requested):
What a beast of a PR, and what a grueling effort as well! Fingers crossed we can now stop re-running the tests until they pass :rofl:  

Seeing this finished, the changes are actually quite easy to follow.

I've left very minor comments. I guess my biggest concern is with managing a lot of `google::protobuf::Message` objects, which is not very clear to me why. Other than that, nicely done, LGTM :+1:

include/faabric/state/StateServer.h
```diff
@@ -13,27 +13,40 @@ class StateServer final : public faabric::transport::MessageEndpointServer
...
 
-    void recvSize(faabric::transport::Message& body);
+    // Sync methods
 
-    void recvPull(faabric::transport::Message& body);
+    std::unique_ptr<google::protobuf::Message> recvSize(
+      faabric::transport::Message& body);
 
-    void recvPush(faabric::transport::Message& body);
+    std::unique_ptr<google::protobuf::Message> recvPull(
```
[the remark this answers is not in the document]
Shillaker: Yes that's a good point, I'll revisit this and see if I can standardise it a bit. I'd like the rule to be: our `Message` wrapper inside the transport layer, and protobuf/ flatbuffers everywhere else.

include/faabric/transport/MessageEndpointServer.h
```diff
@@ -1,72 +1,69 @@
...
+// one for asynchronous. Each is run inside its own background thread.
+class MessageEndpointServer;
+
+class MessageEndpointServerThread
 {
   public:
-    MessageEndpointServer(int portIn);
+    MessageEndpointServerThread(MessageEndpointServer* serverIn, bool asyncIn);
+
+    void start(std::shared_ptr<faabric::util::Latch> latch);
```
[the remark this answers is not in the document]
Shillaker: The latch here is used to make sure that the `start` method of the server doesn't finish until both sockets have been opened. Without it you can never be sure that the server is really ready, so have to add an arbitrary timeout afterwards. In this case I don't think this is just good for testing, it's how I'd expect it to behave in all scenarios. I'll add a comment to that effect.

In addition, this is part of the `MessageEndpointServerThread` class which is internal to the `MessageEndpointServer`, so a little bit of oddness is ok as it's contained. 

Re. the testing point you are right though and it's always a balance. The other stuff in here with latches _is_ to do with testing (i.e. letting a test wait on the completion of the next async operation). There are probably more elegant ways to do this, but equally given that the code is so in flux at the moment I wouldn't worry too much.

include/faabric/transport/MessageEndpointServer.h
```diff
@@ -1,72 +1,69 @@
...
-                      const std::string& returnHost,
-                      int returnPort);
+    virtual void doAsyncRecv(faabric::transport::Message& header,
+                             faabric::transport::Message& body) = 0;
+
+    virtual std::unique_ptr<google::protobuf::Message> doSyncRecv(
+      faabric::transport::Message& header,
+      faabric::transport::Message& body) = 0;
+
+    void sendSyncResponse(google::protobuf::Message* resp);
```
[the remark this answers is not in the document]
Shillaker: Good point, as said above I'll look at this and standardise the interface.

include/faabric/util/latch.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
```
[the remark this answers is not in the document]
Shillaker: Atm we're not actually up to C++ 20 but you're right that that should be a drop-in replacement when we do switch.

include/faabric/util/latch.h
```diff
@@ -0,0 +1,33 @@
...
+    // WARNING: this latch must be shared between threads using a shared
+    // pointer, otherwise there seems to be some nasty race conditions related
+    // to its destruction.
+    static std::shared_ptr<Latch> create(
+      int count,
+      int timeoutMs = DEFAULT_LATCH_TIMEOUT_MS);
+
+    void wait();
+
+    explicit Latch(int countIn, int timeoutMsIn);
```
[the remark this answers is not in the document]
Shillaker: Yep, also a good point.

src/scheduler/FunctionCallServer.cpp
```diff
@@ -1,62 +1,71 @@
...
-                                faabric::transport::Message& body)
+void FunctionCallServer::doAsyncRecv(faabric::transport::Message& header,
+                                     faabric::transport::Message& body)
 {
     assert(header.size() == sizeof(uint8_t));
     uint8_t call = static_cast<uint8_t>(*header.data());
     switch (call) {
-        case faabric::scheduler::FunctionCalls::Flush:
-            this->recvFlush(body);
+        case faabric::scheduler::FunctionCalls::ExecuteFunctions: {
```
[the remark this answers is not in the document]
Shillaker: I had a look and I didn't spot any rules. Might be more of a clang-tidy thing unfortunately (or I wasn't looking hard enough).

src/scheduler/FunctionCallServer.cpp
```diff
@@ -1,62 +1,71 @@
...
+{
     // Clear out any cached state
     faabric::state::getGlobalState().forceClearAll(false);
 
     // Clear the scheduler
     scheduler.flushLocally();
 
-    faabric::EmptyResponse response;
-    SEND_SERVER_RESPONSE(response, msg.returnhost(), FUNCTION_CALL_PORT)
+    return std::make_unique<faabric::EmptyResponse>();
```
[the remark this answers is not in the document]
Shillaker: Re. comment above, I think `transport::Message` should be contained within the `transport` module, but I'll see what makes sense when I've tidied up the interfaces.

src/scheduler/FunctionCallServer.cpp
```diff
@@ -1,62 +1,71 @@
...
 }
 
 void FunctionCallServer::recvExecuteFunctions(faabric::transport::Message& body)
```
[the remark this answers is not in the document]
Shillaker: Yes that's fair enough, now that we're both familiar with the changes it probably won't be too bad.

src/scheduler/MpiWorld.cpp
```diff
@@ -4,28 +4,87 @@
...
 static thread_local std::set<int> iSendRequests;
+
 static thread_local std::map<int, std::pair<int, int>> reqIdToRanks;
 
+// These long-lived sockets are used by each world to communicate rank-to-host
+// mappings. They are thread-local to ensure separation between concurrent
+// worlds executing on the same host
+static thread_local std::unique_ptr<
+  faabric::transport::AsyncRecvMessageEndpoint>
+  ranksRecvEndpoint;
```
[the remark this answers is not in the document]
Shillaker: Yes I went back and forth on these endpoints, they can't be ephemeral otherwise we have into issues running locally when the sender might get destructed before the receiver and it hangs. For this reason they have to be linked to a scope higher in the stack, but they can't be tied to the world itself as the thread that creates it might not be the one that ends up interacting with the socket. Therefore they ended up living in this TLS.

src/scheduler/SnapshotClient.cpp
```diff
@@ -68,24 +69,26 @@ void clearMockSnapshotRequests()
 // Snapshot client
 // -----------------------------------
 
-#define SEND_FB_REQUEST(T)                                                     \
-    sendHeader(T);                                                             \
-    mb.Finish(requestOffset);                                                  \
-    uint8_t* buffer = mb.GetBufferPointer();                                   \
-    int size = mb.GetSize();                                                   \
-    send(buffer, size);
+#define SEND_FB_MSG(T, mb)                                                     \
```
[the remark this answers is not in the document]
Shillaker: Yep, good point.

src/transport/context.cpp
```diff
@@ -0,0 +1,48 @@
...
+        return;
+    }
+
+    SPDLOG_TRACE("Destroying global ZeroMQ context");
+
+    // Force outstanding ops to return ETERM
+    instance->shutdown();
+
+    // Close the context
+    instance->close();
```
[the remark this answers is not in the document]
Shillaker: Once this has been closed we can just leave the standard object lifecycle to call the destructor in this case.

src/transport/MessageEndpoint.cpp
```diff
@@ -1,258 +1,208 @@
...
+    } catch (zmq::error_t & e) {                                               \
+        SPDLOG_ERROR("Caught ZeroMQ error for {} on address {}: {} ({})",      \
+                     label,                                                    \
+                     address,                                                  \
+                     e.num(),                                                  \
+                     e.what());                                                \
+        throw;                                                                 \
+    }
+
+#define CATCH_ZMQ_ERR_RETRY(op, label)                                         \
```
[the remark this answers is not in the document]
Shillaker: Yes that seems sensible.

src/transport/Message.cpp
```diff
@@ -1,57 +1,47 @@
+#include <cstddef>
```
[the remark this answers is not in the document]
Shillaker: I think it's because I removed an include from a header higher up the chain, and this is actually the lowest point that it's needed, but I'll double check.

PR #122, opened 2021-06-23
Changing zeromq dependency, and updating deprecated socket option

csegarragonz:
I originally [forked the C++ binding for zeromq](https://github.com/faasm/cppzmq) because I couldn't manage to disable a CMake option from our `CMakeLists.txt`. After some re-factoring in later PRs, this is in fact working, so we can change the dependency.

When the PR is merged in I will delete the fork.

### DIFF
--- a/src/transport/MessageEndpoint.cpp
+++ b/src/transport/MessageEndpoint.cpp
@@ -81,8 +81,8 @@ void MessageEndpoint::open(faabric::transport::MessageContext& context,
     }
 
     // Set socket options
-    this->socket->setsockopt(ZMQ_RCVTIMEO, recvTimeoutMs);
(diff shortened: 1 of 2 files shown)


src/transport/MessageEndpoint.cpp
```diff
@@ -81,8 +81,8 @@ void MessageEndpoint::open(faabric::transport::MessageContext& context,
     }
 
     // Set socket options
-    this->socket->setsockopt(ZMQ_RCVTIMEO, recvTimeoutMs);
```
csegarragonz, line 84: This outputed a deprecated warning when I compiled it. The line in the source code where this is depcreated is [here](https://github.com/zeromq/cppzmq/blob/8e6066741349f4e9599d6178c432db9c82cfcdd2/zmq.hpp#L1655).

Shillaker (approved):

PR #126, opened 2021-07-05
Push snapshot diffs before reset, and thread results after

Shillaker:
When a thread completes on a remote host, we need to do three things: reset the executor, send the snapshot diffs, send the thread's results. Previously sending snapshot diffs was bundled with sending thread results, and done _after_ resetting the executor. However, this doesn't make sense as the resetting may wipe out any changes that should be captured in the diffs.

Therefore, we need to send the diffs, reset the executor, then send the thread results (note that the thread results must always be sent _after_ resetting the executor to ensure it's ready to be reused). There is already a stand-alone snapshot diffs message, so this PR switches to using that and removes the snapshot diffs from the thread results message. 

The downside here is that completing a thread on a remote host now requires sending two messages rather than one, but this is unavoidable. 

Changes:

- Rearranged `Executor` code to push snapshot diffs before calling `reset`
- Switched to using dedicated snapshot diffs message rather than thread results message (for diffs)
- Added missing checks for zero-sized snapshots
- Added distributed test to catch the original issue (using distributed threads)


include/faabric/snapshot/SnapshotServer.h
```diff
@@ -30,10 +30,5 @@ class SnapshotServer final : public faabric::transport::MessageEndpointServer
     void recvDeleteSnapshot(const uint8_t* buffer, size_t bufferSize);
 
     void recvThreadResult(const uint8_t* buffer, size_t bufferSize);
-
-  private:
-    void applyDiffsToSnapshot(
-      const std::string& snapshotKey,
-      const flatbuffers::Vector<flatbuffers::Offset<SnapshotDiffChunk>>* diffs);
```
Shillaker, line 37: This was never implemented/ used

src/scheduler/Executor.cpp
```diff
@@ -221,21 +224,21 @@ void Executor::threadPoolThread(int threadPoolIdx)
         bool isLastTask = oldTaskCount == 1;
 
         SPDLOG_TRACE("Task {} finished by thread {}:{} ({} left)",
-                     msg.id(),
+                     faabric::util::funcToString(msg, true),
                      id,
                      threadPoolIdx,
-                     executingTaskCount);
+                     oldTaskCount - 1);
```
Shillaker, line 230: This logging wasn't thread-safe by accessing `executingTaskCount` directly.

tests/dist/scheduler/functions.cpp
```diff
@@ -74,6 +77,108 @@ int handleFakeDiffsFunction(faabric::scheduler::Executor* exec,
...
+        req->set_type(faabric::BatchExecuteRequest::THREADS);
+
+        for (int i = 0; i < nThreads; i++) {
+            auto& m = req->mutable_messages()->at(i);
+            m.set_appindex(i);
+            m.set_inputdata(std::string("thread_" + std::to_string(i)));
+            m.set_snapshotkey(snapshotKey);
+        }
+
+        // Dispatch the message, we expect them all to be executed other hosts
```
csegarragonz, line 121: small typo "executed on other hosts"

csegarragonz (approved):
LGTM

tests/dist/scheduler/functions.cpp
Shillaker: Good spot, will fix in next PR

PR #128, opened 2021-07-13
Fix segmentation fault on big message serialization

kubasz:
I'm not entirely sure why the original code crashes with SIGSEGV on some messages in my testing, but Variable-Length Arrays aren't generally standard C++ anyway and this fix also has the benefit of having one less copy of the serialized message.


Shillaker:
Nice one, LGTM, cheers @kubasz !

PR #129, opened 2021-07-14
Fine-grained snapshot diffs

Shillaker:
Previously snapshot diffs were only down to a page-level granularity, which meant that distributed threads modifying a shared array would trample on each other (as they would all have their own modified version of the same page).

This PR adds a couple of steps to the existing snapshot diff process:

- Push the snapshot to worker hosts before thread execution
- Map thread memory from the snapshot
- Reset dirty page checks
- Execute the thread
- Get dirty pages
- _Perform a bytewise comparison of the thread's memory with the original snapshot_
- _Push the bytewise diffs back to the original host_ (instead of the page-wise diffs)

Changes:

- Add logic for working out bytewise diffs based on dirty pages
- Encapsulate all the batch-level variables in an executor inside an `ExeutorTask` class (instead of a mix of a `std::pair` and class members on the `Executor`)
- Move logic related to applying diffs into the `SnapshotData` class 
- Share test code via `SnapshotTestFixture`
- Check for zero-size snapshots (previously missing)

### DIFF
--- a/src/snapshot/SnapshotRegistry.cpp
+++ b/src/snapshot/SnapshotRegistry.cpp
@@ -12,6 +12,11 @@ SnapshotRegistry::SnapshotRegistry() {}
 faabric::util::SnapshotData& SnapshotRegistry::getSnapshot(
   const std::string& key)
 {
+    if (key.empty()) {
+        SPDLOG_ERROR("Attempting to get snapshot with empty key");
+        throw std::runtime_error("Getting snapshot with empty key");
+    }
+
     if (snapshotMap.count(key) == 0) {
         SPDLOG_ERROR("Snapshot for {} does not exist", key);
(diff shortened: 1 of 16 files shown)


src/snapshot/SnapshotRegistry.cpp
```diff
@@ -12,6 +12,11 @@ SnapshotRegistry::SnapshotRegistry() {}
 faabric::util::SnapshotData& SnapshotRegistry::getSnapshot(
   const std::string& key)
 {
+    if (key.empty()) {
```
csegarragonz, line 15: AFAICT (I could have missed it), this is not tested in the tests. Could you add a very simple check for attempting to get a snapshot with an empty key?

csegarragonz (approved):
LGTM, very minor couple things.

PR #138, opened 2021-09-08
Fix race condition in executor test

Shillaker:
This has been hanging around for a while and periodically failing.


tests/test/scheduler/test_executor.cpp
```diff
@@ -106,6 +106,8 @@ class TestExecutor final : public Executor
                 nThreads = std::stoi(msg.inputdata());
             }
 
+            SPDLOG_DEBUG("TestExecutor spawning {} threads", nThreads);
```
Shillaker, line 109: This test was really hard to debug without this logging, which may prove useful to someone in future.

csegarragonz (approved):

PR #141, opened 2021-09-13
Add generic distributed synchronous operations

Shillaker:


include/faabric/scheduler/DistributedCoordination.h
```diff
@@ -0,0 +1,81 @@
+#pragma once
```
csegarragonz, line 1: My biggest design concern re. the `DistributedCoordination` class is that it is shared among all applications. Thus forcing us to implement a distributed coordination to access the `DistributedCoordination` which feels a bit unnatural.

I am wondering, could we have a per-app instance of the `DistributedCoordination` class? I imagine there must be some strong argument against this, but just pointing it out. 

Plus, I am assuming that contention will be a problem, but I don't have empirical proof which may be a case of pre-optimisation.

src/scheduler/DistributedCoordination.cpp
```diff
@@ -0,0 +1,236 @@
...
+{
+    doLocalNotify(appId, true);
+}
+
+void DistributedCoordination::notify(const faabric::Message& msg)
+{
+    DISTRIBUTED_SYNC_OP(localNotify, client.coordinationNotify)
+}
+
+void DistributedCoordination::localBarrier(int32_t appId)
```
csegarragonz, line 179: I am wondering if it is a good idea to add timeouts in this operations, to prevent applications waiting on barriers forever.

This has caused issues in the past, so maybe we could design the distributed coordination with this in mind.

csegarragonz (changes requested):
Mainly two design concerns/questions I have, rather than objections on the implementation:
* Would it be possible to have a per-application `DistributedCoordination` instance?
* Do we want to add timeouts to distributed coordination operations to prevent deadlocks?

Happy to discuss this offline

src/scheduler/DistributedCoordination.cpp
```diff
@@ -0,0 +1,236 @@
...
+    if (msg.masterhost() != sch.getThisHost()) {
+        SPDLOG_ERROR("Setting group {} size not on master ({} != {})",
+                     msg.appid(),
+                     msg.masterhost(),
+                     sch.getThisHost());
+
+        throw std::runtime_error("Setting sync group size on non-master");
+    }
+
+    appSizes[msg.appid()] = appSize;
```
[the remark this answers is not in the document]
Shillaker: Unfortunately adding to a map isn't atomic (I guess it needs to do the copy and update the index). We've seen issues in the past where you get either a partially initialised object or a segfault under heavy contention.

include/faabric/scheduler/DistributedCoordination.h
Shillaker: Yes, all good points, although I think coordinating access to the coordinator will always be required, even if it's one per app, as there will be multiple executors from that app on any given host.

Whether contention is a bottleneck will depend on the workload. If it's something CPU-intensive which performs distributed coordination at the end of a big calculation to share results, it won't be a problem. If it's something with an atomic section nested deep in some inner loop that's called thousands of times, then it will be a problem. 

A bigger design flaw that I've spotted is that we actually need multiple coordinators _per app_, to handle nested multi-threaded constructs. As a result, I'll need to change it to look up coordinators on an arbitrary ID, rather than the app ID. I'll re-request a review and update the description once this is done.

src/scheduler/DistributedCoordination.cpp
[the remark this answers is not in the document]
csegarragonz: Sorry, there was a typo in the comment. I meant the map **can't** be concurrently modified.

It looks to me that we don't protect against concurrent modifications here.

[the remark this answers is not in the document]
Shillaker: Haha ah yes, I was looking at this in isolation 😄. In its current state this is ok, as this would only be called once, by a function before spawning a group of chained functions that would do the coordination. However, this bit of code is getting removed in my current rework, so no need to worry.

include/faabric/scheduler/FunctionCallApi.h
```diff
@@ -8,6 +8,9 @@ enum FunctionCalls
     Flush = 2,
     Unregister = 3,
     GetResources = 4,
-    SetThreadResult = 5,
```
Shillaker, line 11: Note that `SetThreadResult` is no longer used.

src/scheduler/DistributedCoordination.cpp
Shillaker: Yes, there are already timeouts implemented in the barriers (explicitly in this class or via the `faabric::util::Barrier` class), but I'll add them to the lock operations.

PR #142, opened 2021-09-16
Snapshot diff merge operations

Shillaker:
The default merge strategy for snapshot diffs is to overwrite the master with the diffs. This works if each distributed function is writing to different offsets (for example, if each function is writing to a different index in an array), but if they are writing to the _same_ offset, the merging will lose information by overwriting each time.

This PR adds merge operations for summation, subtraction, multiplication, min and max, but this list could be extended in future, primarily focused on in-place reductions.

Applications can set these merge operations at byte-level granularity, e.g. specifying that a 4-byte region at offset `X` should be summed as an integer when merging snapshots.

At the moment I've only included integer operations, as these are all that's needed for the current use-case, but again, this could be extended in future.

### DIFF
--- a/src/util/snapshot.cpp
+++ b/src/util/snapshot.cpp
@@ -78,7 +78,7 @@ std::vector<SnapshotDiff> SnapshotData::getChangeDiffs(const uint8_t* updated,
             bool isInMergeRegion =
               mergeIt != mergeRegions.end() &&
               offset >= mergeIt->second.offset &&
-              offset <= (mergeIt->second.offset + mergeIt->second.length);
+              offset < (mergeIt->second.offset + mergeIt->second.length);
 
             if (isDirtyByte && isInMergeRegion) {
                 SnapshotMergeRegion region = mergeIt->second;
@@ -152,9 +152,7 @@ std::vector<SnapshotDiff> SnapshotData::getChangeDiffs(const uint8_t* updated,
 
                 // Bump the loop variable to the end of this region (note that
                 // the loop itself will increment onto the next)
-                int nextOffset = region.offset + region.length;
-                int jump = nextOffset - offset;
-                b += jump;
+                b = (region.offset - pageOffset) + (region.length - 1);
 
                 // Move onto the next merge region
                 ++mergeIt;
--- a/src/util/snapshot.cpp
+++ b/src/util/snapshot.cpp
@@ -4,6 +4,10 @@
 
 namespace faabric::util {
 
+SnapshotData::SnapshotData()
+  : merger(std::make_shared<SnapshotDiffMerger>())
+{}
+
 std::vector<SnapshotDiff> SnapshotData::getDirtyPages()
 {
     if (data == nullptr || size == 0) {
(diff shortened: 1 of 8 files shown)


src/util/snapshot.cpp
```diff
@@ -31,27 +37,128 @@ std::vector<SnapshotDiff> SnapshotData::getDirtyPages()
...
         int offset = pageOffset;
         for (int b = 0; b < HOST_PAGE_SIZE; b++) {
             offset = pageOffset + b;
             bool isDirtyByte = *(data + offset) != *(updated + offset);
-            if (isDirtyByte && !diffInProgress) {
+
+            bool isInMergeRegion =
+              mergeIt != mergeRegions.end() &&
+              offset >= mergeIt->second.offset &&
+              offset <= (mergeIt->second.offset + mergeIt->second.length);
```
csegarragonz, line 81: Minor point here, I think this should be a `<` not a `<=`.

If we have a merge region starting at offset 0 with length 3, we want to mark bytes 0,1,2 as dirty and belonging to this merge region. Byte 3 (i.e. 0 + 3) should be out of the region if I am not mistaken.

csegarragonz: Actually, after having gone through the code, afaict we will only have the case in which `offset == mergeIt->second.offset` as then we move the `offset` by `region.length`.

Maybe we could add an assertion after line 83 like:
```cpp
if (isDirtyByte && isInMergeRegion) {
    assert(offset == mergeIt->second.offset);
...
```

src/util/snapshot.cpp
```diff
@@ -31,27 +37,128 @@ std::vector<SnapshotDiff> SnapshotData::getDirtyPages()
...
+                        throw std::runtime_error(
+                          "Merge region for unhandled data type");
+                    }
+                }
+
+                // Add the diff to the list
+                diffs.emplace_back(diff);
+
+                // Bump the loop variable to the end of this region (note that
+                // the loop itself will increment onto the next)
```
csegarragonz, line 154: Again, I think that `offset + length` should point to the first byte after the region, and then the `for` increment would skip one byte.

A test that would check this would consist of two consecutive regions with `int`s that need to be added. If I am not mistaken that test should now fail.

tests/test/snapshot/test_snapshot_client_server.cpp
```diff
@@ -124,6 +126,169 @@ TEST_CASE_METHOD(SnapshotClientServerFixture,
...
+TEST_CASE_METHOD(SnapshotClientServerFixture,
+                 "Test detailed snapshot diffs with merge ops",
+                 "[snapshot]")
+{
+    // Set up a snapshot
+    std::string snapKey = std::to_string(faabric::util::generateGid());
+    faabric::util::SnapshotData snap = takeSnapshot(snapKey, 5, false);
+
+    // Set up a couple of ints in the snapshot
+    int offsetA1 = 5;
```
csegarragonz, line 138: As mentioned earlier, could we have a test where one diff is at offset `x` and the other one at offset `x + sizeof(int)`?

csegarragonz (changes requested):
Implementation is very clear and easy to follow. Just one note on us skipping a byte at some point in the loop.

tests/test/util/test_snapshot.cpp
```diff
@@ -0,0 +1,413 @@
...
+                                         diffOther.data + diffOther.size);
+    REQUIRE(actualOtherData == otherData);
+
+    deallocatePages(snap.data, snapPages);
+}
+
+TEST_CASE_METHOD(SnapshotTestFixture,
+                 "Test edge-cases of snapshot merge regions",
+                 "[util]")
+{
```
Shillaker, line 129: @csegarragonz here's the test that covers the edge cases

src/util/snapshot.cpp
Shillaker: Yes, great spot on both of these comments. There was an off-by-one error that meant that the second of two adjacent merge regions would be lost if both had changed. Have added the relevant `-1`s and a test.

csegarragonz (approved):

PR #143, opened 2021-09-17
Not use emplace for editing vector elements

csegarragonz:
Pretty simple, silly, yet nasty bug with the MPI implementation.

In two different `std::vector`s (which we previously filled by `emplace_back`-ing `nullptr`s, we used `emplace(<offset>, <thing to be moved>)` to **modify** the element at offset `<offset>`.

Turns out that this is **not** how `emplace` works. Citing [cppreference](https://en.cppreference.com/w/cpp/container/vector/emplace):
> Inserts a new element into the container directly before pos.

Thus, our vector elements were moved all around the place in a non-deterministic manner. The fact that this worked sometimes is because:
* We lazy-initialise these vectors. If the element is not initialised, we create a new one. Thus, every time we created a new element at the right index, and used it (just once). This only happened a limited number of times.
* Given that the rank distribution we test with is always lower-half host 1, upper-half host 2, and the vectors are sparse by nature, overwrites were occasional.

This would have worsen dramatically with more aggressive testing or different rank-host distributions. I checked, and this does not happen anywhere else in the codebase. 

Should have read the reference better before hand.

### DIFF
--- a/include/faabric/scheduler/MpiWorld.h
+++ b/include/faabric/scheduler/MpiWorld.h
@@ -180,8 +180,6 @@ class MpiWorld
 
     double getWTime();
 
-    // ---------- Testing only methods ----------
-
     std::vector<bool> getInitedRemoteMpiEndpoints();
 
     std::vector<bool> getInitedUMB();
--- a/include/faabric/scheduler/MpiWorld.h
+++ b/include/faabric/scheduler/MpiWorld.h
@@ -180,6 +180,12 @@ class MpiWorld
 
     double getWTime();
 
+    // ---------- Testing only methods ----------
+
+    std::vector<bool> getInitedRemoteMpiEndpoints();
+
+    std::vector<bool> getInitedUMB();
+
   private:
     int id = -1;
     int size = -1;
--- a/include/faabric/transport/MessageEndpoint.h
+++ b/include/faabric/transport/MessageEndpoint.h
@@ -33,7 +33,7 @@ class MessageEndpoint
     MessageEndpoint(const std::string& hostIn, int portIn, int timeoutMsIn);
 
     // Delete assignment and copy-constructor as we need to be very careful with
-    // socping and same-thread instantiation
+    // scoping and same-thread instantiation
     MessageEndpoint& operator=(const MessageEndpoint&) = delete;
 
     MessageEndpoint(const MessageEndpoint& ctx) = delete;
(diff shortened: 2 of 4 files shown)


include/faabric/transport/MessageEndpoint.h
```diff
@@ -33,7 +33,7 @@ class MessageEndpoint
     MessageEndpoint(const std::string& hostIn, int portIn, int timeoutMsIn);
 
     // Delete assignment and copy-constructor as we need to be very careful with
-    // socping and same-thread instantiation
+    // scoping and same-thread instantiation
```
csegarragonz, line 36: Caught this re-visiting the sources so figured out i'd just change it.

Shillaker (changes requested):
Good catch. With bugs like this it's good to add a test that would have failed before, but now passes (if possible). In this case perhaps it could combine calls to `initRemoteMpiEndpoint` and `getUnackedMessageBuffer` to check that things work?

include/faabric/scheduler/MpiWorld.h
```diff
@@ -180,6 +180,12 @@ class MpiWorld
 
     double getWTime();
 
+    // ---------- Testing only methods ----------
```
Shillaker, line 183: Although these methods are just used in tests, they're still a valid external API for this class, providing a way to query information on the internal state of the MPI world. As a result, I don't think we need to label them as test-only or have the checks in their implementation. If someone wants to remove them, they'll see they're only used in tests, and can either leave them, or rewrite the tests. Equally if someone wants to call them in other areas of the code, I think that's also ok; they shouldn't throw an exception just because they were originally written to be used in tests.

src/scheduler/MpiWorld.cpp
```diff
@@ -1379,6 +1377,38 @@ double MpiWorld::getWTime()
     return t / 1000.0;
 }
 
+// This method is only used for testing
+std::vector<bool> MpiWorld::getInitedRemoteMpiEndpoints()
+{
+    if (!faabric::util::isTestMode()) {
+        SPDLOG_ERROR("Called method for testing outside test mode");
+        throw std::runtime_error("This method can only be called in testing");
+    }
```
Shillaker, line 1386: As mentioned above, I don't think there's a reason this shouldn't be called outside of a test if someone wants to.

src/scheduler/MpiWorld.cpp
```diff
@@ -1379,6 +1377,38 @@ double MpiWorld::getWTime()
...
+
+    std::vector<bool> retVec(mpiMessageEndpoints.size());
+    for (int i = 0; i < mpiMessageEndpoints.size(); i++) {
+        retVec.at(i) = mpiMessageEndpoints.at(i) != nullptr;
+    }
+
+    return retVec;
+}
+
+// This method is only used for testing
```
Shillaker, line 1396: Same point re. these comments.

Shillaker (commented):
Nice work on the tests. I've added some comments about how I don't think we need to mark them as test-only or throw exceptions, so once that stuff's been removed I think it's good to go.

Shillaker (approved):

PR #147, opened 2021-09-30
Multi-threaded servers

Shillaker:
Using [0MQ router/ dealer pattern](https://zguide.zeromq.org/docs/chapter2/#Multithreading-with-ZeroMQ).


src/transport/MessageEndpointServer.cpp
```diff
@@ -11,150 +13,300 @@ namespace faabric::transport {
...
+    // we don't do this, zmq will direct messages to sockets that are in the
+    // process of shutting down and cause errors.
+    for (int i = 0; i < nThreads; i++) {
+        SPDLOG_TRACE("Sending async shutdown message {}/{} to port {}",
+                     i + 1,
+                     nThreads,
+                     asyncPort);
+
+        setShutdownLatch();
+        asyncShutdownSender.send(shutdownHeader.data(), shutdownHeader.size());
```
csegarragonz, line 259: Maybe add a comment here that the two threads waiting on the `shutdownLatch` are: this one, and the server handler which will wait on the latch once all threads have been shut down.

src/transport/MessageEndpointServer.cpp
```diff
@@ -11,150 +13,300 @@ namespace faabric::transport {
...
+        awaitShutdownLatch();
```
csegarragonz, line 260: It is counter-intuitive that we use this method here, and not in line 165. I imagine it is because `awaitShutdownLatch` nulls out the pointer, but still makes it hard to debug and follow the code.

csegarragonz (commented):
One high-level observation is why are we using a thread-pool, and not assigning a different `inproc` address to each thread in our already existing thread pool.

I am concerned that ZeroMQ's scheduling is out of our control, and we rely on setting an arbitrarily adequate number of working threads, and we could always find a corner-case to break it (as it happened with MPI async layer).
As far as I can tell the current execution flow is (correct me if I am wrong):
* Local executor thread blocks due to a distributed coordination operation.
* Underlying transport thread (randomly assigned to this executor) blocks.
* Remote executor thread unlocks the local executor thread, thus sends a message to notify.
* A different transport thread logically unlocks the local transport thread that was locked.
Main concern with this design is that it strongly depends on the availability of a free transport thread, and on ZeroMQ actually scheduling work to threads that are not blocked, which again is out of our control.

My alternative would be to give each executor in faabric an inproc socket, or use a PUB/SUB scheme over the fan-out in-proc fabric with topic being something like `appid/<function/snapshot/state>/exec-id`.

Then, implement the distributed coordination operations as transport primitives. This is, use ZeroMQ's blocking functionality to implement distributed barriers.

I have only drafted the idea here, happy to dismiss or discuss offline, as I am sure the reasoning needs a lot of polishing.

PR #150, opened 2021-10-05
Change default `MPI_BASE_PORT` and turn into env. variable

csegarragonz:
All tests running on Knative with, specifically, 10 ranks would fail, claiming that the port `9090` was already in use. Sure enough, ports `9090` and `9091` are used by `knative-serving` on the `faasm-worker` service.

The fact that this only happens with 10 ranks (i.e. `MPI_WORLD_SIZE = 10`) is because the ports we use are a combination of the `outbound-inbound` ranks _and_ the world size.

To solve this problem, I change the default MPI base port to a port area, apparently, less cluttered: `10800`. However, for us to actually benefit from this patch we will now have to: 
1. Merge this PR into `faabric`.
2. Bump the faabric dependency in `faasm`.
3. Bump the `faasm` version and re-build containers.

To prevent this from occuring in the future, I also make the `MPI_BASE_PORT` an environment variable in the `faabric` config.

One may argue that the over-arching problem is that, in a multi-tenant environment with big MPI jobs, we end up using too many ports, and errors like this will keep appearing. I could agree with that. And given that we now will start working on multi-tenant MPI jobs, it's about time to think about this in detail. I say we move discussion for this offline.


Shillaker (approved):
Yes this seems like the right approach for now. Let's take this offline, I have some ideas for how we can do this without taking up hundreds of ports.

Issue #156, opened 2021-10-14
Scaling issues


eigenraven:
This is a grab-bag of issues encountered when scaling Faasm and Faabric to many requests per second.

Solutions to most of these problems are implemented in this fork: https://github.com/auto-ndp/faabric

There is a corresponding issue in Faasm: https://github.com/faasm/faasm/issues/504

- [x] Executor initialization is synchronous within the scheduler thread before the child executor thread is spawned (thus blocking the scheduler). Instead it can be moved to just before the thread loop in the Executor.
- [x] ZeroMQ socket explosion - for each Executor we open several sockets to handle the different client/ server pairs. This means we open `n_executors * n_ports_per_executor`, which hits system-wide limits with a lot of Executors. Instead we could multiplex more/ all calls through a single socket. Because Executors are handled with separate threads, it would not be possible to cache the clients (as 0MQ sockets cannot be shared between threads). AFAICT #260 should address this for MPI which was the main culprit.
- [x] ZeroMQ lacking thread-safety - There are a few places where the transport code is (potentially unnecessarily) complex, due to the lack of ZeroMQ thread safety. ZeroMQ's successor (x2) [nng](https://github.com/nanomsg/nng) could be used as a thread-safe replacement. (Done in #286)
- [x] The current HTTP Endpoint implementation can quickly become a bottleneck when serving lots of external requests. An alternative, async HTTP implementation based on [Boost beast](https://www.boost.org/doc/libs/1_73_0/libs/beast/doc/html/index.html) has worked well and can handle thousands of concurrent connections per worker thread. (Done in #274)
- [ ] The use of Redis to return function results is unnecessary when the call is synchronous, as it just needs to be returned to the calling host. This can be done with a direct ZeroMQ message (similar to how we handle thread results). 
- [x] Use of Redis for discovery may also be unnecessary if we can have a proxy running somewhere in the cluster, or use some form of broadcast (#300)
- [x] Executor shutdown doesn't clean up resources, it just moves the Executor to a vector of dead Executors. This is done to avoid some kind of deadlock (according to the comments), but it causes a serious memory leak if the Executors load a lot of data. There should be a way to prune these dead executors, either based on time or the overall function lifecycle. #252 
[...]

PR #157, opened 2021-10-15
A large number of performance changes (related to #156)

eigenraven:
* Add ability for Executors to run setup() on thread pool start rather than construction
* Boost Beast-based async HTTP endpoint (Redis-related waiting is synchronous, async redis api exists but is not used yet in this PR as it should be bypassed by most HTTP-related requests)
* Redis bypass for fully local calls and calls explicitly setting directresulthost
* "Soft shutdown" for executors to free up some resources immediately on threadpool exit
* Executors ignore tasks if their timeout has already passed by the time they are in the front of the queue
* Optional scheduler monitoring file that provides regularly-updated counters of tasks in the system

This is a fairly straightforward cherrypicked set of changes from my fork of faabric, I'm open to discussing api changes/tests/etc., otherwise it is ready to merge if you'd like.


eigenraven:
This is very out of date, I might still extract out individual bits of my changes into smaller PRs and submit them later.

PR #169, opened 2021-11-01
Scheduler hints and fixes for distributed coordination bugs

Shillaker:
This is a mixed bag, all changes required to get a larger OpenMP application working and do some tidy-up along the way. 

### Scheduler hints

At the moment "hints" are taken on face value by the scheduler (so they're more like scheduler plans). This is useful in a multithreaded application that repeatedly performs an operation using the same number of threads and doesn't need to keep repeating the scheduling decision. How applications decide to cache the scheduling decision is up to them. This change is relatively simple, and just involves adding one more scheduler method:

```cpp
faabric::util::SchedulingDecision callFunctions(
    std::shared_ptr<faabric::BatchExecRequest> req, 
    const faabric::util::SchedulingDecision &hint
    );
```

Where `hint` is the proposed scheduling decision. To implement this I just refactored the scheduling logic into two parts: (i) calculating the decision (only performed when a hint _isn't_ passed); (ii) executing a decision (performed both when a hint is and isn't passed).

### Snapshot diff whitelisting

Snapshot merge regions are now specified as a whitelist rather than a blacklist. This makes more sense, as we were adding a lot of logic around ignoring and doing default diffs, when in fact applications will know which regions they want to merge. The "worst case" is for a generic threaded application with no consistency guarantees where it would want to merge the whole heap, in which case it could specify a single merge region for it. 

This rendered `Ignore` regions obsolete, so I removed them. 

### Bug fixes

_Deadlock and inconsistencies with waiting for point-to-point groups to be enabled_ - added more checks in `PointToPointBroker`, and the `FlagWaiter` class, which allows an arbitrary number of threads to wait on a boolean flag set by another (can't use a barrier as we don't know in advance how many threads will need to wait). 

_Nested thread groups deadlocking when scheduled on same thread pool thread as their parent_ - if thread `A` spawns a nested group of threads `B1` and `B2`, we must ensure that neither `B1` nor `B2` is queued on the same thread pool thread as `A` (as `A` is dependent on the completion of `B1` and `B2`, but they wouldn't execute until `A` is finished). Previously the thread pool scheduling could allow this even when the `Executor` was not overloaded. This PR changes the thread pool allocation to allocate one task to every free thread pool thread before we start overloading them, and when we do, we avoid indexes `0` and `1` as these are likely to contain blocking threads. This doesn't completely remove the problem, but ensures it won't happen under normal (i.e. un-overloaded) execution.

_Locking on non-existent groups on snapshot pushes_ - locking in the `SnapshotServer` only needs to happen on the master host when snapshot _diffs_ are being returned. However, previously the `SnapshotServer` tried to lock on the group on _any_ snapshot or snapshot diff push, causing an error when the group didn't exist (i.e. on non-master hosts). It now only happens on snapshot diff pushes when the group exists on that host.

_Snapshot diffing inefficiency_ - previously this was going through all the dirty pages byte-by-byte, checking them against merge regions, and adding a default overwrite mode. Instead, it's much more efficient to _only_ do merging in specified merge regions, and by default do no merging. This makes the code cleaner, and the diffing process more efficient. 

_Snapshot skipping missing diffs_ - previously executors would only restore a snapshot once, however, this actually needs to be done every time, as the snapshot may have changed between executions. Ideally we'd apply the diffs directly to all the mapped regions rather than have to re-restore from the original snapshot, but this is left as a potential optimisation for the future.


include/faabric/util/snapshot.h
```diff
@@ -67,6 +62,19 @@ class SnapshotDiff
...
+  public:
+    uint32_t offset = 0;
+    size_t length = 0;
+    SnapshotDataType dataType = SnapshotDataType::Raw;
+    SnapshotMergeOperation operation = SnapshotMergeOperation::Overwrite;
+
+    void addDiffs(std::vector<SnapshotDiff>& diffs,
+                  const uint8_t* original,
+                  const uint8_t* updated);
+};
```
Shillaker, line 76: Moved this class down the file as it now depends on the `SnapshotDiff` class which is declared above.

include/faabric/transport/PointToPointServer.h
```diff
@@ -11,7 +11,7 @@ class PointToPointServer final : public MessageEndpointServer
     PointToPointServer();
 
   private:
-    PointToPointBroker& reg;
+    PointToPointBroker& broker;
```
Shillaker, line 14: This used to be called a "registry" hence the `reg` name, but it was a bit confusing.

src/scheduler/Scheduler.cpp
```diff
@@ -224,228 +223,339 @@ faabric::util::SchedulingDecision Scheduler::callFunctions(
         throw std::runtime_error("Message with no master host");
     }
 
-    // Set up scheduling decision
```
Shillaker, line 227: The diff of the `callFunctions` methods isn't particularly easy to parse, full version [here](https://github.com/faasm/faabric/blob/dist-lock-fixes/src/scheduler/Scheduler.cpp).

tests/test/util/test_snapshot.cpp
```diff
@@ -500,84 +477,9 @@ TEST_CASE_METHOD(SnapshotMergeTestFixture,
     deallocatePages(snap.data, snapPages);
 }
 
-TEST_CASE_METHOD(SnapshotMergeTestFixture, "Test cross-page ignores", "[util]")
```
Shillaker, line 503: Ignores have been removed

tests/test/scheduler/test_executor.cpp
```diff
@@ -430,94 +448,6 @@ TEST_CASE_METHOD(TestExecutorFixture,
     }
 }
 
-TEST_CASE_METHOD(TestExecutorFixture,
-                 "Test executing remote chained threads",
-                 "[executor]")
```
Shillaker, line 435: This test is fiddly and fragile as it's trying to recreate a remote execution. Everything it checks is now covered in the distributed tests which are much cleaner, so I've removed it.

src/scheduler/Executor.cpp
```diff
@@ -142,31 +134,58 @@ void Executor::executeTasks(std::vector<int> msgIdxs,
...
-        if (isThreads) {
-            assert(threadPoolSize > 1);
-            threadPoolIdx = (msg.appidx() % (threadPoolSize - 1)) + 1;
+        int threadPoolIdx = -1;
+        if (availablePoolThreads.empty()) {
+            // Here all threads are still executing, so we have to overload.
+            // If any tasks are blocking we risk a deadlock, and can no longer
+            // guarantee the application will finish.
+            // In general if we're on the master host and this is a thread, we
+            // should avoid the zeroth and first pool threads as they are likely
```
csegarragonz, line 156: I have read about this in the PR description, and this seems to me a risky game to be playing.

How likely, i.e. when, would such an overload happen?

src/scheduler/Scheduler.cpp
```diff
@@ -224,228 +223,341 @@ faabric::util::SchedulingDecision Scheduler::callFunctions(
...
+        SPDLOG_DEBUG("Forwarding {} back to master {}", funcStr, masterHost);
 
         getFunctionCallClient(masterHost).executeFunctions(req);
         decision.returnHost = masterHost;
         return decision;
     }
 
+    faabric::util::FullLock lock(mx);
+
+    // -------------------------------------------
```
csegarragonz, line 239: Is there any strong argument against abstracting this to a `takeSchedulingDecision` helper method? Then basically scheduling would be a combination of `takeSchedulingDecision` and `doCallFunctions` which seems natural and in-line with that described in the text.

If it is much fuss (e.g. moving a lot of local state) I'm fine with leaving it like this.

src/transport/PointToPointBroker.cpp
```diff
@@ -74,15 +76,45 @@ std::shared_ptr<PointToPointGroup> PointToPointGroup::getGroup(int groupId)
...
+    if (groupExists(groupId)) {
+        return;
+    }
+
+    addGroup(appId, groupId, groupSize);
+}
+
+void PointToPointGroup::clearGroup(int groupId)
+{
+    groups.erase(groupId);
```
csegarragonz, line 117: Don't we need a 
```cpp
faabric::util::FullLock lock(groupsMutex);
``` 
for this?

csegarragonz (approved):
LGTM, just a couple minor points

src/scheduler/Executor.cpp
Shillaker: Overloads like this would only happen when there aren't enough resources available, so they're unlikely. When the system it overloaded it can either keep accepting functions and do its best to execute them, or start rejecting requests with an error. Faabric does the former and starts queueing, as this would then trigger the underlying cluster to scale out in a "real" deployment. 

This specific bit of code is only related to threading, i.e. when an application spawns more threads than we have cores in the system. Well-written multi-threaded applications ought to request the level of parallelism available in the environment, at which point the system can specify an appropriate limit that will avoid this scenario (as is the case with OpenMP).

src/scheduler/Scheduler.cpp
Shillaker: Yes good point, I think I tried this when writing it originally and like you say it was a bit fiddly, but will have another go.

src/transport/PointToPointBroker.cpp
Shillaker: Yes, probably a good idea.

PR #170, opened 2021-11-02
Concurrent collections experiment

Shillaker:


Shillaker:
Can resurrect if needed in future.

PR #174, opened 2021-11-11
Snapshot diff merging fixes

Shillaker:
These changes were part of the effort to get pthreads working in Faasm again. The original issue was an off-by-one error in the diff detection loop, but to make diagnosis simpler I refactored the diff detection logic. 

Changes:

- Move logic around working out dirty region offsets into new `getDirtyRegions` function. This is preferable to the old `getDirtyPages` which would return a list of page numbers that the diffing function eventually converted into regions. This logic is now isolated and tested separately.
- Refactored diff detection loop to push all remaining logic down into the `MergeRegion` class (now all the grimness is held in once place 😄)
- Allow `Overwrite` merge regions to specify `length=0`, which means they will extend to the end of the memory. This is necessary to merge back changes from memory allocated outside the original snapshot size.

### DIFF
--- a/src/util/memory.cpp
+++ b/src/util/memory.cpp
@@ -151,4 +151,29 @@ std::vector<int> getDirtyPageNumbers(const uint8_t* ptr, int nPages)
 
     return pageNumbers;
 }
+
+std::vector<std::pair<uint32_t, uint32_t>> getDirtyRegions(const uint8_t* ptr,
+                                                           int nPages)
+{
+    std::vector<int> dirtyPages = getDirtyPageNumbers(ptr, nPages);
+
+    // Add a new region for each page, unless the one before it was also dirty,
+    // in which case we merge them
+    std::vector<std::pair<uint32_t, uint32_t>> regions;
+    for (int p = 0; p < dirtyPages.size(); p++) {
+        int thisPageNum = dirtyPages.at(p);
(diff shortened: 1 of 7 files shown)


src/util/memory.cpp
```diff
@@ -151,4 +151,29 @@ std::vector<int> getDirtyPageNumbers(const uint8_t* ptr, int nPages)
...
+    // in which case we merge them
+    std::vector<std::pair<uint32_t, uint32_t>> regions;
+    for (int p = 0; p < dirtyPages.size(); p++) {
+        int thisPageNum = dirtyPages.at(p);
+        uint32_t thisPageStart = thisPageNum * HOST_PAGE_SIZE;
+        uint32_t thisPageEnd = thisPageStart + HOST_PAGE_SIZE;
+
+        if (p > 0 && dirtyPages.at(p - 1) == thisPageNum - 1) {
+            // Previous page was also dirty, just update last region
+            regions.back().second = thisPageEnd;
```
csegarragonz, line 170: Could `regions` here be empty, thus `back()` cause undefined behaviour?

I am aware this may not have sense logically, but maybe worth catching?

csegarragonz (approved):
LGTM, I am not even sure what I suggest is worth doing

src/util/memory.cpp
Shillaker: Yes good point, I'll add an assert as that empty `back()` problem can be hard to diagnose.

PR #175, opened 2021-11-12
Track Message Type in Execution Graph

csegarragonz:
I am half-hearted on wether we should merge this in or not. On the one hand it is useful to have, and doesn't do any harm. On the other hand, it's just a profiling tool we have used at a point in time.

I am slightly leaning towards merging it in, but also happy to keep this out.

### DIFF
--- a/include/faabric/util/exec_graph.h
+++ b/include/faabric/util/exec_graph.h
@@ -16,4 +16,6 @@ void incrementCounter(faabric::Message& msg,
                       const int valueToIncrement = 1);
 
 static inline std::string const mpiMsgCountPrefix = "mpi-msgcount-torank-";
+
+static inline std::string const mpiMsgTypeCountPrefix = "mpi-msgtype-torank";
 }
--- a/include/faabric/util/exec_graph.h
+++ b/include/faabric/util/exec_graph.h
@@ -7,6 +7,7 @@
 #include <map>
 
 namespace faabric::util::exec_graph {
+
 void addDetail(faabric::Message& msg,
                const std::string& key,
                const std::string& value);
@@ -14,8 +15,4 @@ void addDetail(faabric::Message& msg,
 void incrementCounter(faabric::Message& msg,
                       const std::string& key,
                       const int valueToIncrement = 1);
-
-static inline std::string const mpiMsgCountPrefix = "mpi-msgcount-torank-";
-
-static inline std::string const mpiMsgTypeCountPrefix = "mpi-msgtype-torank";
 }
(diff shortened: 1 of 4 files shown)


tests/test/scheduler/test_mpi_exec_graph.cpp
```diff
@@ -36,8 +36,6 @@ TEST_CASE_METHOD(MpiTestFixture,
           rankA1, rankA2, BYTES(buffer), MPI_INT, messageData.size(), &status);
     }
 
-    REQUIRE(msg.intexecgraphdetails_size() == 1);
```
csegarragonz, line 39: I would have had to increase this number every time we tracked a different thing. It did not seem a useful thing to test/easy to forget to change.

include/faabric/util/exec_graph.h
```diff
@@ -16,4 +16,6 @@ void incrementCounter(faabric::Message& msg,
                       const int valueToIncrement = 1);
 
 static inline std::string const mpiMsgCountPrefix = "mpi-msgcount-torank-";
+
+static inline std::string const mpiMsgTypeCountPrefix = "mpi-msgtype-torank";
```
Shillaker, line 20: I can't remember what the conclusion of our last discussion on these two points was, but to reiterate (feel free to ignore) I think these constants should (a) be `#define`s to fit with the style of the repo; (b) live in the MPI headers if possible because (i) they're MPI specific; (ii) this list will continue to grow and potentiall become quite big otherwise.

Either way, re. your point below about tracking the number of fields, I think it's still useful, but to more obviously connect them up, I'd add a constant wherever these labels are defined, e.g. `#define N_MPI_EXEC_GRAPH_DETAILS 2`, then use that in the test.

tests/test/scheduler/test_mpi_exec_graph.cpp
Shillaker: See point about constant above, I think this is worth testing.

src/scheduler/MpiWorld.cpp
```diff
@@ -589,15 +589,13 @@ void MpiWorld::send(int sendRank,
...
-          faabric::util::exec_graph::mpiMsgCountPrefix +
-            std::to_string(recvRank));
+          *thisRankMsg, MPI_MSG_COUNT_PREFIX + std::to_string(recvRank));
 
         // Work out the message type breakdown
         faabric::util::exec_graph::incrementCounter(
           *thisRankMsg,
           fmt::format("{}-{}-{}",
-                      faabric::util::exec_graph::mpiMsgTypeCountPrefix,
+                      MPI_MSGTYPE_COUNT_PREFIX,
```
Shillaker, line 598: Does this string format then result in a double hyphen at the beginning (as the prefix already has a hyphen)?

Shillaker (approved):
Nice, one small question, feel free to disregard.

src/scheduler/MpiWorld.cpp
csegarragonz: For the `MSGTYPE_COUNT` we have to provide two values, so the constant actually hasn't got a trailing hyphen. I will change it to make them both consistent.

PR #176, opened 2021-11-18
Snapshots overhaul

Shillaker:
Apologies, this PR has turned into a beast, but I think it's worth spending time on, as it's one of the more fiddly parts of the system.

The logic for creating, updating, diffing and merging snapshots is currently spread over several classes, quite difficult to understand because of the unclear ownership of data, and sprinkled with data races. In this PR I refactor all the snapshot-related code into something (hopefully) easier to understand, and safer to use.

The `snapshot.h` and `snapshot.cpp` files have gotten way too big and should be split up by class, however, that would make the diff for this PR even larger than it already is, so can be done in a separate PR.

Functionality changes:

- Merge regions are cleared after pushing snapshot diffs, as these regions only apply to individual batches of threads.
- `SnapshotData` now has a max size, and will allocate a contiguous portion of virtual memory into which it may expand in future (in the case of wasm this will be of size 4GB). If data is being copied into the snapshot that is larger than the current size, it will allocate new portions of memory.
- `SnapshotData` and `SnapshotDiff` take ownership of their own data, and a new `MemoryView` class is used to represent views of memory (which was previously rolled into these two classes).
- Threads on the main host now diff memory with their original snapshot and write back changes. Without this, changes from these threads are not visible in the snapshot as the mapping of memory is private, hence one-way. 

Refactoring:

- The logic for merging diffs into a snapshot was held in the `SnapshotServer` class, but is now in the `SnapshotData` class.
- `SnapshotData` uses proper RAII to manage its contents. 
- `SnapshotDiff`s are now queued and applied only at explicit synchronisation points. Without this, changes to the snapshot on the main host may be interleaved and mess up the diffing process on that host. This has an additional benefit of making the process simpler to reason about.
- Everything deals with `std::shared_ptr<SnapshotData>` rather than references. This reduces risk of confusion over copying and shared ownership of data (and we can delete the copy and assignment constructors).
- All logic around memory provisioning and file descriptors is now in the `SnapshotData` class itself rather than spread between it and the `SnapshotRegistry`.
- Low-level handling of memory allocation and mapping is now contained in utility functions to avoid having `mmap` and `mprotect` calls inline.
- Snapshot `getDirtyPages` is now `getDirtyRegions` to support a minimal set of diffs (rather than one per page even if the pages are contiguous).


src/util/memory.cpp
```diff
@@ -177,4 +177,150 @@ std::vector<std::pair<uint32_t, uint32_t>> getDirtyRegions(const uint8_t* ptr,
...
+// -------------------------
+
+uint8_t* allocateSharedMemory(size_t size)
+{
+    if (size < HOST_PAGE_SIZE) {
+        SPDLOG_WARN("Allocating less than a page of memory");
+    }
+
+    void* mmapRes =
+      ::mmap(nullptr, size, PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
```
kubasz, line 192: The protection flags should most likely be `PROT_READ|PROT_WRITE` here. While x86 doesn't have write-only mappings, so r+w is implicitly used, this is not what this function is requesting. Same for other mmap calls below

tests/test/snapshot/test_snapshot_client_server.cpp
```diff
@@ -139,41 +140,71 @@ TEST_CASE_METHOD(SnapshotClientServerFixture,
...
-    // Set up a snapshot
+    // Set up a snapshot that's got enough memory to expand into
     std::string snapKey = std::to_string(generateGid());
-    SnapshotData snap = takeSnapshot(snapKey, 5, true);
+    size_t initialSnapSize = 5 * HOST_PAGE_SIZE;
+    size_t expandedSnapSize = 10 * HOST_PAGE_SIZE;
+
+    faabric::util::SnapshotData snap;
+    snap.size = initialSnapSize;
+    snap.data = faabric::util::allocateVirtualMemory(expandedSnapSize);
```
kubasz, line 150: Could you introduce a std::unique_ptr-like (or even just std::unique_ptr with a custom deleter) wrapper for these allocations? I've been purging all uses of new/delete and using RAII memory management where possible, this feels like a step backwards. I know this could be impossible for the general case of applying snapshots onto faaslets, but at least in testing code the lifetime of these mappings is limited to a function call scope.

src/snapshot/SnapshotServer.cpp
```diff
@@ -84,18 +83,25 @@ std::unique_ptr<google::protobuf::Message> SnapshotServer::recvPushSnapshot(
...
-    // this data?
-    data.data = (uint8_t*)mmap(
-      nullptr, data.size, PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
-    std::memcpy(data.data, r->contents()->Data(), data.size);
+    // TODO - somehow avoid this copy?
+    std::memcpy(snap.data, r->contents()->Data(), snap.size);
 
-    reg.takeSnapshot(r->key()->str(), data, true);
+    // TODO - avoid a further copy in here when setting up fd
+    reg.takeSnapshot(r->key()->str(), snap, true);
```
kubasz, line 104: At least one copy could be avoided here by adding an API to create a snapshot (including its memfd) of a given size without data, then making the caller responsible for filling in the data. The call would have to create the memfd, ftruncate it and mmap it, saving the mmap location in the `data` field

Shillaker: Yes that might work, I'll see what I can do.

src/util/snapshot.cpp
```diff
@@ -10,27 +10,51 @@ namespace faabric::util {
 // class, but it can't be copy-constructed.
 static std::mutex snapMx;
```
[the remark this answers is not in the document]
Shillaker: This has gone now with the refactoring.

tests/test/snapshot/test_snapshot_client_server.cpp
Shillaker: Two observations after a bit of fiddling:

- To interface properly with Faasm, the snapshot API still needs to support raw pointers (as both WAMR and WAVM still use them for wasm memory), hence we can't go full RAII on a lot of the changes introduced in this PR. We could introduce a parallel API that supports smart pointers alongside raw pointers, but this will result in a fair bit of extra code.

- I agree that using RAII in tests for memory provisioned on the fly is a good idea, and I can almost get something with a `unique_ptr` working, however, the custom deleter needs to call `munmap` and needs to know the size of the memory that was orinally `mmap`ed, which makes defining the delete a little more tricky. I got this far:

```
std::unique_ptr<uint8_t[]> doAlloc(size_t size, int prot, int flags)
{
    auto deleter = [size](uint8_t* u) { munmap(u, size); };
    std::unique_ptr<uint8_t[], decltype(deleter)> mem(
        (uint8_t*)::mmap(nullptr, size, prot, flags, -1, 0), deleter);
    return mem;
}
```

However, this won't compile as the return type of the function also needs to specify the type of the deleter, but I can't work out the syntax. Any ideas?

kubasz: You can use `std::function<void(uint8_t*)>` as the deleter type to store that lambda

Shillaker: The alternative is to use a `shared_ptr`, which I think should be more flexible re. the custom deleter.

kubasz: Here's a minimal working example with unique_ptr:
```
#include <memory>
#include <functional>
#include <sys/mman.h>

using OwnedMmapRegion = std::unique_ptr<uint8_t[], std::function<void(uint8_t*)>>;

OwnedMmapRegion doAlloc(size_t size, int prot, int flags)
{
    auto deleter = [size](uint8_t* u) { munmap(u, size); };
    OwnedMmapRegion mem(
        (uint8_t*)::mmap(nullptr, size, prot, flags, -1, 0), deleter);
    return mem;
}
```

<https://godbolt.org/z/TdTzaz9rW>

I'm not sure what you mean by shared_ptr being more flexible with the custom deleter, the only difference would be that you could use reference counting to keep track when the last user is gone.

Shillaker: Awesome, cheers @kubasz. What I meant re. `shared_ptr` being more flexible is that I don't think the deleter type needs to be included in the pointer type, unlike a `unique_ptr`, so the return type of the function could still be `std::shared_ptr<uint8_t[]>` even with the deleter (and I could dodge the original problem I was facing). Anywho, your version is great so I'll see how things turn out with that.

include/faabric/util/snapshot.h
```diff
@@ -74,12 +76,45 @@ class SnapshotData
...
+
+    ~SnapshotData();
+
+    bool isRestorable();
+
+    void makeRestorable(const std::string& fdLabel);
+
+    void setSnapshotSize(size_t newSize);
+
+    void copyInData(std::vector<uint8_t> buffer, uint32_t offset = 0);
```
kubasz, line 105: Passing `std::vector` by-value is almost always wrong and results in extra copies, `const std::vector<uint8_t>&` makes much more sense. But also this is the perfect spot to use `std::span<uint8_t>` as it's a pointer+size pair, can be constructed easily from pointers with sizes or std containers, including sub-buffers like: `span(vector).subspan(offset)` - this doesn't do any copies, just some pointer arithmetic

Shillaker: Ah yes good spot and thanks for the tip re. `span`. This PR is still very much a WIP and I'm in the middle of hacking through the refactor, so will be going through and cleaning up stuff like this at the end. I'll request a review once I'm happy with the finished version.

.github/workflows/sanitisers.yml
```diff
@@ -99,7 +99,7 @@ jobs:
             ${{ runner.os }}-
       # --- Build with thread sanitising
       - name: "Build tests with address sanitising"
-        run: inv dev.sanitise Address --noclean
+        run: inv dev.sanitise Address faabric_tests --noclean
```
Shillaker, line 102: This was so that I could sanitise and run other targets such as the distributed tests.

src/scheduler/Scheduler.cpp
```diff
@@ -83,8 +83,7 @@ void Scheduler::addHostToGlobalSet()
 
 void Scheduler::resetThreadLocalCache()
 {
-    auto tid = (pid_t)syscall(SYS_gettid);
-    SPDLOG_DEBUG("Resetting scheduler thread-local cache for thread {}", tid);
+    SPDLOG_DEBUG("Resetting scheduler thread-local cache");
```
Shillaker, line 86: Thread ID is included in all log statements by default so this was unnecessary.

src/snapshot/SnapshotRegistry.cpp
```diff
@@ -33,75 +34,38 @@ bool SnapshotRegistry::snapshotExists(const std::string& key)
...
-    }
-
-    void* mmapRes =
-      mmap(target, d->size, PROT_WRITE, MAP_PRIVATE | MAP_FIXED, d->fd, 0);
-
-    if (mmapRes == MAP_FAILED) {
-        SPDLOG_ERROR(
-          "mmapping snapshot failed: {} ({})", errno, ::strerror(errno));
-        throw std::runtime_error("mmapping snapshot failed");
-    }
```
Shillaker, line 56: This logic is now in the `SnapshotData` class.

src/snapshot/SnapshotRegistry.cpp
```diff
@@ -112,16 +76,6 @@ void SnapshotRegistry::deleteSnapshot(const std::string& key)
...
 
-    auto d = snapshotMap[key];
-
-    // Note - the data referenced by the SnapshotData object is not owned by the
-    // snapshot registry so we don't delete it here. We only remove the file
-    // descriptor used for mapping memory
-    if (d->fd > 0) {
-        ::close(d->fd);
-        d->fd = 0;
-    }
```
Shillaker, line 123: Closing the file descriptor is also handled in `SnapshotData` now.

src/snapshot/SnapshotRegistry.cpp
```diff
@@ -140,38 +94,6 @@ SnapshotRegistry& getSnapshotRegistry()
...
-    if (werror == -1) {
-        SPDLOG_ERROR("Write call failed with error {}", werror);
-        throw std::runtime_error("Failed writing memory to fd (write)");
-    }
-
-    // Record the fd
-    data.fd = fd;
-
-    SPDLOG_DEBUG("Wrote snapshot {} to fd {}", key, fd);
-    return fd;
```
Shillaker, line 175: Moved to `SnapshotData`.

kubasz (changes requested):

src/snapshot/SnapshotRegistry.cpp
```diff
@@ -33,75 +34,38 @@ bool SnapshotRegistry::snapshotExists(const std::string& key)
...
-    auto shared_data =
-      std::make_shared<faabric::util::SnapshotData>(std::move(data));
-    snapshotMap[key] = shared_data;
+    SPDLOG_TRACE("Registering snapshot {} size {}", key, data->size);
 
-    // Write to fd to be locally restorable
-    if (locallyRestorable) {
-        writeSnapshotToFd(key, *shared_data);
-    }
+    snapshotMap[key] = data;
```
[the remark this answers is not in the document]
Shillaker: I'm not sure about adding an implicit `move` in here. The caller may wish to reuse the object they're passing in (especially in tests).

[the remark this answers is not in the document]
kubasz: The move will be from the argument, which is already copied by value when passed to the function. It's a move of the shared_ptr, not of the underlying value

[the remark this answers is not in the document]
Shillaker: Ah yes, sorry I couldn't see the function signature in the little diff view.

src/util/memory.cpp
```diff
@@ -177,4 +177,121 @@ std::vector<std::pair<uint32_t, uint32_t>> getDirtyRegions(const uint8_t* ptr,
...
 }
+
+// -------------------------
+// Allocation
+// -------------------------
+
+OwnedMmapRegion doAlloc(size_t size, int prot, int flags)
+{
+    auto deleter = [size](uint8_t* u) { munmap(u, size); };
+    OwnedMmapRegion mem((uint8_t*)::mmap(nullptr, size, prot, flags, -1, 0),
```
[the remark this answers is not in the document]
Shillaker: Yes sorry, lazy copy-paste on my part.

src/util/memory.cpp
```diff
@@ -177,4 +177,121 @@ std::vector<std::pair<uint32_t, uint32_t>> getDirtyRegions(const uint8_t* ptr,
...
+}
+
+void mapMemory(uint8_t* target, size_t size, int fd)
+{
+    if (!faabric::util::isPageAligned((void*)target)) {
+        SPDLOG_ERROR("Mapping memory to non page-aligned address");
+        throw std::runtime_error("Mapping memory to non page-aligned address");
+    }
+
+    if (fd == 0) {
```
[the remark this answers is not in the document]
Shillaker: Yep you're right, but even though it's technically a valid fd in some scenarios, it's unexpected and not supported in this case. It's frequently a result of the fd being uninitialised (e.g. due to an error in the `SnapshotData` class), hence I think it's a valid check, and the error message is accurate; it's not saying that it's not valid, just that it's zero.

src/util/memory.cpp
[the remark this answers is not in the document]
Shillaker: I think all the other uses of `mmap` and friends are checked, it's just this one.

src/util/snapshot.cpp
```diff
@@ -4,33 +4,177 @@
...
+    // Allocate virtual memory big enough for the max size if provided
+    ownedData = faabric::util::allocateVirtualMemory(maxSize);
+
+    // Claim just the snapshot region
+    faabric::util::claimVirtualMemory(BYTES(ownedData.get()), size);
+}
+
+SnapshotData::SnapshotData(std::vector<uint8_t>& dataIn)
+  : SnapshotData(dataIn.data(), dataIn.size())
+{}
```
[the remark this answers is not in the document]
Shillaker: In general the rules for this class are: if you're giving it some data, you're the owner of that data. If you're just telling it a size, you're letting the `SnapshotData` take ownership. I'll add some comments elsewhere as you mention below.

src/util/snapshot.cpp
```diff
@@ -4,33 +4,177 @@
...
+    }
+
+    // Copy in new data
+    ::memcpy(data + offset, buffer, bufferSize);
+
+    // Update fd
+    if (fd > 0 && sizeChanged) {
+        faabric::util::appendDataToFd(fd, fdSize, size, data);
+        fdSize = size;
+    }
```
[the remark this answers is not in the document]
Shillaker: Hmm yes good point, my gut is that this is actually the only place it has to be updated to function correctly with the current Faasm use-case, but it should be more generalised than that. I'll take a look.

src/snapshot/SnapshotServer.cpp
```diff
@@ -163,28 +162,31 @@ SnapshotServer::recvPushSnapshotDiffs(const uint8_t* buffer, size_t bufferSize)
                 break;
             }
             case (faabric::util::SnapshotDataType::Int): {
-                const auto* value =
+                const auto* diffValue =
```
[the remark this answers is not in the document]
Shillaker: Yes another good point, although the functions in `util/bytes.h` use `std::byte`, and without changing their signature to `uint8_t` in place of `std::byte` it looks like I can't use them directly on a `uint8_t` pointer. Given that everything else uses `uint8_t` I'd be inclined to change them to be in sync with the rest of the codebase, but would this break things for you?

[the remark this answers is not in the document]
kubasz: It's fine to change it, `std::byte` is fairly recent and meant to represent chunks of arbitrarily-typed memory as opposed to specifically arrays of 8-bit integers, they can be easily converted between each other (but not implicitly). I think using it to mean "block of memory" is useful, but it can be annoying to keep converting it back to uint8_t which most APIs expect - up to you which route you want to go.

src/util/memory.cpp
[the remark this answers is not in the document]
kubasz: You're right, when quickly skimming the code I thought the `doAlloc` calls were `mmap` calls because of argument layout

src/util/memory.cpp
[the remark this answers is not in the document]
kubasz: Yes, what I had in mind is that the default initializer for the member could be changed to -1 (and checks could be changed to `<0` or `<=0` if you want to make sure if a zero is still somewhere it gets caught)

Shillaker:
Thanks for the comments @kubasz , I'll finish them off tomorrow then re-request a review as what I've just pushed is only half there.

src/snapshot/SnapshotServer.cpp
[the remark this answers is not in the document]
Shillaker: Ok cool, I think `std:byte` is the way to go, definitely easier to read, but perhaps a wholesale change in a single PR might be better than doing it bit by bit. I'll change them to `uint8_t` in this PR and add an issue.

include/faabric/util/bytes.h
```diff
@@ -26,18 +26,18 @@ int safeCopyToBuffer(const uint8_t* dataIn,
                      int bufferLen);
 
 template<class T>
-T unalignedRead(const std::byte* bytes)
+T unalignedRead(const uint8_t *bytes)
```
Shillaker, line 29: As discussed in PR, have switched this back to `uint8_t` for now to fit with rest of the API, but definitely worth moving to `std::byte` wholesale (see https://github.com/faasm/faabric/issues/199).

src/snapshot/SnapshotServer.cpp
```diff
@@ -136,75 +136,22 @@ SnapshotServer::recvPushSnapshotDiffs(const uint8_t* buffer, size_t bufferSize)
           ->localLock();
     }
 
-    // Iterate through the chunks passed in the request
+    // Convert chunks to snapshot diff objects
+    std::vector<SnapshotDiff> diffs;
     for (const auto* chunk : *r->chunks()) {
-        uint8_t* dest = snap->data + chunk->offset();
```
Shillaker, line 141: The following logic around merging snapshot diffs has also been moved to `SnapshotData`.

src/util/snapshot.cpp
```diff
@@ -1,81 +1,126 @@
...
-                           HOST_PAGE_SIZE);
-    }
+    // Claim just the snapshot region
+    faabric::util::claimVirtualMemory({ BYTES(_data.get()), size });
 
-    SPDLOG_DEBUG("Snapshot has {}/{} dirty pages", diffs.size(), nPages);
+    // Set up the fd with a two-way mapping to the data
+    std::string fdLabel = "snap_" + std::to_string(generateGid());
+    fd = createFd(size, fdLabel);
+    mapMemoryShared({ _data.get(), size }, fd);
```
Shillaker, line 34: This will solve the issues around `fd` getting out of sync with snapshot data.

bin/workon.sh
```diff
@@ -75,6 +75,14 @@ export PS1="(faabric) $PS1"
 
 export PATH=/build/faabric/static/bin:${PATH}
 
+# ----------------------------
+# Sanitisers
+# ----------------------------
+# Check these match the CI build
+export ASAN_OPTIONS = "verbosity=1:halt_on_error=1"
+export TSAN_OPTIONS="halt_on_error=1:suppressions=${PROJ_ROOT}/thread-sanitizer-ignorelist.txt:history_size=7:second_deadlock_stack=1"
+export UBSAN_OPTIONS="print_stacktrace=1:halt_on_error=1"
```
Shillaker, line 84: Important for replicating CI problems locally (without `halt_on_error=1` it's hard to tell if anything's breaking).

tests/dist/DistTestExecutor.h
```diff
@@ -44,4 +41,15 @@ class DistTestExecutorFactory : public faabric::scheduler::ExecutorFactory
     std::shared_ptr<faabric::scheduler::Executor> createExecutor(
       faabric::Message& msg) override;
 };
+
+typedef int (*ExecutorFunction)(
```
kubasz, line 45: While a function pointer is more efficient, a `std::function<int(tests::DistTestExecutor*,...)>` could also accept lambdas with captures

tests/test/scheduler/test_scheduler.cpp
```diff
@@ -418,16 +417,13 @@ TEST_CASE_METHOD(SlowExecutorFixture,
     SECTION("Functions") { execMode = faabric::BatchExecuteRequest::FUNCTIONS; }
 
     // Set up snapshot if necessary
-    faabric::util::SnapshotData snapshot;
     faabric::snapshot::SnapshotRegistry& snapRegistry =
       faabric::snapshot::getSnapshotRegistry();
 
-    std::unique_ptr<uint8_t[]> snapshotDataAllocation;
+    size_t snapSize = 1234;
```
kubasz, line 423: I know it was like this before, but shouldn't snapshot sizes be multiples of the OS page size (4096)?

include/faabric/scheduler/Scheduler.h
```diff
@@ -69,7 +69,7 @@ class Executor
 
     void releaseClaim();
 
-    virtual faabric::util::SnapshotData snapshot();
+    virtual std::shared_ptr<faabric::util::MemoryView> getMemoryView();
```
kubasz, line 72: See my other comment re: MemoryView

include/faabric/util/snapshot.h
```diff
@@ -73,27 +76,95 @@ class SnapshotMergeRegion
...
+class MemoryView
+{
+  public:
+    // Note - this object is just a view of a section of memory, and does not
+    // own the underlying data
+    explicit MemoryView(std::span<const uint8_t> dataIn);
+
+    MemoryView(const MemoryView&) = delete;
+
+    MemoryView& operator=(const MemoryView&) = delete;
```
kubasz, line 157: If the MemoryView is just a view object, not a resource-owner, it should be trivially copyable and moveable (with default rather than deleted constructors) - and not wrapped in `std::shared_ptr`s where used, as it's just a pointer+length pair (span), especially because you made it immutable so there's no risk of something modifying a MemoryView object.

include/faabric/util/snapshot.h
```diff
@@ -73,27 +76,95 @@ class SnapshotMergeRegion
...
                         SnapshotMergeOperation operation,
                         bool overwrite = false);
 
+    void clearMergeRegions();
+
+    std::map<uint32_t, SnapshotMergeRegion> getMergeRegions();
+
+    size_t getQueuedDiffsCount();
+
+    void queueDiffs(const std::vector<SnapshotDiff>& diffs);
```
kubasz, line 117: Accept any std::span instead of just std::vector (it's implicitly constructible, so this change allows more ways to call the function - like with a local stack-allocated array - without breaking the vector-based one)
```suggestion
    void queueDiffs(std::span<const SnapshotDiff> diffs);
```

kubasz (changes requested):
The PR looks very good now, just a few minor comments/questions left from me

tests/test/scheduler/test_scheduler.cpp
Shillaker: I don't think so. AFAICT there's nothing in the implementation that restricts them to being page-aligned, although the use of `mmap` under the hood will cause the underlying memory and mappings to be rounded up to the nearest OS page. 

When used in Faasm they will all be multiples of the wasm page size (64k), so in practice they will all be OS-page aligned, but I don't think we necessarily need to bake that into the API at the moment.

include/faabric/util/snapshot.h
Shillaker: Nice, we have a lot of functions that take refs to vectors that can be replaced with this; will keep my eyes peeled 👀.

tests/dist/DistTestExecutor.h
Shillaker: Yes good point, that would definitely be clearer. This was originally a bit hacked together but has proven quite useful and expanded. Rather than change it now and further bloat this PR I'd like to give the distributed tests a bit more love in general in another PR.

include/faabric/util/snapshot.h
```diff
@@ -40,13 +40,13 @@ class SnapshotDiff
...
+    SnapshotDataType getDataType() const { return dataType; }
 
-    SnapshotMergeOperation getOperation() const;
+    SnapshotMergeOperation getOperation() const { return operation; }
 
-    uint32_t getOffset() const;
+    uint32_t getOffset() const { return offset; }
 
-    std::span<const uint8_t> getData() const;
+    std::span<const uint8_t> getData() const { return data; }
```
kubasz, line 49: These (and the ones below, lines 121, 123, 162) have to be marked `inline` to avoid linker errors if this file is included from multiple cpp files in the same library - everything else looks good to me

Shillaker: Interesting, I've never encountered this problem before; I don't think we've declared any other header-defined functions `inline` in the codebase. In addition, this file is included in both `snapshot/SnapshotServer.cpp` and `snapshot/SnapshotRegistry.cpp` (i.e. in the same CMake library, not sure if that's what you mean by library here). Can you be more specific about where this would this cause a problem?

kubasz: Oh nevermind, apparently any definitions inside the `class {}` body are implicitly inline, I didn't know this

kubasz (approved):

include/faabric/util/snapshot.h
Shillaker: Ah interesting, good to know!

PR #178, opened 2021-11-22
Get `ExecGraph` hosts

csegarragonz:
Add utility function to get the distinct hosts in an execution graph.

Also bump the code version to re-build the containers as required in #177


tests/test/scheduler/test_exec_graph.cpp
```diff
@@ -72,6 +72,25 @@ TEST_CASE("Test execution graph", "[scheduler][exec-graph]")
     checkExecGraphEquality(expected, actual);
 }
 
+TEST_CASE("Test execution graph node count", "[scheduler][exec-graph]")
```
Shillaker, line 75: This is very nitpicky but the API is about getting the unique host names from the graph, not the count. 

I'd have thought this test would be called `Test get unique hosts from exec graph`, then the assertion at the end would look like:

```cpp
std::set<std::string> expected = {"foo", "bar", "baz"};
REQUIRE(hosts == expected);
```

Currently the test also doesn't cover duplicates, it's got 3 messages each executed on a different host. It would be good to add another message for a different function but executed on the same host as one of the others to be extra sure.

src/scheduler/ExecGraph.cpp
```diff
@@ -25,6 +25,28 @@ int countExecGraphNodes(const ExecGraph& graph)
     return count;
 }
 
+std::set<std::string> getExecGraphHostsForNode(const ExecGraphNode& node)
+{
+    std::set<std::string> hostsForNode;
+    hostsForNode.insert(node.msg.executedhost());
+
+    if (!node.children.empty()) {
```
Shillaker, line 33: I think this `if` is unnecessary (as `for(auto c : node.children)` will do nothing if it's empty).

src/scheduler/ExecGraph.cpp
```diff
@@ -25,6 +25,28 @@ int countExecGraphNodes(const ExecGraph& graph)
...
+    }
+
+    return hostsForNode;
+}
+
+std::set<std::string> getExecGraphHosts(const ExecGraph& graph)
+{
+    ExecGraphNode rootNode = graph.rootNode;
+    std::set<std::string> hosts = getExecGraphHostsForNode(rootNode);
+    return hosts;
```
Shillaker, line 47: Very very nitpicky but if you're changing things I'd say this would be just as clear with:

```cpp
return getExecGraphHostsForNode(graph.rootNode)
```

Shillaker (changes requested):

tests/test/scheduler/test_exec_graph.cpp
csegarragonz: All good points, thanks.

Shillaker (approved):

PR #180, opened 2021-11-24
Add Scheduling Topology Hints

csegarragonz:
In this PR I add support to pass scheduling topology hints as a parameter to `callFunctions`. Scheduling hints affect the way batches are scheduled to the available resources. The currently supported hints are:
* `NORMAL`: will bin-pack requests to hosts until it runs out of slots, and then will overload the master (default).
* `NEVER_ALONE`: will never schedule a request to a host without other requests of the batch. There is one exception: master requests (first in the batch) will be allocated to a new host (even on their own) if the master is out of resources.

These hints only affect the `SchedulingDecision` so the separation recently made between making the scheduling decision and actually calling the functions has proven extremely useful for the implementation and testing. I add tooling for testing these and future hints we may add easily, together with a number of tests.


tests/test/util/test_scheduling.cpp
```diff
@@ -93,4 +95,283 @@ TEST_CASE("Test converting point-to-point mappings to scheduling decisions",
...
+        config.expectedHosts = { masterHost, masterHost, "hostA", "hostA",
+                                 "hostB",    "hostB",    "hostC", "hostC" };
+    }
+
+    SECTION("Uneven slot ditribution across hosts")
+    {
+        config.slots = { 3, 2, 2, 1 };
+        config.expectedHosts = { masterHost, masterHost, masterHost, "hostA",
+                                 "hostA",    "hostB",    "hostB",    "hostC" };
+    }
```
Shillaker, line 289: The way these tests are parameterised is great, really nice and clear. We could probably make more use of use-case-specific fixtures like this to make the tests more readable and modifiable.

src/scheduler/Scheduler.cpp
```diff
@@ -296,8 +311,20 @@ faabric::util::SchedulingDecision Scheduler::makeSchedulingDecision(
                 int available = r.slots() - r.usedslots();
                 int nOnThisHost = std::min(available, remainder);
 
-                for (int i = 0; i < nOnThisHost; i++) {
-                    hosts.push_back(h);
+                // Under the pairs topology hint, we never allocate a single
```
Shillaker, line 314: My gut here would instead be:

```cpp
if(topologyHint == faabric::util::SchedulingTopologyHint::PAIRS &&
      nOnThisHost < 2) {
    // Move on if we can't colocate function with at least one other
    continue;
}
```

src/scheduler/Scheduler.cpp
```diff
@@ -323,13 +350,22 @@ faabric::util::SchedulingDecision Scheduler::makeSchedulingDecision(
...
                 int nOnThisHost = std::min(available, remainder);
 
-                // Register the host if it's exected a function
-                if (nOnThisHost > 0) {
-                    registeredHosts[funcStr].insert(h);
-                }
+                bool stickToPreviousHost =
+                  (topologyHint ==
+                     faabric::util::SchedulingTopologyHint::PAIRS &&
+                   nOnThisHost == 1 && hosts.size() > 0);
```
Shillaker, line 356: Same point above re. logic.

Shillaker (changes requested):

tests/test/util/test_scheduling.cpp
```diff
@@ -3,14 +3,16 @@
...
+#include <faabric/util/config.h>
 #include <faabric/util/func.h>
 #include <faabric/util/scheduling.h>
 
 using namespace faabric::util;
 
 namespace tests {
 
-TEST_CASE("Test building scheduling decisions", "[util]")
+TEST_CASE("Test building scheduling decisions", "[util][scheduling-decisions]")
```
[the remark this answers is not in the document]
csegarragonz: I agree thanks, for future reference, you can run all tests in `tests/test/scheduler/test_scheduling_decisions.cpp` by running:
```bash
faabric_tests -# [#test_scheduling_decisions]
```

src/scheduler/Scheduler.cpp
csegarragonz: I don't think this behaves as we expect it. For example:
* We have 4 hosts with 4 slots each.
* We want to schedule 9 requests (with the `NEVER_ALONE` hint).
* We expect 4 requests scheduled to the first host, and 5 scheduled to the latter.
However, I think your solution will schedule 5 requests on the first host and 4 on the second, as it would exhaust all possible hosts (`nOnThisHost == 1` for all of them), and resort to overload the master.

include/faabric/scheduler/Scheduler.h
```diff
@@ -177,6 +179,11 @@ class Scheduler
 
     void clearRecordedMessages();
 
+    faabric::util::SchedulingDecision publicMakeSchedulingDecision(
+      std::shared_ptr<faabric::BatchExecuteRequest> req,
+      bool forceLocal,
+      faabric::util::SchedulingTopologyHint topologyHint);
```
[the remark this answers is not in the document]
csegarragonz: Yes, switching to `callFunctions` is pretty easy with the only caveat that we need to use the `TestExecutor` and `TestExecutorFactory` classes that currently lived in `./tests/tests/scheduler/test_executor.cpp`.

I have moved the declaration of these classes to `fixtures.h` and kept the definition where it is.

tests/test/util/test_scheduling.cpp
```diff
@@ -93,4 +95,283 @@ TEST_CASE("Test converting point-to-point mappings to scheduling decisions",
...
+    // Helper struct to configure one scheduling decision
+    struct SchedulingConfig
+    {
+        std::vector<std::string> hosts;
+        std::vector<int> slots;
+        int numReqs;
+        bool forceLocal;
+        SchedulingTopologyHint topologyHint;
+        std::vector<std::string> expectedHosts;
+    };
```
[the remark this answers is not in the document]
csegarragonz: I opted for the struct so that each test configuration is packed nicely and can be passed around easily. Plus, I find it helps tidying up the test set-up process.

I am slightly reluctant to remove it as I feel it helps with readability. Do you have any strong feelings regarding actually keeping it?

tests/test/scheduler/test_executor.cpp
```diff
@@ -29,238 +29,224 @@ namespace tests {
 std::atomic<int> restoreCount = 0;
 std::atomic<int> resetCount = 0;
 
-class TestExecutor final : public Executor
+TestExecutor::TestExecutor(faabric::Message& msg)
```
csegarragonz, line 32: Changes here involve moving from a class declaration + definition to just definition.

tests/test/scheduler/test_scheduling_decisions.cpp
```diff
@@ -0,0 +1,362 @@
...
+    SECTION("Very uneven slot distribution across hosts")
+    {
+        config.slots = { 1, 0, 0, 0 };
+        config.expectedHosts = {
+            masterHost, masterHost, masterHost, masterHost,
+            masterHost, masterHost, masterHost, masterHost
+        };
+    }
+
+    SECTION("Decreasing to one and increasing slot distribution")
```
csegarragonz, line 271: I add this test that before would have failed with `NEVER_ALONE`.

src/scheduler/Scheduler.cpp
```diff
@@ -342,11 +358,24 @@ faabric::util::SchedulingDecision Scheduler::makeSchedulingDecision(
         // At this point there's no more capacity in the system, so we
         // just need to overload locally
         if (remainder > 0) {
-            SPDLOG_DEBUG(
-              "Overloading {}/{} {} locally", remainder, nMessages, funcStr);
+            std::string overloadedHost = thisHost;
+
+            if (topologyHint ==
+                  faabric::util::SchedulingTopologyHint::NEVER_ALONE &&
+                hosts.size() > 0) {
```
Shillaker, line 365: I think this would produce a clang-tidy warning about using `empty` instead, i.e. `!hosts.empty()` rather than `hosts.size() > 0`.

tests/test/scheduler/test_executor.cpp
```diff
@@ -29,238 +29,224 @@ namespace tests {
...
+  : Executor(msg)
+{}
+
+TestExecutor::~TestExecutor() {}
```
Shillaker, line 36: This should also give a clang-tidy warning about using `= default` instead I think.

Shillaker (approved):

Shillaker:
@csegarragonz forgot to add a comment to that approval. Couple of nitpicky comments but other than that LGTM

PR #181, opened 2021-11-25
Retry `GET` from Redis for messages in the Execution Graph

csegarragonz:
In this PR I fix a race condition that happened when running `Scheduler::getFunctionExecGraph()` immediately after `Scheduler::getFunctionResult()` in a distributed setting.

In particular, we sometimes generated the execution graph before all remote executors had published their resulting message. As a consequence, `redis.get` would return an empty byte array, and we'd cast it to an empty `faabric::Message`.

I add a naive retry mechanism (bear in mind _generating_ execution graphs is off the hot path), and a test that reproduces the failure with very little retries (in fact 10 tries might be an overkill).

### DIFF
--- a/src/scheduler/Scheduler.cpp
+++ b/src/scheduler/Scheduler.cpp
@@ -21,6 +21,8 @@
 #include <unordered_set>
 
 #define FLUSH_TIMEOUT_MS 10000
+#define GET_EXEC_GRAPH_SLEEP_MS 500
+#define MAX_GET_EXEC_GRAPH_RETRIES 3
 
 using namespace faabric::util;
 using namespace faabric::snapshot;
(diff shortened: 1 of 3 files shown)


src/scheduler/Scheduler.cpp
```diff
@@ -1012,7 +1012,29 @@ ExecGraphNode Scheduler::getFunctionExecGraphNode(unsigned int messageId)
 
     // Get the result for this message
     std::string statusKey = faabric::util::statusKeyFromMessageId(messageId);
+
+    // We want to make sure the message bytes have been populated by the time
+    // we get them from Redis. For the time being, we retry a number of times
+    // and fail if we don't succeed.
     std::vector<uint8_t> messageBytes = redis.get(statusKey);
+    int maxNumRetries = 3;
```
Shillaker, line 1020: This and the sleep time are magic numbers. As a rule, magic numbers should always be put into constants (i.e. a `#define` at the top of this file).

src/scheduler/Scheduler.cpp
```diff
@@ -1012,7 +1012,29 @@ ExecGraphNode Scheduler::getFunctionExecGraphNode(unsigned int messageId)
...
+    int numRetries = 0;
+    while (messageBytes.empty() && numRetries < maxNumRetries) {
+        SPDLOG_WARN(
+          "Retry GET message for ExecGraph node with id {} (Retry {}/{})",
+          messageId,
+          numRetries + 1,
+          maxNumRetries);
+        SLEEP_MS(500);
+        messageBytes = redis.get(statusKey);
+    }
```
Shillaker, line 1030: Is this actually incrementing the `numRetries` anywhere?

tests/dist/scheduler/test_exec_graph.cpp
```diff
@@ -0,0 +1,52 @@
...
+        auto execGraph = sch.getFunctionExecGraph(m.id());
+        REQUIRE(countExecGraphNodes(execGraph) == nFuncs);
+
+        REQUIRE(execGraph.rootNode.msg.id() == m.id());
+        for (int i = 1; i < nFuncs; i++) {
+            auto node = execGraph.rootNode.children.at(i - 1);
+            REQUIRE(node.msg.id() == req->mutable_messages()->at(i).id());
+        }
+    }
+}
```
Shillaker, line 51: This test isn't checking that it will fail when the max retries are hit, which seems like the easier case to check for.

Shillaker (changes requested):

tests/dist/scheduler/test_exec_graph.cpp
csegarragonz: I add another test that checks that you can't get an execution graph for a message that hasn't been published.

Shillaker (approved):

PR #182, opened 2021-11-25
Adding support for sanitised builds

csegarragonz:
In this PR I introduce a task to build the tests with different sanitisers.

I also add a workflow that runs in parallel to the tests that checks for _newly added_ sanitising errors. It checks against a set of files I commit in this PR containing the output of running the sanitising as of today.

After offline discussion, I think the changes to allow for builds with sanitising are to be kept. The actual workflow file is a different discussion. Before using it, the codebase would have to be in a state where we trigger no warnings. The worflow file is usable and ready to be merged in: it only runs `on: workflow_dispatch` i.e., if manually triggered from the UI.

The low-hanging fruit to reduce a bunch of warnings is:
* Address Sanitiser: #183 
* Thread Sanitiser: disable ZeroMQ-related code and look into our latch implementation.
* Undefined Sanitiser: most warnings come unaligned ints in snapshots.
* Leak Sanitiser: leak sanitiser is a subset of address sanitiser, so we could consider not using it at all.
* Memory Sanitiser: not use in any hypothetical workflow and add errors on `-Wuninitialized` and `-Wmaybe-uninitialized`.


kubasz:
I'd have a couple of general comments on this:
 * In cmake, add an explicit None/empty string elseif branch, and use a `message(FATAL_ERROR)` on wrong values to catch typos
 * Just check the return code from the programs, the logs will sometimes have things like total bytes allocated, etc. so there's no point in trying to diff them
 * If you are capturing logs into a file, use `tee` to redirect it instead of `>` so they also show up in the actions log, otherwise debugging will be difficult - you can also use a always-condition to run a specific action like `cat` regardless of whether the build succeeds or fails

csegarragonz:
@kubasz thanks for your points, I add some replies:
> * In cmake, add an explicit None/empty string elseif branch, and use a message(FATAL_ERROR) on wrong values to catch typos

We usually run CMake through Python tasks, and I add the typo checking there.

> * Just check the return code from the programs, the logs will sometimes have things like total bytes allocated, etc. so there's no point in trying to diff them

Oh so you are saying the output of sanitised runs is non-deterministic? That can be an issue. The problem is that there are now _many_ warnings. What I was trying to do was to check for just newly added warnings.

I was deliberatly not using `tee` for that same reason. The output at the moment is intelligible.

kubasz:
> > * In cmake, add an explicit None/empty string elseif branch, and use a message(FATAL_ERROR) on wrong values to catch typos
> We usually run CMake through Python tasks, and I add the typo checking there.
True, I just like having stringly-typed values checked at every step if it doesn't impact performance. The scripts might evolve at some point and the Cmake or python version might become out of sync, and without the sanitizer it could just silently pass tests.

> > * Just check the return code from the programs, the logs will sometimes have things like total bytes allocated, etc. so there's no point in trying to diff them
> 
> Oh so you are saying the output of sanitised runs is non-deterministic? That can be an issue. The problem is that there are now _many_ warnings. What I was trying to do was to check for just newly added warnings.
I don't remember the exact format of those outputs, but at least leak checking did output general memory statistics when I last used it, so it would be mostly deterministic, but change with any code changes due to added/removed allocations.

> I was deliberatly not using `tee` for that same reason. The output at the moment is intelligible.
Even if it is hard to read, it's useful to have it - and in the actions view you can just collapse that section if you don't want to look at it at this time.

.github/workflows/sanitisers.yml
```diff
@@ -0,0 +1,200 @@
...
+      - name: "Ping redis"
+        run: redis-cli -h redis ping
+      # --- Build dependencies
+      - name: "Build dependencies to be shared by all sanitiser runs"
+        run: inv dev.cmake -b Debug
+      - name: "Upload built CMake dependencies"
+        uses: actions/upload-artifact@v2
+        with:
+          name: cmake-cache
+          path: /build/faabric
```
kubasz, line 43: The conan cache is actually in `/root/.conan`, and it is hashed with build configuration so should generally work well, while caching the cmake output could cause problems (and you force a clean run which deletes the contents of this folder anyway)

csegarragonz: I disabled the clean build (afaict) but it was still not working. I decided to look at it in another moment, but thanks for pointing this out :grinning:

kubasz: Sorry for the wrong information, I just checked the logs and apparently on the actions runners the folder is at ` /github/home/.conan/`, I guess `$HOME/.conan` could work if standard environment variables can work properly in that context? It is `/root/.conan` in the docker build context though.

kubasz: You can also use `conan config home` to print out the path to shell reliably

csegarragonz: Tilde substitution [should work](https://github.com/actions/upload-artifact#environment-variables-and-tilde-expansion) for the `path` variable of the `upload-artifact` action. Let's see.

kubasz: It looks like the caching system doesn't handle this many output files very well, it's been uploading for a few minutes now and it's only at 8%. Tarring the folder like in <https://github.com/actions/upload-artifact#maintaining-file-permissions-and-case-sensitive-files> and untarring it on extract could help? If it doesn't, the upload takes longer than rebuilding the dependencies so I'm not sure if it's worth it

kubasz: @csegarragonz I'm happy to tinker around with this and try to fix it if you're working on other things, I'm somewhat interested in running the sanitizers myself too

csegarragonz: @kubasz i'm working on other things, and every now and then checking this; so very glad if you can take over :smile:

kubasz:
This is now ready for review, the only thing that I would potentially look into before merging is the MPI TSan failures (currently disabled in the ignore list file), send/recv do some things with buffers that I don't quite understand and TSan sees it as a data race - it might be a false positive (if the synchronization happens through sockets which is invisible for tsan), or not.

CMakeLists.txt
```diff
@@ -22,6 +22,19 @@ set(CMAKE_EXE_LINKER_FLAGS "-fuse-ld=lld")
...
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
+elseif (FAABRIC_USE_SANITISER STREQUAL "Thread")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=thread -fsanitize-ignorelist=${CMAKE_CURRENT_LIST_DIR}/thread-sanitizer-ignorelist.txt")
+elseif (FAABRIC_USE_SANITISER STREQUAL "Undefined")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined")
+elseif (FAABRIC_USE_SANITISER STREQUAL "Leak")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=leak")
+elseif (FAABRIC_USE_SANITISER STREQUAL "Memory")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=memory")
+endif()
```
Shillaker, line 36: Although we do have a check in Python, I'd say it's still worth adding a failure here if we get a non-matching, non-empty `FAABRIC_USE_SANITISER` value, especially as the list of options is quite long and case sensitive. It would be easy for the Python and CMake checks to get out of sync, but if we put the check here, we can remove the one from Python and remove the risk of out-of-syncness.

src/redis/Redis.cpp
```diff
@@ -594,10 +572,8 @@ void Redis::setLong(const std::string& key, long value)
 {
     // Format is NOT printf compatible contrary to what the docs say, hence %i
     // instead of %l.
-    auto reply =
-      (redisReply*)redisCommand(context, "SET %s %i", key.c_str(), value);
-
-    freeReplyObject(reply);
+    auto reply = wrapReply(
+      (redisReply*)redisCommand(context, "SET %s %i", key.c_str(), value));
```
Shillaker, line 576: Does this not cause an unused variable warning? Can we now just call `wrapReply` and not do any cleanup?

src/redis/Redis.cpp
```diff
@@ -324,171 +320,155 @@ void Redis::flushPipeline(long pipelineLength)
...
     return ttl;
 }
 
 void Redis::expire(const std::string& key, long expiry)
 {
-    auto reply =
-      (redisReply*)redisCommand(context, "EXPIRE %s %d", key.c_str(), expiry);
-    freeReplyObject(reply);
+    auto reply = wrapReply(
+      (redisReply*)redisCommand(context, "EXPIRE %s %d", key.c_str(), expiry));
```
Shillaker, line 471: Same point re. unused var here.

src/redis/Redis.cpp
```diff
@@ -324,171 +320,155 @@ void Redis::flushPipeline(long pipelineLength)
...
     long long int result = reply->integer;
-    freeReplyObject(reply);
     return result;
 }
 
 void Redis::flushAll()
 {
-    auto reply = (redisReply*)redisCommand(context, "FLUSHALL");
-    freeReplyObject(reply);
+    auto reply = wrapReply((redisReply*)redisCommand(context, "FLUSHALL"));
```
Shillaker, line 441: Same point re. unused var here.

src/redis/Redis.cpp
```diff
@@ -324,171 +320,155 @@ void Redis::flushPipeline(long pipelineLength)
...
-    freeReplyObject(reply);
 }
 
 void Redis::srem(const std::string& key, const std::string& value)
 {
-    auto reply = (redisReply*)redisCommand(
-      context, "SREM %s %s", key.c_str(), value.c_str());
-    freeReplyObject(reply);
+    auto reply = wrapReply((redisReply*)redisCommand(
+      context, "SREM %s %s", key.c_str(), value.c_str()));
```
Shillaker, line 347: Same point re. unused var here (I'll stop adding these now as there are a few).

src/redis/Redis.cpp
```diff
@@ -11,6 +11,11 @@
 
 namespace faabric::redis {
 
+UniqueRedisReply wrapReply(redisReply* r)
```
Shillaker, line 14: Looks like we have to do a cast to `redisReply` every time we call this, so making the argument whatever `redisCommand` returns and moving the `redisReply` cast into here here would save a few characters on each call.

To make it really succinct, you might me able to move the call to `redisCommand` into here too and use varargs to forward the arguments from this function. Could call it something like `safeRedisCommand` instead.

src/scheduler/Executor.cpp
```diff
@@ -66,14 +66,18 @@ void Executor::finish()
         threadTaskQueues[i].enqueue(
           ExecutorTask(POOL_SHUTDOWN, nullptr, nullptr, false, false));
 
+        faabric::util::UniqueLock threadsLock(threadsMutex);
+        // copy shared_ptr to avoid racing
```
Shillaker, line 70: Style in the repo is to start comments with capital letters (but unfortunately couldn't find a rule to enforce it).

src/scheduler/Scheduler.cpp
```diff
@@ -95,16 +95,29 @@ void Scheduler::reset()
...
+    {
+        faabric::util::FullLock lock(mx);
+        deadExecutorsList = deadExecutors;
+    }
     for (auto& e : deadExecutors) {
         e->finish();
     }
-
+    deadExecutorsList.clear();
+    faabric::util::FullLock lock(mx);
```
Shillaker, line 120: I'd add a newline and comment between `deadExecutorsList.clear();` and `faabric::util::FullLock lock(mx);` to make it obvious where the lock is being retaken and why. It's easy to miss this in the block as it is.

src/state/StateKeyValue.cpp
```diff
@@ -236,7 +236,7 @@ void StateKeyValue::flagChunkDirty(long offset, long len)
 void StateKeyValue::markDirtyChunk(long offset, long len)
 {
     isDirty |= true;
-    memset(BYTES(dirtyMask) + offset, ONES_BITMASK, len);
+    memset(dirtyMask.get() + offset, ONES_BITMASK, len);
```
Shillaker, line 239: I know this was like this before, but in general we're trying to use a leading `::` on functions from the global namespace to avoid confusion. No need to change these now (unless you want to), but just a note if coming across them in future. 

Another thing I looked for a `clang-format`/`-tidy` rule for but couldn't find.

Shillaker (changes requested):
@kubasz @csegarragonz awesome work guys, this looks great. A few nitpicky comments mixed in with some small changes/ questions.

.github/workflows/sanitisers.yml
```diff
@@ -0,0 +1,225 @@
...
+      # --- Cache based on Conan version and ExternalProjects cmake source
+      - name: Get Conan version
+        id: get-conan-version
+        run: |
+          echo "::set-output name=conan-version::$(conan --version | tr -d '[:alpha:] ')"
+        shell: bash
+      - uses: actions/cache@v2
+        with:
+          path: '~/.conan'
+          key: ${{ runner.os }}-${{ steps.get-conan-version.outputs.conan-version }}-${{ hashFiles('cmake/ExternalProjects.cmake') }}
```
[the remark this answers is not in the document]
kubasz: 430 MB, the limit is 10GB per repo (<https://github.com/actions/cache>), and this should be a cache hit most times as the dependency versions don't change all that often

.github/workflows/sanitisers.yml
```diff
@@ -0,0 +1,225 @@
...
+      # --- Build dependencies
+      - name: "Build dependencies to be shared by all sanitiser runs"
+        run: inv dev.cmake -b Debug
+
+  address-sanitiser:
```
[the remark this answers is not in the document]
kubasz: It all runs in parallel, and each sanitizer requires a full rebuild of the source (but not dependencies), so this saves about 20 mins of time on tests

src/redis/Redis.cpp
kubasz: Yes, I was batch-fixing this and didn't notice the specific cases here, will fix in a sec

.github/workflows/sanitisers.yml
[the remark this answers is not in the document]
csegarragonz: I also think this will be a great speedup to the runtime of general tests if we use the caching there (#187), but @kubasz it looks as though we are using the wrong key: https://github.com/faasm/faabric/runs/4337385997?check_suite_focus=true#step:7:4

(if the logs are to be trusted of course)

could you double-check?

.github/workflows/sanitisers.yml
```diff
@@ -0,0 +1,225 @@
+name: Sanitiser Checks
+
+# on: workflow_dispatch
```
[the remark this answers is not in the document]
kubasz: It runs on separate VMs, I removed the commented line

.github/workflows/sanitisers.yml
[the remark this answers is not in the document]
kubasz: Huh, seems that hashFiles doesn't work, let me look into it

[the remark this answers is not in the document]
kubasz: I had to replace it with manual hashing in bash+sha256sums, it works in the most recent commit. @csegarragonz This change will probably need porting over to #187

thread-sanitizer-ignorelist.txt
```diff
@@ -0,0 +1,13 @@
...
+race:close
+race:socket
+# Ignore ZeroMQ races
+race:zmq::*
+# Config only changes in tests, and in places where being slightly racy doesn't matter
+race:faabric::util::SystemConfig::*
+# Catch2 allocates in its signal handler, this prevents showing the wrong crash report
+signal:*
+
+# TODO: Remove: There's something weird going on with MPI code I don't understand
```
csegarragonz, line 12: So for me to see the races tsan is complaining about I'd just have to re-run commenting out the next line right?

kubasz: Yes, it would terminate on the first detected race. For me that was in the AllGather implementation

src/redis/Redis.cpp
kubasz: I added a `safeRedisCommand` as suggested, `wrapReply` is still used in a couple of places where it wraps things not directly coming out of `redisCommand` calls

csegarragonz:
The only thing I would add to Simon's comments is to see if you can bump the faabric dependency in faasm as well; just to double check nothing is broken.

.github/workflows/sanitisers.yml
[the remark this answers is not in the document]
Shillaker: Ok cool, is that in theory or in practice? I'm just surprised that GHA gives out resources that freely.

Shillaker (approved):
Awesome LGTM 👍

kubasz:
It looks like a faasm distributed test fails with this merged (on top of PR <https://github.com/faasm/faasm/pull/540>):
@csegarragonz Does this look like it could be caused directly by my changes, or is it something to do with the races in distributed tests?
```
[20:44:04:619] [1] [info] =============================================
[20:44:04:619] [1] [info] TEST: Test running an MPI function spanning two hosts
[20:44:04:619] [1] [info] =============================================
[20:44:04:619] [1] [debug] Resetting scheduler
[20:44:04:619] [1] [debug] Resetting scheduler thread-local cache for thread 1
[20:44:04:620] [1] [debug] Resetting dirty tracking after pushing diffs mpi/mpi_bcast
[20:44:04:620] [1] [debug] Resetting dirty tracking
[20:44:04:620] [1] [debug] Scheduling 1/1 calls to mpi/mpi_bcast locally
[20:44:04:622] [1] [debug] Scaling mpi/mpi_bcast from 0 -> 1
[20:44:04:622] [1] [debug] Starting executor 172.18.0.7_966818832
[20:44:04:622] [1] [debug] WAVM module cache initialising mpi/mpi_bcast
[20:44:04:622] [1] [debug] Loading main module mpi/mpi_bcast
[20:44:04:631] [1] [debug] Instantiating module mpi/mpi_bcast  
[20:44:04:633] [1] [debug] Finished instantiating module mpi/mpi_bcast  
[20:44:04:633] [1] [debug] Opened preopened fd at /
[20:44:04:633] [1] [debug] Opened preopened fd at .
[20:44:04:633] [1] [debug] Creating 2 thread stacks
[20:44:04:633] [1] [debug] Successfully executed __wasm_call_ctors for mpi/mpi_bcast
[20:44:04:633] [1] [debug] Successfully executed zygote for mpi/mpi_bcast
[20:44:04:633] [1] [debug] heap_top=11075584 initial_pages=169 initial_table=5
[20:44:04:648] [1] [debug] Wrote snapshot mpi/mpi_bcast_reset to fd 96
[20:44:04:648] [37] [debug] Thread pool thread 172.18.0.7_966818832:0 starting up
[20:44:04:648] [37] [debug] Added thread id 37 to /sys/fs/cgroup/cpu/faasm/tasks
[20:44:04:648] [37] [debug] Not using network ns, support off
[20:44:04:649] [37] [debug] S - args_sizes_get - 4194296 4194300
[20:44:04:649] [37] [debug] S - sbrk - 0
[20:44:04:649] [37] [debug] S - sbrk - 65536
[20:44:04:649] [37] [debug] S - args_get - 11075632 11075600
[20:44:04:649] [37] [debug] S - MPI_Init (create) 0 0
[20:44:04:649] [37] [debug] Initialising world 966818837
[20:44:04:650] [37] [debug] Adding new function call client for 172.18.0.6
[20:44:04:652] [37] [debug] Resetting dirty tracking after pushing diffs mpi/mpi_bcast
[20:44:04:652] [37] [debug] Resetting dirty tracking
[20:44:04:652] [37] [debug] Scheduling 2/3 calls to mpi/mpi_bcast on 172.18.0.6
[20:44:04:653] [37] [debug] Scheduling 1/3 calls to mpi/mpi_bcast locally
[20:44:04:653] [37] [debug] Scaling mpi/mpi_bcast from 1 -> 2
[20:44:04:653] [37] [debug] Starting executor 172.18.0.7_966818849
[20:44:04:662] [37] [debug] MPI-0 S - MPI_Comm_rank 4196616 4194284
[20:44:04:662] [37] [debug] MPI-0 S - MPI_Comm_size 4196616 4194280
[20:44:04:662] [37] [debug] MPI-0 S - MPI_Bcast 4194240 4 4196620 2 4196616
[20:44:04:663] [38] [debug] Thread pool thread 172.18.0.7_966818849:0 starting up
[20:44:04:663] [38] [debug] Added thread id 38 to /sys/fs/cgroup/cpu/faasm/tasks
[20:44:04:663] [38] [debug] Not using network ns, support off
[20:44:04:663] [38] [debug] S - args_sizes_get - 4194296 4194300
[20:44:04:663] [38] [debug] S - sbrk - 0
[20:44:04:663] [38] [debug] S - sbrk - 65536
[20:44:04:663] [38] [debug] S - args_get - 11075632 11075600
[20:44:04:663] [38] [debug] S - MPI_Init (join) 0 0
[20:44:04:663] [38] [debug] MPI-1 S - MPI_Comm_rank 4196616 4194284
[20:44:04:663] [38] [debug] MPI-1 S - MPI_Comm_size 4196616 4194280
[20:44:04:663] [38] [debug] MPI-1 S - MPI_Bcast 4194240 4 4196620 2 4196616
[20:44:04:820] [37] [debug] S - fd_fdstat_get - 1 4194200 (/dev/stdout)
Broadcast succeeded
[20:44:04:821] [37] [debug] MPI-0 S - MPI_Finalize
[20:44:04:821] [37] [debug] Resetting after mpi/mpi_bcast:966818830 (snap key mpi/mpi_bcast_reset)
[20:44:04:823] [38] [debug] S - fd_fdstat_get - 1 4194200 (/dev/stdout)
Broadcast succeeded
[20:44:04:823] [38] [debug] MPI-1 S - MPI_Finalize
[20:44:04:824] [37] [debug] Memory already correct size for snapshot (11075584)
[20:44:04:824] [38] [debug] Resetting after mpi/mpi_bcast:966818840 (snap key mpi/mpi_bcast_reset)

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
dist_tests is a Catch v2.13.7 host application.
Run with -? for options

-------------------------------------------------------------------------------
Test running an MPI function spanning two hosts
-------------------------------------------------------------------------------
/usr/local/code/faasm/tests/dist/mpi/test_remote_execution.cpp:13
...............................................................................

/usr/local/code/faasm/tests/dist/mpi/test_remote_execution.cpp:44: FAILED:
  REQUIRE( hosts.size() == 2 )
with expansion:
  3 == 2

```

csegarragonz:
@kubasz yes, this should be fixed by #181

kubasz:
@csegarragonz It won't contain it as this is a couple of commits behind on faabric, and the way faabric replacement works is it fully replaces the tree with this one. I will merge master and rerun the tests just to make sure, then it'll be ready to merge (I can't as I don't have perms on this repo)

kubasz:
This is a moment where a tool like https://github.com/bors-ng/bors-ng is useful, but I don't think faasm is at a scale where it's worth setting up yet, it's just a temporary influx of PRs with all of us working on some changes right now

csegarragonz:
@kubasz yes, this is ready to go; once it's ready ping me on here and i will merge it. thanks again, this looks great

kubasz:
Huh `Warning: Docker pull failed with exit code 143, back off 2.786 seconds before retry.` when fetching containers for ubsan

kubasz:
@csegarragonz I think it's ready to merge, it fails on the same things that your PR was failing in faasm, and only in dist-tests, so I don't think it needs to be blocked by that

PR #186, opened 2021-11-29
Re-factor MPI native tests to run as distributed tests

csegarragonz:
In this PR I adapt all the existing `mpi-native` tests to run in our current distributed test setting. Most of the code in place can be re-used with very minimal changes.

Two tests are left with the `[.]` tag as they used `MPI_Win_create` which we removed some time ago.


src/scheduler/ExecGraph.cpp
```diff
@@ -43,6 +43,33 @@ std::set<std::string> getExecGraphHosts(const ExecGraph& graph)
     return getExecGraphHostsForNode(graph.rootNode);
 }
 
+std::vector<std::string> getMpiRankHostsFromExecGraphNode(
```
csegarragonz, line 46: I reckon this method may be too MPI-specific for `ExecGraph.cpp` but I am not sure where to put it.

.github/workflows/tests.yml
```diff
@@ -112,26 +112,7 @@ jobs:
...
-      - name: "Build MPI native library"
-        run: inv dev.cc faabricmpi_native --shared
-      - name: "Install MPI native library"
-        run: inv dev.install faabricmpi_native --shared
       - name: "Build examples"
         run: inv examples
       - name: "Run example to check"
         run: inv examples.execute check
-
-  mpi_native:
```
csegarragonz, line 124: The `mpi_native` tests will now run as part of the distributed tests. What also means that will be checked against each new commit in a PR (not just when bumping docker images).

tests/dist/mpi/examples/mpi_allgather.cpp
```diff
@@ -1,15 +1,9 @@
 #include <faabric/mpi/mpi.h>
```
csegarragonz, line 1: For all MPI examples I've just had to remove the `main` function, and change the method name and namespace.

Note that I've opted to add a nested namespace, `tests::mpi`, as some function names where quite general.

csegarragonz:
May re-open this in the future.

PR #187, opened 2021-11-29
Re-use conan cache for built dependencies

csegarragonz:
I cherry pick this change from #182, and faasm/faasm#542.

Upon cache hit the tests workflow file runs in under 7'.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,7 +37,7 @@ jobs:
       - name: "Check C/C++ formatting changes"
         run: git diff --exit-code
 
-  tests:
+  conan-cache:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -37,8 +37,50 @@ jobs:
       - name: "Check C/C++ formatting changes"
         run: git diff --exit-code
 
+  conan-cache:
```
Shillaker, line 40: See question on https://github.com/faasm/faabric/pull/182 re. how large this cache will be and GHA limit on cache size.

kubasz:
@csegarragonz This needs updating to the different cache key generation as now present in #182 to fix the broken hashFiles call

kubasz:
Looks like examples need a separate cache key, as they are building in release mode rather than debug like the other tests.

.github/workflows/tests.yml
```diff
@@ -105,6 +149,25 @@ jobs:
...
+        shell: bash
+      - name: "Hash ExternalProjects"
+        id: hash-external
+        run: |
+          echo "::set-output name=hash-external::$(sha256sum cmake/ExternalProjects.cmake | cut '-d ' -f 1)"
+        shell: bash
+      - uses: actions/cache@v2
+        with:
+          path: '~/.conan'
+          key: ${{ runner.os }}-${{ steps.get-conan-version.outputs.conan-version }}-${{ steps.hash-external.outputs.hash-external }}-release
```
kubasz, line 166: `-release` Should be at the beginning of the key, as the restore-keys matches any keys starting with the given strings, so it would restore the debug key instead of forcing a fresh release key.

kubasz (changes requested):
Only a small change needed re. release cache keys, otherwise lgtm

.github/workflows/tests.yml
Shillaker: To clarify, we want a completely different key space for the `release` versions, so we'd have:

```
key: release-${{ runner.os }}-${{ steps.get-conan-version.outputs.conan-version }}-${{ steps.hash-external.outputs.hash-external }}
restore-keys: |
  release-${{ runner.os }}-${{ steps.get-conan-version.outputs.conan-version }}-
  release-${{ runner.os }}-
```

.github/workflows/tests.yml
```diff
@@ -61,6 +61,24 @@ jobs:
       # --- Set-up ---
       - name: "Ping redis"
         run: redis-cli -h redis ping
+      # --- Cache based on Conan version and ExternalProjects cmake source
+      - name: Get Conan version
+        id: get-conan-version
+        run: |
+          echo "::set-output name=conan-version::$(conan --version | tr -d '[:alpha:] ')"
+        shell: bash
```
Shillaker, line 69: Rather than repeating `shell: bash` on these commands, can we set `shell: bash` in defaults? May apply to other workflow files.

Shillaker (changes requested):

.github/workflows/tests.yml
csegarragonz: yes, i did not know it but you should be able to do [this](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-5):
```yaml
defaults:
  run:
    shell: bash
```

kubasz (approved):

Shillaker (approved):
Nice, LGTM 👍

PR #188, opened 2021-11-30
Enable/disable scheduling topology hints through an env. variable

csegarragonz:
With this PR scheduling topology hints may be disabled by setting the environment variable: `NO_TOPOLOGY_HINTS=on` in faam's worker.

### DIFF
--- a/src/scheduler/Scheduler.cpp
+++ b/src/scheduler/Scheduler.cpp
@@ -284,7 +284,10 @@ faabric::util::SchedulingDecision Scheduler::makeSchedulingDecision(
     std::string funcStr = faabric::util::funcToString(firstMsg, false);
 
     // If topology hints are disabled, unset the provided topology hint
-    if (conf.useTopologyHints == "off") {
+    if (conf.noTopologyHints == "on" &&
+        topologyHint != faabric::util::SchedulingTopologyHint::NORMAL) {
+        SPDLOG_WARN("Ignoring topology hint passed to scheduler as hints are "
+                    "disabled in the config");
         topologyHint = faabric::util::SchedulingTopologyHint::NORMAL;
     }
 
--- a/src/scheduler/Scheduler.cpp
+++ b/src/scheduler/Scheduler.cpp
@@ -283,6 +283,11 @@ faabric::util::SchedulingDecision Scheduler::makeSchedulingDecision(
     faabric::Message& firstMsg = req->mutable_messages()->at(0);
     std::string funcStr = faabric::util::funcToString(firstMsg, false);
 
+    // If topology hints are disabled, unset the provided topology hint
+    if (conf.useTopologyHints == "off") {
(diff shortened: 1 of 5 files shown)


src/scheduler/Scheduler.cpp
```diff
@@ -283,6 +283,11 @@ faabric::util::SchedulingDecision Scheduler::makeSchedulingDecision(
     faabric::Message& firstMsg = req->mutable_messages()->at(0);
     std::string funcStr = faabric::util::funcToString(firstMsg, false);
 
+    // If topology hints are disabled, unset the provided topology hint
+    if (conf.useTopologyHints == "off") {
+        topologyHint = faabric::util::SchedulingTopologyHint::NORMAL;
+    }
```
Shillaker, line 289: I would change this to check if they're requesting anything other than `NORMAL`, then print a warning if so. We don't want to have this on accidentally.

src/util/config.cpp
```diff
@@ -100,6 +101,7 @@ void SystemConfig::print()
     SPDLOG_INFO("--- Scheduling ---");
     SPDLOG_INFO("NO_SCHEDULER               {}", noScheduler);
     SPDLOG_INFO("OVERRIDE_CPU_COUNT         {}", overrideCpuCount);
+    SPDLOG_INFO("USE_TOPOLOGY_HINTS         {}", useTopologyHints);
```
Shillaker, line 104: As this is switching off something we want on by default, i would change this to `NO_TOPOLOGY` and `noTopology`.

tests/test/util/test_config.cpp
```diff
@@ -22,6 +22,7 @@ TEST_CASE("Test default system config initialisation", "[util]")
```
Shillaker, line 22: Missing a test to check that the flag actually switches off topology in the scheduler.

Shillaker (changes requested):

src/scheduler/Scheduler.cpp
csegarragonz: Good point. Now we:
1. Check if env var set
2. If set and hint different than `NORMAL`, print a warning and unset.

tests/test/scheduler/test_scheduling_decisions.cpp
```diff
@@ -186,6 +186,37 @@ TEST_CASE_METHOD(SchedulingDecisionTestFixture,
     testActualSchedulingDecision(req, config);
 }
 
+TEST_CASE_METHOD(SchedulingDecisionTestFixture,
+                 "Test scheduling hints can be disabled through the config",
+                 "[scheduler]")
+{
+    SchedulingConfig config = {
```
csegarragonz, line 193: Here I add the new test

csegarragonz:
I every now and then get errors in the sanitised builds unrelated to the changes introduced. Moving the discussion about this offline.

Shillaker (approved):

Issue #189, opened 2021-12-01
Potential/ confirmed concurrency issues


csegarragonz:
This issue is a list of all the potentially/ confirmed dodgy bits of concurrent code.

Useful links:

- [Clang TS analysis](https://clang.llvm.org/docs/ThreadSafetyAnalysis.html)
- [C++ memory order](https://en.cppreference.com/w/cpp/atomic/memory_order)
- [C++ memory model](https://en.cppreference.com/w/cpp/language/memory_model)

## Miscellaneous

- [ ] Reading from a variables without taking a lock, while writing to it through a mutex. Instead, wrap single variables in an atomic, and use acquire/release accesses. For groups of variables, make sure to take a shared lock. 
- [ ] Double locking without atomic accesses (i.e. `if(!x) {lock(); if(!x) {}}` if `x` is not atomic). Instead make `x` atomic.
- [ ] Use of maps as registries for looking up objects like snapshots, distributed coordination groups, and message queues. These are accessed concurrently and often include some lazy-initialisation, all of which is guarded by mutexes scattered around the place. Could experiment with concurrent containers such as [`libcds`](https://github.com/khizmax/libcds) and [`phmap`](https://github.com/greg7mdp/parallel-hashmap). 
- [ ] Returning map values without an associated lock: often some `cacheMap.get(key)` operation is protected by a mutex (or in the future a concurrent hash map), but if it returns a reference or pointer, `cacheMap.clear()` from another thread can invalidate that reference. Make sure to either return `std::shared_ptr`s where appropriate, or a pair of a reference with a `SharedLock` on the entire cache to make sure nothing can invalidate that reference while it's being used outside the `get` implementation.

## Thread sanitizer

List of known issues raised by the thread sanitiser, together with details to fix them. 

## Leak sanitizer

- [ ] Segmentation fault in snapshot test [`"TEST: Test snapshot diffs with merge ops"`](https://github.com/faasm/faabric/runs/4504435160?check_suite_focus=true#step:10:1200).
- [ ] Wrong test result in test [`"TEST: Test direct messaging"`](https://github.com/faasm/faabric/runs/4739026659?check_suite_focus=true#step:11:1828)

kubasz:
For the first one: take a shared lock in getLockOwner(), alternatively convert the lockOwnerIdx to an atomic int and do acquire loads/release stores (free on x86_64)

PR #192, opened 2021-12-06
Minor fixes

kubasz:
Minor code tidying fixes for a bunch of compiler warnings:
 * Adding default initializers to fields without one
 * Remove padding by reordering members in a couple of structs
 * A few trivial copy->move changes
 * Use `= default`ed constructors and destructors where appropriate
 * Also fixes a flaky PTP test, it was modifying an atomic just after rather than before the barrier.
 * PTP group race condition fix
 * Attempt to make scheduler tests less unreliable under TSan: with tsan overheads, I think the executor was able to finish executing its dummy task before the scheduler was able to execute the next one, resulting in re-use where it normally wouldn't occur

This should fix all of the current TSan issues listed in #189

### DIFF
--- a/include/faabric/scheduler/MpiWorld.h
+++ b/include/faabric/scheduler/MpiWorld.h
@@ -230,7 +230,7 @@ class MpiWorld
 
     // Track ranks that are local to this world, and local/remote leaders
     // MPITOPTP - this information exists in the broker
-    int localLeader;
+    int localLeader = -1;
     std::vector<int> localRanks;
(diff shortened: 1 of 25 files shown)


Shillaker (approved):
Awesome, LGTM 👍

kubasz:
@Shillaker can you merge? "The base branch restricts merging to authorized users"

Shillaker:
@kubasz think I've given you sufficient permissions now. Let me know if not.

PR #195, opened 2021-12-09
Fix bug in `MPI_Reduce`

csegarragonz:
In this PR I solve a bug in distributed reduce operations that re-use the send and receive buffers across different `MPI_Reduce` calls. This bug was first introduced with #191.

When performing the locality-aware reduce in a local leader that is not co-located with the rank that originated the reduce, we perform the following operations:
1. Receive data from local ranks.
2. Reduce data from local ranks (and ours).
3. Send reduced results to the rank that originated the reduce.

As the local leader is not expected to be receiving anything, it may be the case that, the `recvBuffer` in the method's signature is a `nullptr` (seems to be an unwritten agreement across MPI developers). Therefore, in (2) we were accumulating the partial reduce in the same buffer the local leader's data was provided (i.e. `sendBuffer`), and then sending this buffer in (3).

Unofrtunately, said `sendBuffer` is a pointer to the application's memory, and the application is not expecting that memory to be modified. In particular, if subsequent reduces are performed, the results returned by reduce are incorrect.

### DIFF
--- a/src/scheduler/MpiWorld.cpp
+++ b/src/scheduler/MpiWorld.cpp
@@ -1106,6 +1106,12 @@ void MpiWorld::reduce(int sendRank,
         // the receiver is not co-located with us, do a reduce with the data of
         // all our local ranks, and then send the result to the receiver
         if (getHostForRank(recvRank) != thisHost) {
+            // In this step we reduce our local ranks data. It is important
+            // that we do so in a copy of the send buffer, as the application
+            // does not expect said buffer's contents to be modified.
+            auto sendBufferCopy = std::make_unique<uint8_t[]>(bufferSize);
+            memcpy(sendBufferCopy.get(), sendBuffer, bufferSize);
+
             for (const int r : localRanks) {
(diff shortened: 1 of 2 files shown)


kubasz:
Does this fix any of the ignored mpi TSan race conditions by any chance? IIRC it was failing on reductions

Shillaker (approved):

csegarragonz:
@kubasz trying now. whilst we are at it, does #192 fix the ignore list entry [here](https://github.com/faasm/faabric/blob/3d6ea14588c0dbd1adad58ee410024dbc48b656b/thread-sanitizer-ignorelist.txt#L16)?

kubasz:
> @kubasz trying now. whilst we are at it, does #192 fix the ignore list entry [here](https://github.com/faasm/faabric/blob/3d6ea14588c0dbd1adad58ee410024dbc48b656b/thread-sanitizer-ignorelist.txt#L16)?

Yes, it does, I forgot to remove it from the ignorelist

PR #197, opened 2021-12-13
Async executed time

csegarragonz:


thread-sanitizer-ignorelist.txt
```diff
@@ -11,6 +11,3 @@ signal:*
 
 # TODO: Remove: There's something weird going on with MPI code I don't understand
 race:faabric::scheduler::MpiWorld::*
-
```
csegarragonz, line 14: I also remove this from here as discussed offline with jakub (don't think it is worth tagging them here).

Shillaker:
Parsing a string like this will end up being quite fragile and we'll have to hard-code the output format in all the clients that consume it. 

Instead it would be more robust and flexible to make the return value a JSON string dump of the `result` object given that it's already there.

Shillaker (changes requested):
Parsing a string like this is quite fragile and a bit of a hack, plus we'll have to hard-code the output format in all the clients that consume it. 

Instead it would be more robust and flexible to make the return value a JSON string dump of the `result` object given that it's already there. This is probably how it should have been in the first place to be honest, I don't remember why I did the strings.

Admittedly this will break all existing clients expecting those strings, but we can do a one-off update of each one and hopefully it'll be more useful long term. We should be able to use `util::messageToJson`.

PR #198, opened 2021-12-13
Locality aware gather

csegarragonz:
In this PR I change the implementation for the gather collective communication algorithm to minimise cross-host messaging as we previously did with broadcast and reduce.

Gather is slightly more tricky than broadcast and reduce as every rank contributes with a different bit of data, and the order of these bits must be preserved. This means that each remote leader sends a message with the to-be-gathered data from its local ranks coalesced. The receiver must ensure that the bits are then correctly re-mapped in the receive buffer.

To implement gather, I found very useful to have a dict where the key was the host and the value a vector with all the ranks assigned to that host. As there was already an overload of methods and variables tracking rank-to-host infromation, I re-factor the three collective communications algorithms to use the same dictionary. Further, I set the remote leader for each host to be the first element in said vector.

This means that I can get rid of the `localRanks`, and `remoteLeaders` vectors. I keep the local leader variable, but it could also be removed. Lastly, it felt natural that this dict was named `ranksForHost`, however, this was very close to the `rankHosts` vector we use to broadcast the mappings upon world creation. Given that the latter actually stores the host for a given rank, I re-factor it to `hostForRank`.

Note that it is preferrable to maintain the `hostForRank` vector (even though we use the dict around) as it is easier to broadcast and receive.

### DIFF
--- a/tests/test/scheduler/test_remote_mpi_worlds.cpp
+++ b/tests/test/scheduler/test_remote_mpi_worlds.cpp
@@ -1297,4 +1297,124 @@ TEST_CASE_METHOD(RemoteMpiTestFixture,
     otherWorld.destroy();
     thisWorld.destroy();
 }
+
+std::set<int> getMsgCountsFromMessages(
+  std::vector<std::shared_ptr<faabric::MPIMessage>> msgs)
+{
+    std::set<int> counts;
+    for (const auto& msg : msgs) {
+        counts.insert(msg->count());
+    }
+
+    return counts;
+}
+
+TEST_CASE_METHOD(RemoteMpiTestFixture,
+                 "Test number of messages sent during gather",
+                 "[mpi]")
+{
+    int worldSize = 4;
+    setWorldSizes(worldSize, 2, 2);
+    std::vector<int> messageData = { 0, 1, 2 };
+    int nPerRank = messageData.size();
+
+    // Init worlds
+    MpiWorld& thisWorld = getMpiWorldRegistry().createWorld(msg, worldId);
+    faabric::util::setMockMode(true);
+    thisWorld.broadcastHostsToRanks();
+    REQUIRE(getMpiHostsToRanksMessages().size() == 1);
+    otherWorld.initialiseFromMsg(msg);
+
+    std::set<int> expectedSentMsgRanks;
+    std::set<int> expectedSentMsgCounts;
+    int expectedNumMsgSent;
+    int sendRank;
+    int recvRank;
+
+    SECTION("Call gather from receiver (local), and receiver is local leader")
+    {
+        recvRank = 0;
+        sendRank = recvRank;
+        expectedNumMsgSent = 0;
+        expectedSentMsgRanks = {};
+        expectedSentMsgCounts = {};
+    }
+
+    SECTION(
+      "Call gather from receiver (local), and receiver is non-local leader")
+    {
+        recvRank = 1;
+        sendRank = recvRank;
+        expectedNumMsgSent = 0;
+        expectedSentMsgRanks = {};
+        expectedSentMsgCounts = {};
+    }
+
+    SECTION("Call gather from non-receiver, colocated with receiver, and local "
+            "leader")
+    {
+        recvRank = 1;
+        sendRank = 0;
+        expectedNumMsgSent = 1;
+        expectedSentMsgRanks = { recvRank };
+        expectedSentMsgCounts = { nPerRank };
+    }
+
+    SECTION("Call gather from non-receiver, colocated with receiver")
+    {
+        recvRank = 0;
+        sendRank = 1;
+        expectedNumMsgSent = 1;
+        expectedSentMsgRanks = { recvRank };
+        expectedSentMsgCounts = { nPerRank };
+    }
+
+    SECTION("Call gather from non-receiver rank, not colocated with receiver, "
+            "but local leader")
+    {
+        recvRank = 0;
+        sendRank = 2;
+        expectedNumMsgSent = 1;
+        expectedSentMsgRanks = { recvRank };
+        expectedSentMsgCounts = { 2 * nPerRank };
+    }
+
+    SECTION("Call gather from non-receiver rank, not colocated with receiver")
+    {
+        recvRank = 0;
+        sendRank = 3;
+        expectedNumMsgSent = 1;
+        expectedSentMsgRanks = { 2 };
+        expectedSentMsgCounts = { nPerRank };
+    }
+
+    std::vector<int> gatherData(worldSize * nPerRank);
+    if (sendRank < 2) {
+        thisWorld.gather(sendRank,
+                         recvRank,
+                         BYTES(messageData.data()),
+                         MPI_INT,
+                         nPerRank,
+                         BYTES(gatherData.data()),
+                         MPI_INT,
+                         nPerRank);
+    } else {
+        otherWorld.gather(sendRank,
+                          recvRank,
+                          BYTES(messageData.data()),
+                          MPI_INT,
+                          nPerRank,
+                          BYTES(gatherData.data()),
+                          MPI_INT,
+                          nPerRank);
+    }
+    auto msgs = getMpiMockedMessages(sendRank);
+    REQUIRE(msgs.size() == expectedNumMsgSent);
+    REQUIRE(getReceiversFromMessages(msgs) == expectedSentMsgRanks);
+    REQUIRE(getMsgCountsFromMessages(msgs) == expectedSentMsgCounts);
+
+    faabric::util::setMockMode(false);
+    otherWorld.destroy();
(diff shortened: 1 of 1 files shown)


src/scheduler/MpiWorld.cpp
```diff
@@ -1065,18 +1161,8 @@ void MpiWorld::reduce(int sendRank,
     auto rankData = std::make_unique<uint8_t[]>(bufferSize);
 
     if (sendRank == recvRank) {
-        // If we're the receiver of the reduce, await inputs from our local
-        // ranks (besides ourselves) and remote leaders
         SPDLOG_TRACE("MPI - reduce ({}) all -> {}", operation->id, recvRank);
 
-        // Work out the list of all the ranks we need to wait for
```
csegarragonz, line 1072: We no longer need this after the refactor.

src/scheduler/MpiWorld.cpp
```diff
@@ -379,37 +379,32 @@ std::string MpiWorld::getHostForRank(int rank)
...
-    assert(rankHosts.size() == size);
+    assert(hostForRank.size() == size);
 
-    std::vector<int> ranksForHost;
-    for (int i = 0; i < rankHosts.size(); i++) {
-        if (rankHosts.at(i) == host) {
-            ranksForHost.push_back(i);
-        }
+    for (int rank = 0; rank < hostForRank.size(); rank++) {
+        std::string host = hostForRank.at(rank);
```
csegarragonz, line 388: Wrote this in two lines for the sake of clarity, reading `hostForRank` and `ranksForHost` on the same line can be a bit of a tong-twister. Happy to merge to one though.

src/scheduler/MpiWorld.cpp
```diff
@@ -379,37 +379,32 @@ std::string MpiWorld::getHostForRank(int rank)
...
-            localRanks = ranksInHost;
-            localLeader =
-              *std::min_element(ranksInHost.begin(), ranksInHost.end());
-        } else {
-            remoteLeaders.push_back(
-              *std::min_element(ranksInHost.begin(), ranksInHost.end()));
+    for (auto it : ranksForHost) {
+        // Persist the local leader in this host for further use
+        if (it.first == thisHost) {
+            localLeader = *std::min_element(it.second.begin(), it.second.end());
```
csegarragonz, line 403: As mentioned in the PR description, I keep the `localLeader` variable to avoid having to write `ranksForHost[thisHost].front()`, but I could do so.

src/scheduler/MpiWorld.cpp
```diff
@@ -912,53 +918,138 @@ void MpiWorld::gather(int sendRank,
...
-                     recvChunk,
+                     rankData.get(),
                      recvType,
-                     recvCount,
+                     recvCount * it.second.size(),
                      nullptr,
                      faabric::MPIMessage::GATHER);
+
+                // Copy each received chunk to its offset
+                for (int r = 0; r < it.second.size(); r++) {
```
csegarragonz, line 964: This is necessary as we can not assume that ranks for the same host are always consecutive numbers.

The way MPI works atm this only happens if we are overloading the master after running out of resources.

src/scheduler/MpiWorld.cpp
```diff
@@ -912,53 +918,138 @@ void MpiWorld::gather(int sendRank,
...
+        if (getHostForRank(recvRank) != thisHost) {
+            // If we are the local leader (but not the receiver of the gather)
+            // and the receiver is not co-located with us, we gather all our
+            // local rank's data and send it in a single message
+            auto rankData = std::make_unique<uint8_t[]>(
+              ranksForHost[thisHost].size() * sendOffset);
+
+            for (int r = 0; r < ranksForHost[thisHost].size(); r++) {
+                if (ranksForHost[thisHost].at(r) == sendRank) {
+                    // If gathering in-place from a rank that is not-receiving
```
csegarragonz, line 981: Gather has an added factor of grimness because the offset were the data to be sent lives in the send buffer depends:
* If doing an `MPI_Allgather` with `MPI_IN_PLACE`, the send buffer already contains space for the gather result, and the send data is at the offset corresponding to the sending rank.
* In any other situation it is directly at offset zero.

src/scheduler/MpiWorld.cpp
Shillaker: We construct this vector on world creation time don't we? Can we sort it at that point to simplify this code by knowing that ranks are consecutive?

src/scheduler/MpiWorld.cpp
```diff
@@ -912,53 +918,138 @@ void MpiWorld::gather(int sendRank,
 
     bool isInPlace = sendBuffer == recvBuffer;
 
-    // If we're the root, do the gathering
```
Shillaker, line 915: The code is now lacking comments like this that say what it's doing at a high level. At the start of the function you could put a comment summarising the four (?) different scenarios, then make it clearer by factoring out some of the booleans to make the big conditional easier to parse:

```cpp
bool isGatherRoot = sendRank == recvRank;
bool isLocalLeader = sendRank == localLeader;
bool rootOnThisHost = getHostForRank(recvRank) == thisHost;

if(isGatherRoot) {
    // If this is the root rank of the gather, do xyz
    ...
} else if(isLocalLeader && rootOnThisHost) {
    // If this_ rank is local leader and the root rank is on this host too, we do xyz
} else if(isLocalLeader) {
    // If this rank is the local leader and root is on another host we do xyz
} else {
    // If this rank is not local leader or root, do xyz
}
```

src/scheduler/MpiWorld.cpp
```diff
@@ -912,53 +918,138 @@ void MpiWorld::gather(int sendRank,
...
+                 rankData.get(),
+                 sendType,
+                 sendCount * ranksForHost[thisHost].size(),
+                 faabric::MPIMessage::GATHER);
+        } else {
+            // If we are the local leader but are co-located with the receiver
+            // of the gather, we just send our data for gathering.
+            // Note that we, again, make a distinction between sending with
+            // the in place flag or not.
+            if (isInPlace) {
```
Shillaker, line 1019: This code is very nested and with all the branching it's difficult to work out what's going on.

There are a few places it could be simplified where there's two `if`s on consecutive lines, or `else` and `if` on consecutive lines which can be merged, i.e.:

```cpp
if(a) { 

} else {
    if(b) {
    }
}
```

Can be:

```cpp
if(a) {

} else if(b) {

} 
```

src/scheduler/MpiWorld.cpp
```diff
@@ -912,53 +918,138 @@ void MpiWorld::gather(int sendRank,
...
+                     faabric::MPIMessage::GATHER);
             }
         }
     } else {
+        // If we are neither the receiver of the gather, nor a local leader,
+        // we send our data for gathering. We send it to either the receiver
+        // or our local leader, depending on whether we are colocated with
+        // the receiver or not.
+        int realRecvRank =
+          getHostForRank(recvRank) == thisHost ? recvRank : localLeader;
```
Shillaker, line 1041: This branch could be simplified by moving this variable out to the top, e.g.

```cpp
// At the top
int actualRecvRank = rootOnThisHost ? recvRank : localLeader;

// This branch:
else if (isInPlace) {

}
```

Shillaker (changes requested):

src/scheduler/MpiWorld.cpp
csegarragonz: Yes. What I was trying to say here though is that if rank X, local leader of host Y, is sending the contribution of all its local ranks including itself `[X, X1, X2, X3]`, we can't assume `X, X1, X2, X3` to be consecutive numbers.

As a consequence, the receiver of the gather (that has alocated a buffer with `worldSize` slots), can't just copy the four coalesced bits of information starting at `X`s offset.

src/scheduler/MpiWorld.cpp
csegarragonz: Yes, I actually think this was a very good idea and it makes the code much more readable (imo).

src/scheduler/MpiWorld.cpp
```diff
@@ -906,64 +911,156 @@ void MpiWorld::gather(int sendRank,
...
+
+        // Send the locally-gathered data to the receiver rank
+        send(sendRank,
+             recvRank,
+             rankData.get(),
+             sendType,
+             sendCount * ranksForHost[thisHost].size(),
+             faabric::MPIMessage::GATHER);
+
+    } else if (isLocalLeader && isLocalGather) {
```
csegarragonz, line 1030: Scenario 3 and Scenario 5 are, after the re-factor, identical. However, they are, semantically, different situations.

I have decided to keep both as different branches for readability, and to help follow the code using the comment at the beginning.

Shillaker (approved):

PR #201, opened 2021-12-20
Merge regions overhaul

Shillaker:
This PR addresses the following problems:

- Merge regions were not being transferred across hosts (see below).
- Whitelisting merge regions for big applications became unfeasible (impossible to specify all places where the memory might change).
- Merge region code only supported `Raw` and `Int` types (needed more types including `Bool`, `Long`, `Float` and `Double`).

When it comes to transferring merge regions across hosts, we have to consider three scenarios where snapshot data is being transferred:

1. First time executing a thread on a remote host, in which case the full snapshot is pushed.
2. Subsequent executions of threads on a remote host (in which case just an update to the snapshot is sent)
3. Remote hosts returning snapshot diffs after executing a thread.

In the first two, we are not currently including merge regions, which means, if the main thread on the main host sets up some merge regions, they are not respected by remote hosts. 

Changes:

- Distinguish between a snapshot update (2) and simple snapshot diffs (3).
- Include snapshot merge regions in (1) and (2).
- Reintroduce `Ignore` merge regions to blacklist regions of memory (e.g. guard pages), then automatically fill the rest of the snapshot memory with `Overwrite` regions.
- Add double, float, bool and long support.


src/flat/faabric.fbs
```diff
@@ -1,24 +1,33 @@
...
   key:string;
 }
 
-table SnapshotDiffChunk {
+table SnapshotDiffRequest {
   offset:int;
-  dataType:int;
-  mergeOp:int;
+  data_type:int;
+  merge_op:int;
```
Shillaker, line 22: `SnapshotDiff` and `SnapshotMergeRegion` names clash with those in `faabric::util`. I've appended `Request` as this fits with the style of the rest of the FB objects. 

I also switched these fields to all use snake case as FB prints a warning if you do anything else.

Shillaker:
@kubasz I added you as a reviewer given that you looked at the previous snapshot stuff, but feel free to ignore if you don't have scope to keep checking this stuff 😄

kubasz:
I don't have much time to review this, but from a quick glance I can offer two quick comments:
 * `const std::vector<T>&` -> `std::span<const T>` in arguments
 * when pushing/emplacing back in a loop, make sure there's a reserve before the loop with the number of iterations

Shillaker:
Ok nice, cheers and have a good Xmas!

src/util/snapshot.cpp
```diff
@@ -453,19 +410,84 @@ std::string snapshotMergeOpStr(SnapshotMergeOperation op)
     }
 }
 
+void SnapshotMergeRegion::addOverwriteDiff(
+  std::vector<SnapshotDiff>& diffs,
+  std::span<const uint8_t> original,
+  const uint8_t* updated,
+  std::pair<uint32_t, uint32_t> dirtyRange)
```
Shillaker, line 417: The logic for this region is copied from above and unchanged.

tests/test/snapshot/test_snapshot_diffs.cpp
```diff
@@ -36,9 +36,13 @@ TEST_CASE_METHOD(SnapshotTestFixture,
 
     int sharedMemPages = 8;
     size_t sharedMemSize = sharedMemPages * HOST_PAGE_SIZE;
-    MemoryRegion sharedMem = allocateSharedMemory(sharedMemSize);
+    MemoryRegion sharedMem = allocatePrivateMemory(sharedMemSize);
```
csegarragonz, line 39: Just to double check, the variable is called `sharedMem` and is defined by calling `allocatePrivateMemory` which feels odd.

tests/test/util/test_memory.cpp
```diff
@@ -433,7 +433,7 @@ TEST_CASE("Test mapping memory", "[util]")
 TEST_CASE("Test mapping memory fails with invalid fd", "[util]")
 {
     size_t memSize = 10 * HOST_PAGE_SIZE;
-    MemoryRegion sharedMem = allocateSharedMemory(memSize);
+    MemoryRegion sharedMem = allocatePrivateMemory(memSize);
```
csegarragonz, line 436: Same point re. variable wording.

tests/test/util/test_snapshot.cpp
```diff
@@ -179,12 +179,12 @@ TEST_CASE_METHOD(SnapshotMergeTestFixture,
     std::vector<uint8_t> expectedSnapMem = snap->getDataCopy();
 
     // Set up two shared mem regions
-    MemoryRegion sharedMemA = allocateSharedMemory(snapSize);
-    MemoryRegion sharedMemB = allocateSharedMemory(snapSize);
+    MemoryRegion sharedMemA = allocatePrivateMemory(snapSize);
```
csegarragonz, line 182: Same point re. variable naming throughout this test.

tests/test/util/test_snapshot.cpp
```diff
@@ -590,6 +592,214 @@ TEST_CASE_METHOD(SnapshotMergeTestFixture,
         expectedData = faabric::util::valueToBytes<int>(diffValue);
     }
 
+    SECTION("Long")
```
csegarragonz, line 595: Nice nesting of tests here :+1:

csegarragonz (approved):
LGTM. Just pointed out two very minor issues, feel free to ignore them.

I have deliberately marked all occurings I have found to aid in the re-factor if you end up going for it.

Shillaker:
I'll merge this in the interest of moving things along and fixing downstream, then I'll put the refactoring into https://github.com/faasm/faabric/pull/210

PR #206, opened 2021-12-27
Add in-proc PAIR sockets

Shillaker:
Wrapper around [0MQ PAIR sockets](https://zeromq.org/socket-api/#exclusive-pair-pattern)

### DIFF
--- a/tests/test/transport/test_message_endpoint_client.cpp
+++ b/tests/test/transport/test_message_endpoint_client.cpp
@@ -232,7 +232,7 @@ TEST_CASE_METHOD(SchedulerTestFixture, "Test direct messaging", "[transport]")
 
     std::string inprocLabel = "direct-test";
 
-    AsyncDirectSendEndpoint sender(inprocLabel, TEST_PORT);
+    AsyncDirectSendEndpoint sender(inprocLabel);
     sender.send(msg, expected.size());
 
     AsyncDirectRecvEndpoint receiver(inprocLabel);
@@ -276,7 +276,8 @@ TEST_CASE_METHOD(SchedulerTestFixture,
                 SLEEP_MS(10);
             }
 
-            // Make main thread wait until messages are queued
+            // Make main thread wait until messages are queued (to check no
+            // issue with connecting before binding)
             if (i == 10) {
                 startLatch->wait();
             }
--- a/tests/test/transport/test_message_endpoint_client.cpp
+++ b/tests/test/transport/test_message_endpoint_client.cpp
@@ -1,6 +1,7 @@
 #include "faabric_utils.h"
 #include <catch2/catch.hpp>
 
+#include <atomic>
 #include <thread>
 #include <unistd.h>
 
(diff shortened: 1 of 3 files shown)


tests/test/transport/test_message_endpoint_client.cpp
```diff
@@ -224,6 +225,83 @@ TEST_CASE_METHOD(SchedulerTestFixture,
...
 
-#endif
+TEST_CASE_METHOD(SchedulerTestFixture, "Test direct messaging", "[transport]")
+{
+    std::string expected = "Direct hello";
+    const uint8_t* msg = BYTES_CONST(expected.c_str());
+
+    std::string inprocLabel = "direct-test";
+
+    AsyncDirectSendEndpoint sender(inprocLabel, TEST_PORT);
```
csegarragonz, line 235: Afaict `TEST_PORT` here is passed as the `timeoutMs` argument.

tests/test/transport/test_message_endpoint_client.cpp
```diff
@@ -224,6 +225,83 @@ TEST_CASE_METHOD(SchedulerTestFixture,
...
+            std::string expected = "Direct hello " + std::to_string(i);
+            const uint8_t* msg = BYTES_CONST(expected.c_str());
+            sender.send(msg, expected.size());
+
+            if (i % 100 == 0) {
+                SLEEP_MS(10);
+            }
+
+            // Make main thread wait until messages are queued
+            if (i == 10) {
```
csegarragonz, line 280: Why do we need to wait for messages to be queued? Does this mean the receiver thread won't block waiting for messages to be in place?

tests/test/transport/test_message_endpoint_client.cpp
```diff
@@ -224,6 +225,83 @@ TEST_CASE_METHOD(SchedulerTestFixture,
...
+    SECTION("Recv no size")
+    {
+        faabric::transport::Message recvMsg = receiver.recv().value();
+        actual = std::string(recvMsg.data(), recvMsg.size());
+    }
+
+    REQUIRE(actual == expected);
+}
+
+TEST_CASE_METHOD(SchedulerTestFixture,
```
csegarragonz, line 257: For my particular use case, I'd be interested in seeing a stress test with various threads producing and various threads consuming to different in proc labels (each producer matched to one consumer).

Feel free to ignore, but I think it could be a relevant stress test.

csegarragonz (changes requested):

tests/test/transport/test_message_endpoint_client.cpp
Shillaker: Yes, good spot.

tests/test/transport/test_message_endpoint_client.cpp
Shillaker: The receiver thread will block, it's just that under the hood the receiver is doing the bind and the sender is doing a connect. I always like to check that doing a connect before a bind works ok (as in some instances it doesn't IIRC).

tests/test/transport/test_message_endpoint_client.cpp
Shillaker: Yes good idea, have updated.

csegarragonz (approved):

PR #207, opened 2021-12-27
Use PAIR sockets for MPI

csegarragonz:


csegarragonz:
Discarded as serialisation costs make the approach not scalable.

PR #209, opened 2021-12-28
Queue experiment

Shillaker:


Shillaker:
@csegarragonz this is a start on profiling the existing queue implementation. Can discuss offline.

PR #210, opened 2021-12-29
Dirty tracking performance improvements

Shillaker:
This PR is the third (and final) in a series of overhauls of the threading, snapshotting and dirty tracking model.

This PR addresses the underlying issue of the performance of the dirty tracking, which is currently based on soft dirty PTEs. While it works, it has a couple of shortcomings:

- Soft dirty PTEs can only be reset globally, creating a synchronisation point and preventing us from performing dirty tracking on more than one application at a time on each host.
- Resetting and reading soft dirty PTEs requires writing to `/proc/self/clear_refs` and reading from `/proc/self/pagemap` respectively, which can be a drain on performance when done repeatedly in a tight loop.

Ideally we would do this tracking with [`userfaultfd` write-protected pages](https://www.kernel.org/doc/html/latest/admin-guide/mm/userfaultfd.html), however, this is only available in kernels 5.7+, i.e. Ubuntu 22+ which isn't yet stable. In the interim we can use an `mprotect` + `SIGSEGV` approach, where we make pages read-only using `mprotect`, then catch the resulting segfault when they're written to, and mark them as dirty. 

Although in our current use-cases the PTEs are less performant that the segfault approach, this may not be the case for all workloads, so I'd like to keep the soft PTEs approach around.

PTEs, segfaults and `userfaultfd` all result in a different approach to aggregating diffs for a batch of threads. PTEs are system-wide, while segfaults are handled by individual threads (so we have to introduce thread-local tracking), and `userfaultfd` would use a single background tracker thread per application. To abstract the boilerplate for doing this and support switching more easily, I've introduced a single `DirtyTracker` class that abstracts all the details, and a configuration parameter `DIRTY_TRACKING_MODE` which can be set to `softpte` or `segfault` for now, and `uffd` in future.

This results in quite a few changes to the code:

- Abstract all dirty tracking logic to an interface encapsulated in the `DirtyTracker` class. 
- Support both global and thread-local dirty tracking.
- Move all logic related to handling thread snapshots, invoking threads and dirty tracking into the `Executor` class (previously this was scattered across Faabric and Faasm).
- Introduced the concept of a "main thread snapshot", used by threaded applications and controlled by Faabric. Previously this snapshot had to be created and updated outside of Faabric, which was confusing and error-prone. 
- When repeatedly executing batches of threads from the same application, cache each scheduling decision and pass it as a hint to the next batch of the same size.
- Stop using dirty page tracking to track changes made to snapshots. We can do this manually with higher precision as all changes to each snapshot now go through methods on the `SnapshotData` class.
- Improve the performance of inner loops related to dirty tracking and diffing.


src/proto/faabric.proto
```diff
@@ -130,40 +130,41 @@ message Message {
     int64 finishTimestamp = 19;
 
     bool isAsync = 20;
-    bool isPython = 21;
-    bool isStatusRequest = 22;
+    bool isStatusRequest = 21;
+    bool isThreaded = 22;
```
Shillaker, line 134: I'm loathe to add another `isXYZ` flag, but I'll do another PR to move all the feature flags on a message into a `features` dictionary, i.e.

```
{
    "user": "foo",
    "function": "my_func",
    "features": {
          "threaded": true,
          "sgx": true,
          "exec_graph": true
    }
```

dist-test/dev_server.sh
```diff
@@ -13,12 +13,14 @@ elif [[ "$1" == "restart" ]]; then
     docker-compose restart dist-test-server
 elif [[ "$1" == "stop" ]]; then
     docker-compose stop dist-test-server
+elif [[ "$1" == "rm" ]]; then
+    docker-compose rm dist-test-server
```
Shillaker, line 17: Added `rm` for convenience.

include/faabric/util/dirty.h
```diff
@@ -0,0 +1,115 @@
+#pragma once
```
Shillaker, line 1: `dirty.h` and `dirty.cpp` contain all the boilerplate code around dirty page tracking using both soft PTEs and segfaults.

src/util/logging.cpp
```diff
@@ -6,7 +6,7 @@ namespace faabric::util {
 void initLogging()
 {
     // Docs: https://github.com/gabime/spdlog/wiki/3.-Custom-formatting
-    spdlog::set_pattern("[%H:%M:%S:%e] [%t] [%l] %v");
+    spdlog::set_pattern("%^[%H:%M:%S] [%t] [%L]%$ %v");
```
Shillaker, line 9: This adds colour based on log level back into our logging and makes the format more compact. Logs were quite hard to read before.

include/faabric/util/bytes.h
```diff
@@ -25,6 +27,20 @@ int safeCopyToBuffer(const uint8_t* dataIn,
...
+ * the two arrays.
+ */
+std::vector<std::pair<uint32_t, uint32_t>> diffArrayRegions(
+  std::span<const uint8_t> a,
+  std::span<const uint8_t> b);
+
+/*
+ * Returns a list of flags marking which bytes differ between the two arrays.
+ */
+std::vector<bool> diffArrays(std::span<const uint8_t> a,
```
csegarragonz, line 41: Isn't this a bit-wise XOR?

src/scheduler/Executor.cpp
```diff
@@ -99,6 +116,189 @@ void Executor::finish()
...
+    for (const auto& mr : mergeRegions) {
+        snap->addMergeRegion(
+          mr.offset, mr.length, mr.dataType, mr.operation, true);
+    }
+
+    // TODO - here the main thread will wait, so technically frees up a slot
+    // that could be used.
+    std::string cacheKey =
+      std::to_string(msg.appid()) + "_" + std::to_string(req->messages_size());
+    bool hasCachedDecision = false;
```
csegarragonz, line 196: Could we make this a `std::atomic<bool>`?

src/scheduler/Executor.cpp
```diff
@@ -99,6 +116,189 @@ void Executor::finish()
...
+    {
+        faabric::util::SharedLock lock(threadExecutionMutex);
+        hasCachedDecision =
+          cachedDecisionHosts.find(cacheKey) != cachedDecisionHosts.end();
+    }
+
+    if (!hasCachedDecision) {
```
csegarragonz, line 203: And here use:
```cpp
hasCachedDecision.load(std::memory_order_acquire)
```

src/scheduler/Executor.cpp
```diff
@@ -99,6 +116,189 @@ void Executor::finish()
...
+            // Cache the decision for next time
+            SPDLOG_DEBUG(
+              "No cached decision for {} x {}/{}, caching group {}, hosts: {}",
+              req->messages().size(),
+              msg.user(),
+              msg.function(),
+              groupId,
+              faabric::util::vectorToString<std::string>(decision.hosts));
+
+            cachedGroupIds[cacheKey] = groupId;
```
csegarragonz, line 225: I would beneffit from havng the decision cached in the scheduler, not in the executor. I am not asking you to change this, but writing this as a reminder to add a note in my PR that the logic is duplicated.

src/scheduler/Executor.cpp
```diff
@@ -201,11 +411,58 @@ void Executor::executeTasks(std::vector<int> msgIdxs,
     }
 }
 
+std::shared_ptr<faabric::util::SnapshotData> Executor::getMainThreadSnapshot(
+  faabric::Message& msg,
+  bool createIfNotExists)
+{
+    std::string snapshotKey = faabric::util::getMainThreadSnapshotKey(msg);
+    bool exists = false;
```
csegarragonz, line 419: Again, make `exists` a `std::atomic<bool>` and access it using `exists.load(std::memory_order_acquire)`

src/scheduler/Executor.cpp
```diff
@@ -201,11 +411,58 @@ void Executor::executeTasks(std::vector<int> msgIdxs,
...
+    {
+        faabric::util::SharedLock lock(threadExecutionMutex);
+        exists = reg.snapshotExists(snapshotKey);
+    }
+
+    if (!exists && createIfNotExists) {
```
csegarragonz, line 425: Exists access here should use `load`

src/scheduler/Executor.cpp
```diff
@@ -201,11 +411,58 @@ void Executor::executeTasks(std::vector<int> msgIdxs,
...
+                         snapshotKey,
+                         faabric::util::funcToString(msg, false));
+
+            std::shared_ptr<faabric::util::SnapshotData> snap =
+              std::make_shared<faabric::util::SnapshotData>(getMemoryView());
+            reg.registerSnapshot(snapshotKey, snap);
+        } else {
+            return reg.getSnapshot(snapshotKey);
+        }
+    } else if (!exists) {
```
csegarragonz, line 438: Same, access with `exists.load(std::memory_order_acquire)`

src/util/dirty.cpp
```diff
@@ -0,0 +1,382 @@
...
+        SPDLOG_ERROR("Could not read pagemap ({} != {})", nRead, nPages);
+        throw std::runtime_error("Could not read pagemap");
+    }
+
+    // Iterate through the pagemap entries to work out which are dirty
+    std::vector<std::pair<uint32_t, uint32_t>> regions;
+    std::vector<int> dirtyPages;
+    bool isRegionInProgress = false;
+    int pageStart = 0;
+    for (int i = 0; i < nPages; i++) {
```
csegarragonz, line 137: Isn't this very simillar to `faabric::util::getDirtyRegions(...)`?

src/util/dirty.cpp
```diff
@@ -0,0 +1,382 @@
...
+    {
+        PROF_START(GetDirtyRegions)
+        std::vector<std::pair<uint32_t, uint32_t>> dirty;
+        if (regionBase == nullptr) {
+            return dirty;
+        }
+
+        uint32_t diffPageStart = 0;
+        bool diffInProgress = false;
+        for (int i = 0; i < nPages; i++) {
```
csegarragonz, line 209: Same, isn't this essentially the same as `faabric::util::getDiffRegions()`?

csegarragonz (approved):
LGTM, very minor changes but will approve so that you don't have to re-request a review.

src/util/bytes.cpp
```diff
@@ -90,4 +93,57 @@ std::string formatByteArrayToIntString(const std::vector<uint8_t>& bytes)
...
+{
+    if (a.size() != b.size()) {
+        SPDLOG_ERROR(
+          "Cannot diff arrays of different sizes {} != {}", a.size(), b.size());
+        throw std::runtime_error("Cannot diff arrays of different sizes");
+    }
+
+    std::vector<bool> diffs(a.size(), false);
+    for (int i = 0; i < a.size(); i++) {
+        diffs[i] = a.data()[i] != b.data()[i];
```
kubasz, line 143: This might be *very* slow due to the `vector<bool>` specialization, just swapping the type to uint8_t (at the cost of 8x more memory) makes it 11x faster: <https://quick-bench.com/q/d8INA76hIDM3m4gfC4xJKI4HcQM>

src/util/dirty.cpp
```diff
@@ -0,0 +1,382 @@
...
+{
+    // Do nothing - need to preserve thread-local data for getting dirty regions
+    SPDLOG_TRACE("Stopping thread-local tracking on region size {}",
+                 region.size());
+}
+
+void SegfaultDirtyTracker::reinitialise()
+{
+    if (faabric::util::isTestMode()) {
+        // This is a hack because catch changes the segfault signal handler
```
kubasz, line 359: This is easily configurable with catch's defines: <https://github.com/catchorg/Catch2/blob/devel/docs/configuration.md#other-toggles>, specifically `CATCH_CONFIG_NO_POSIX_SIGNALS` can be defined

kubasz (commented):
Just had a quick look at this, because it sounds very similar to what I'm doing to reduce the size of incremental snapshots for offloading in my project. I'd suggest not creating byte-wise diffs, as that's pretty costly performance-wise, you can play around with variants in the quickbench link I put in a comment. The code in `faabric/util/delta` uses a configurable "page" size (any difference in a page = emit a diff for the whole page), in my testing around 64 yielded the best performance without making the diffs larger than a few percent - you might want to test this for your applications. The code could be relatively easily modified to take in an array of modified OS pages to skip known-unmodified pages, I'm happy to do this if you're interested as it's on my todo list anyway, and then you could use the generate/applyDelta functions

src/scheduler/Executor.cpp
kubasz: This (and the other variables you mentioned) is a stack-local, not shared across threads so atomics don't really make sense here. You only need atomic accesses if the same variable (in the same memory location) is accessed from multiple threads without being in a mutex critical section.

src/util/dirty.cpp
Shillaker: Ah awesome, did not know about that.

include/faabric/util/bytes.h
Shillaker: It's currently byte-wise, not bit-wise, but this may change. I noticed that this function isn't actually used, so I should probably just delete it.

Shillaker:
I'm going to merge this and take discussions of improvements and tweaks offline.

PR #213, opened 2022-01-06
Cmake: quote conan revisions

### LINKED ISSUE #212: Conan revisions in cmakelists must be quoted
`#` starts a line comment in CMake unless in quotes, so:
```cmake
# was:
cppzmq/4.8.1@#e0f26b0614b3d812815edc102ce0d881
# must become:
"cppzmq/4.8.1@#e0f26b0614b3d812815edc102ce0d881"
```

csegarragonz:
After merging this in one has to nuke its local `/root/.conan` as all the revisions (which are supposed not to change) were off.

The problem now is that all the stored conan cache pin to the wrong revisions, so will probably have to delete the cache in GHA, which I will do once there is less movement in faabric.

Closes #212


kubasz:
You can add this <https://github.com/auto-ndp/faabric/commit/ad20a14a9f19d22b6797eba4ef52ecce981d688b> to avoid having to manually nuke the conan cache

Shillaker (approved):

PR #214, opened 2022-01-07
Periodically Check For Migration Opportunities

csegarragonz:
In this PR I introduce the ability to migrate detect migration opportunities, and migrate MPI functions.

Function migration in faabric is divided in two independent processes:

**Detection of migration opportunities:**
* A function can opt in to be migrated by setting a positive value in a message field.
* The first time the scheduler schedules a function with said parameter, it starts a background thread that checks for migration opportunities (only the master host for an app. checks for migration opportunities for it).
* The type of migration opportunity to be looked for is specified with a `MigrationStrategy` but currently only `MigrationStrategy::BIN_PACK` is supported: we try to pack functions from hosts with fewer to hosts with more. In-flight apps willing to be checked for migration opportunities, and the opportunities found are stored in shared data structures.
* Lastly, when a migration opportunity is found, the scheduler notifies all hosts running functions for the app that can be migrated.
* The combination of synchronous cross-host function calls and locks in the scheduler guarantees that once the master host releases write lock on the migrations data structure, all other hosts' schedulers are already aware of it.

**Performing the actual migration:**
* The actual migration is handled by faasm (faasm/faasm#565).
* The only think we need to do from faabric is to prepare for the migration. Preparation is ad-hoc to the type of app being migrated, we implement `MpiWorld::prepareMigration`.

**Known issues:**
* At the moment, we only remove a `pendingMigration` when execution finishes. Even if we migrate. It is not clear when is it safe to remove a migration after migrating.


src/scheduler/Scheduler.cpp
```diff
@@ -1129,4 +1162,145 @@ ExecGraphNode Scheduler::getFunctionExecGraphNode(unsigned int messageId)
...
+                continue;
+            }
+
+            faabric::PendingMigrations msg;
+            msg.set_appid(originalDecision.appId);
+            // TODO - generate a new groupId here for processes to wait on
+            // during the migration? msg.set_groupid();
+
+            if (migrationStrategy ==
+                faabric::util::MigrationStrategy::BIN_PACK) {
```
csegarragonz, line 1206: Note that this is quite similar to the `makeSchedulingDecision` code but with some differences that made it preferrable to be implemented as a separate bit of code:
* First, we are not actually scheduling new functions, we are just checking if a particular kind of re-arrangement is possible.
* Second, we don't enforce the changes, therefore we actually don't have to persist them (in the `HostResources`).

include/faabric/util/message.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <faabric/proto/faabric.pb.h>
+
+namespace faabric::util {
+void copyMessage(const faabric::Message* src, faabric::Message* dst);
```
csegarragonz, line 6: I needed a method to deep-copy a `faabric::Message` to set the message in the `faabric::PendingMigrations` from `faabric::BatchExecuteRequest`.

include/faabric/scheduler/FunctionCallApi.h
```diff
@@ -7,6 +7,7 @@ enum FunctionCalls
     ExecuteFunctions = 1,
     Flush = 2,
     Unregister = 3,
-    GetResources = 4
+    GetResources = 4,
+    AddPendingMigration = 5
```
csegarragonz, line 11: This function call is used from the master host of an application to let know other hosts running functions of the same application that a migration opportunity has been found.

src/scheduler/Scheduler.cpp
```diff
@@ -1129,4 +1164,191 @@ ExecGraphNode Scheduler::getFunctionExecGraphNode(unsigned int messageId)
...
+void Scheduler::removePendingMigration(uint32_t appId)
+{
+    faabric::util::FullLock lock(mx);
+
+    inFlightRequests.erase(appId);
+    pendingMigrations.erase(appId);
+}
+
+std::vector<std::shared_ptr<faabric::PendingMigrations>>
+Scheduler::doCheckForMigrationOpportunities(
```
csegarragonz, line 1250: The need for a new method other than re-using `makeSchedulingDecisions`  arises from the fact that:
* In this method we just check for the ability to migrate, but do not actually change the underlying host resources.
* The scheduling and migration policies can differ. The former being more complex and the latter simpler.

src/scheduler/Scheduler.cpp
```diff
@@ -1129,4 +1197,191 @@ ExecGraphNode Scheduler::getFunctionExecGraphNode(unsigned int messageId)
...
+            SPDLOG_TRACE("Skipping app {} as migration opportunity has "
+                         "already been recorded",
+                         originalDecision.appId);
+            continue;
+        }
+
+        faabric::PendingMigrations msg;
+        msg.set_appid(originalDecision.appId);
+
+        if (migrationStrategy == faabric::util::MigrationStrategy::BIN_PACK) {
```
csegarragonz, line 1307: I haven't implemented the `EMPTY_HOSTS` migration strategy, so I haven't implemented support an env. variable to set the migration strategy.

include/faabric/scheduler/FunctionCallApi.h
Shillaker: This feels like a code comment, perhaps on `sendAddPendingMigration`?

include/faabric/scheduler/FunctionCallClient.h
```diff
@@ -42,6 +45,9 @@ class FunctionCallClient : public faabric::transport::MessageEndpointClient
 
     faabric::HostResources getResources();
 
+    void sendAddPendingMigration(
+      std::shared_ptr<faabric::PendingMigrations> req);
```
Shillaker, line 49: Is the `Add` necessary in here? I think `sendPendingMigration` conveys the meaning and is easier to read.

include/faabric/scheduler/FunctionMigrationThread.h
```diff
@@ -0,0 +1,26 @@
...
+#include <mutex>
+#include <thread>
+
+namespace faabric::scheduler {
+class FunctionMigrationThread
+{
+  public:
+    // Start a background thread that, every wake up period, will check if there
+    // are migration opportunities for in-flight apps that have opted in to
+    // being checked for migrations.
```
Shillaker, line 13: This is really a description of the whole class, I'd put the comment on the `class`.

include/faabric/util/message.h
Shillaker: Can't you copy messages using assignment i.e. `src = dst`? What's not copied in that case? If it's still needed, I'd put a comment on this method saying why and what's necessary to copy manually. As per comment below, a function like this will rot as soon as someone changes the `Message` and forgets to update it.

src/scheduler/FunctionCallClient.cpp
```diff
@@ -128,6 +140,22 @@ faabric::HostResources FunctionCallClient::getResources()
...
+  std::shared_ptr<PendingMigrations> req)
+{
+    faabric::PendingMigrations request;
+    faabric::EmptyResponse response;
+
+    if (faabric::util::isMockMode()) {
+        faabric::util::UniqueLock lock(mockMutex);
+        addPendingMigrationRequests.emplace_back(host, req);
+    } else {
+        syncSend(faabric::scheduler::FunctionCalls::AddPendingMigration,
```
Shillaker, line 153: Does this need to be synchronous? I think the sender will eventually wait on responses from the migrated `Executors` so could be async.

src/proto/faabric.proto
```diff
@@ -242,3 +245,20 @@ message PointToPointMappings {
...
+// ---------------------------------------------
+// FUNCTION MIGRATIONS
+// ---------------------------------------------
+
+message PendingMigrations {
+    int32 appId = 1;
+    int32 groupId = 2;
+
+    message PendingMigration {
+        Message msg = 1;
```
Shillaker, line 258: AFAICT this is the message that's sent to worker hosts to tell it to migrate a function across hosts. If that's right, doesn't that host already have the messge, so we just need the message ID here?

include/faabric/scheduler/Scheduler.h
```diff
@@ -220,6 +230,18 @@ class Scheduler
 
     ExecGraph getFunctionExecGraph(unsigned int msgId);
 
+    // ----------------------------------
+    // Function Migration
+    // ----------------------------------
+    void checkForMigrationOpportunities();
+
+    std::shared_ptr<faabric::PendingMigrations> canAppBeMigrated(
```
Shillaker, line 238: Seems like this method is really `getPendingAppMigrations`.

src/scheduler/MpiWorld.cpp
```diff
@@ -1732,4 +1759,85 @@ void MpiWorld::checkRanksRange(int sendRank, int recvRank)
...
+    }
+
+    prepareMigration(thisRank, pendingMigrations);
+    if (faabric::util::isMockMode()) {
+        // When mocking in the tests, the call to get the current executing
+        // executor may fail
+        faabric::util::UniqueLock lock(mockMutex);
+        mpiMockedPendingMigrations.push_back(pendingMigrations);
+    } else {
+        // An executing thread may never return from the next call. However,
```
Shillaker, line 1781: > An executing thread may never return from the next call.

This seems like very strange behaviour and I'm not sure how this would be implemented. I would instead have `tryMigrate` return a `bool`, then if it's `true` it means it's been migrated and the caller (e.g. the MPI function) has to deal with it and ensure a graceful shutdown. 

This would probably best be done by throwing an `ExecutorMigrated` exception, which the main `Executor` loop could catch and self-terminate accordingly (as it does already with exceptions [here](https://github.com/faasm/faabric/blob/master/src/scheduler/Executor.cpp#L531)).

src/util/message.cpp
```diff
@@ -0,0 +1,63 @@
+#include <faabric/util/message.h>
+
+namespace faabric::util {
+void copyMessage(const faabric::Message* src, faabric::Message* dst)
+{
+    dst->set_id(src->id());
```
Shillaker, line 6: This function will rot within the next week or so, see my comment above re. using assignment.

Shillaker (changes requested):

src/util/message.cpp
kubasz: The copy constructor of protobuf objects also does a deep copy already, it's even used in a couple of places in this codebase

include/faabric/scheduler/FunctionMigrationThread.h
```diff
@@ -0,0 +1,26 @@
...
+    void start(int wakeUpPeriodSecondsIn);
+
+    void stop();
+
+    int wakeUpPeriodSeconds;
+
+  private:
+    std::unique_ptr<std::thread> workThread = nullptr;
+    std::mutex mx;
+    std::condition_variable mustStopCv;
```
kubasz, line 23: Now that we have moved to c++20, new code could use `std::jthread` with a stop_token instead of this relatively complex pattern, see:
```c++
#include <thread>
#include <iostream>
 
using namespace std::literals::chrono_literals;
 
void f(std::stop_token stop_token, int value)
{
    while (!stop_token.stop_requested()) {
        std::cout << value++ << ' ' << std::flush;
        std::this_thread::sleep_for(200ms);
    }
    std::cout << std::endl;
}
 
int main()
{
    std::jthread thread(f, 5); // prints 5 6 7 8... for approximately 3 seconds
    std::this_thread::sleep_for(3s);
    // The destructor of jthread calls request_stop() and join().
}
```
from <https://en.cppreference.com/w/cpp/thread/stop_token>

src/scheduler/FunctionCallClient.cpp
csegarragonz: The reason I made this synchronous was to guarantee that all executors see the pending migration at the same migration point. The problem I would want to guard against is:
1. Master host detects migration opportunity.
2. Executor hits `tryMigrate()` in the master host and it sees there is a pending migration.
3. Executor hits `tryMigrate()` in worker host but does not see a pending migration.
4. Worker host updates its migration map.
Note that 1 and 4 take a write lock on the scheduler, whereas 2 and 3 take a read one.

By making it synchronous we force 1 and 4 to happen atomically (with respect to 2 and 3).

include/faabric/scheduler/FunctionCallClient.h
csegarragonz: Agree, also the use of the plural and singular form of migration(s) was confusing. I have renamed the function calls to `sendPendingMigrations` and `recvPendingMigrations`.

include/faabric/scheduler/FunctionMigrationThread.h
Shillaker: Nice, didn't know about this. 

Rather than have new and old code diverge in their styles, this should be straightforward to do wholesale in an isolated PR. Have created an issue: https://github.com/faasm/faabric/issues/216.

Shillaker: (I realise I'm letting the great be the enemy of the good here, but we're in a rush at the moment and as this is already working I'm reluctant to change it, even though the result would be simpler and cleaner)

src/proto/faabric.proto
csegarragonz: All functions in the app will read the same `PendingMigrations` message. They:
* Iterate over the `PendingMigration` repeated field (`migrations`).
* If the message corresponds to their message id, they themselves must trigger their own migration.
* To do so, they need to know where they are headed `dstHost`.

Maybe the `srcHost` field is not strictly needed, but it can be asserted/makes printing and logging easier.

src/scheduler/MpiWorld.cpp
```diff
@@ -1554,12 +1568,10 @@ std::shared_ptr<InMemoryMpiQueue> MpiWorld::getLocalQueue(int sendRank,
 // Note - the queues themselves perform concurrency control
 void MpiWorld::initLocalQueues()
 {
-    // Assert we only allocate queues once
-    assert(localQueues.size() == 0);
```
csegarragonz, line 1558: I remove this assertion as now this method is also called at the end of `prepareMigration` to update the `localQueues` with the new ranks to hosts mapping.

src/scheduler/MpiWorldRegistry.cpp
```diff
@@ -52,9 +52,7 @@ MpiWorld& MpiWorldRegistry::getOrInitialiseWorld(faabric::Message& msg)
     {
         faabric::util::SharedLock lock(registryMutex);
         MpiWorld& world = worldMap[worldId];
-        if (msg.recordexecgraph()) {
-            world.setMsgForRank(msg);
-        }
+        world.setMsgForRank(msg);
```
csegarragonz, line 55: We also need the message when migrating.

src/scheduler/Scheduler.cpp
```diff
@@ -303,6 +309,9 @@ faabric::util::SchedulingDecision Scheduler::makeSchedulingDecision(
 
         // Work out how many we can handle locally
         int slots = thisHostResources.slots();
+        if (topologyHint == faabric::util::SchedulingTopologyHint::UNDERFULL) {
```
csegarragonz, line 312: This is all the logic required to implement the `UNDERFULL` topology hint. I could move it to a separate PR but it is literally just this and the tests.

src/scheduler/Scheduler.cpp
```diff
@@ -886,6 +960,12 @@ void Scheduler::setFunctionResult(faabric::Message& msg)
         throw std::runtime_error("Result key empty. Cannot publish result");
     }
 
+    // Remove the app from in-flight map if still there, and this host is the
+    // master host for the message
+    if (msg.masterhost() == thisHost) {
+        removePendingMigration(msg.appid());
```
csegarragonz, line 966: As mentioned in the PR description, this is the only time we remove a pending migration.

In particular, we don't remove it after we do a migration. It is not clear when it is safe to do that.

The main reason for the complexity to check is that, at the moment, we don't hit a faabric-based barrier after migrating. The PoC for a migration app is designed in a way that functions hit an application-based barrier after migrating.

include/faabric/scheduler/Scheduler.h
```diff
@@ -220,6 +235,27 @@ class Scheduler
...
+
+    void removePendingMigration(uint32_t appId);
+
+    // ----------------------------------
+    // Clients
+    // ----------------------------------
+    faabric::scheduler::FunctionCallClient& getFunctionCallClient(
+      const std::string& otherHost);
+
+    faabric::snapshot::SnapshotClient& getSnapshotClient(
```
csegarragonz, line 256: We need the snapshot client to be a public member in faasm/faasm#565 before migrating a function. In that case we are doing a function chain between a function in a non-master host, and a function in the master host. Faabric won't register,take, and push snapshots in this case; we have to do it ourselves manually. Thus, need access to the snapshot client.

We don't need the function call client to be a public member, but I was reluctant to split the declaration of the clients (happy to revert).

include/faabric/scheduler/FunctionMigrationThread.h
Shillaker: Let's try and capture as many instances of background thread logic as possible in a separate PR (including the bg threads in `FaabricMain`).

src/scheduler/MpiWorld.cpp
```diff
@@ -222,20 +222,31 @@ void MpiWorld::create(faabric::Message& call, int newId, int newSize)
...
     }
 
     std::vector<std::string> executedAt;
     if (size > 1) {
-        // Send the init messages (note that message i corresponds to rank i+1)
-        // By default, we use the NEVER_ALONE policy for MPI executions to
-        // minimise cross-host messaging
+        // 10/01/22 - We add this check to force different scheduling behaviour
+        // if running migration experiments. The check can be deleted
+        // afterwards.
```
Shillaker, line 243: Can we remove this now and have a way to programmatically do it? We could add a `topologyHint` parameter on the message itself alongside the `migrationCheckPeriod`. The `Message` is getting very bloated, but I've created an issue around refactoring it: https://github.com/faasm/faabric/issues/221

Shillaker (changes requested):

src/scheduler/Executor.cpp
```diff
@@ -528,9 +529,29 @@ void Executor::threadPoolThread(int threadPoolIdx)
...
+            // Note that when a task has been migrated, we need to perform all
+            // the normal executor shutdown, but we must NOT set the result for
+            // the call.
+            migrated = true;
+            selfShutdown = true;
+            returnValue = -99;
+
+            // MPI migration
+            if (msg.ismpi()) {
+                // TODO - delete the pending migration
```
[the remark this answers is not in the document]
csegarragonz: At the moment, faabric does not know when the migration (app-wise) has finished. In fact, faabric does not even know when a migration is taking place. Faabric only is aware that some functions throw exceptions because they are are migrated.

The workaround I have is the following:
* When we call `MpiWorld::prepareMigration` the local leader sets a boolean flag in the MPI world (e.g. `hasBeenMigrated`).
* Everytime we hit a barrier, after _all_ ranks have hit the barrier (and before they have left it). The local leader checks for the `hasBeenMigrated` flag.
* If `true` remove the pending migration from the map.

This uses the fact that faasm _will_ call `prepareMigration` when migrating, but can't really be tested from within faabric (I have tested it using the distributed test in faasm).

I can't think of a not ad-hoc way of doing this due to the lack of information in faabric.

Shillaker (approved):

PR #215, opened 2022-01-11
Dirty tracking performance tweaks

Shillaker:
This is the result of hacking away at the dirty tracking and diff merging process to try and make it run a bit faster. As a result, it's a grab-bag of changes, but they're all worth including. At the moment it's still unclear what the best combination of dirty tracking and diffing is, and we won't know until we include the `userfaultfd` dirty tracking implementation. For now, both are configurable, which allows us to benchmark them on different workloads. 

Changes:

- Pass around a list of `char` flags as a `std::vector<char>`  to express the dirty pages in a region. This replaces the old approach of computing a `std::pair<uint32_t, uint32_t>` to represent the dirty _regions_. Computing, then iterating over these regions added unnecessary computation and complex logic to the critical path of diff merging.
- Add new `XOR` diffing mode. This results in larger diffs, but is faster to compute, so may be more efficient for certain workloads. 
- Add a dummy `NoneDirtyTracker` so that users can switch off dirty tracking if it's not needed (also useful for benchmarking).
- Add a `singlehost` flag to `BatchExecuteRequest`s in the scheduler if that request is only executed on one host.
- Don't do dirty tracking and merging diffs when only executing on a single host.
- Avoid using a `std::map` to ensure `SnapshotMergeRegions` are sorted, instead sort once just before they're used. 

Couple of the other changes I tacked on:

- Switch from `try_lock_until` to simply `lock` in `PointToPointBroker` as this was hurting performance in message-heavy applications.
- More trace logging around scheduling.
- Add gdb config to avoid catching `SIGSEGV`. This is necessary to support debugging using the segfault dirty page tracking. This won't be necessary once we switch to `userfaultfd`.
- Reduce minimum CMake version to `3.18`. AFAICT this doesn't break anything, and we should keep it as low as possible to ensure maximum compatibility.
- Switch off catch2 signal handling.


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.21.0)
+cmake_minimum_required(VERSION 3.18.2)
```
Shillaker, line 1: I found this to work for me locally, and as it's a minimum version I don't see any harm in lowering it.

Shillaker:
This has been divided up into smaller PRs.

PR #219, opened 2022-02-02
Use dirty page flags instead of dirty regions

Shillaker:
Previously we were taking the list of dirty pages, converting this into a `std::vector<std::pair<uint32_t, uint32_t>>` to express a set of regions, then passing this set of regions through the code to the snapshot diffing logic. In fact, it's more efficient and simpler to skip this computation and just directly pass around the list of dirty page flags as a `std::vector<char>`. This PR reworks the necessary logic and testing around this change. 

I've also added a couple of other small changes along the way:

- Move from using a `std::map<uint32_t, SnapshotMergeRegion>` to a `std::vector<SnapshotMergeRegion>` on each snapshot. This improves insertion performance, and we can sort the vector when necessary (note that sorting was the point of having the `map` in the first place). 
- Add `none` dirty tracking, which just marks all pages as dirty. This may be optimal for workloads with a small memory where most of that memory will be dirty anyway, so diffing every page outweighs the cost of the dirty tracking.
- Fix a race condition in `Executor::executeThreads`.
- Switch off catch2 signal handling to avoid needing to reinitialise the segfault dirty tracker in tests.
- Remove use of "note" and the like at the start of comments. The fact that it's a comment implies that the reader should take note. I regret doing this, it started with a couple of files then spiralled into doing a whole find and replace. Should have put it in another PR.


src/redis/Redis.cpp
```diff
@@ -82,7 +82,7 @@ std::string RedisInstance::loadScript(redisContext* context,
 Redis::Redis(const RedisInstance& instanceIn)
   : instance(instanceIn)
 {
-    // Note, connect with IP, not with hostname
```
Shillaker, line 85: I removed all instances of this strange habit I have of writing a superfluous "Note, " at the start of comments.

docs/development.md
```diff
@@ -95,10 +95,10 @@ In another terminal, (re)start the server:
 
 ```bash
 # Start
-./dist-tests/dev_server.sh
+./dist-test/dev_server.sh
 
 # Restart
-./dist-tests/dev_server.sh restart
+./dist-test/dev_server.sh restart
```
Shillaker, line 101: This was a typo.

include/faabric/util/bytes.h
```diff
@@ -27,20 +27,6 @@ int safeCopyToBuffer(const uint8_t* dataIn,
                      uint8_t* buffer,
                      int bufferLen);
 
-/*
```
Shillaker, line 30: Changes in logic rendered these functions useless.

include/faabric/util/dirty.h
```diff
@@ -25,24 +25,22 @@ class DirtyTracker
   public:
     virtual void clearAll() = 0;
 
-    virtual void reinitialise() = 0;
+    virtual std::string getType() = 0;
```
csegarragonz, line 28: Could we have add some lines to the class description comment (in lines 20-21) to add the different types of tracking methods and a very brief explanation?

I know this is also explained in each subclass description, but I think it would be helpful to have all information summarised together somewhere.

csegarragonz (approved):

include/faabric/util/dirty.h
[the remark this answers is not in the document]
Shillaker: I agree that seeing the options would be nice, but worry that a comment on this class would eventually get out of date and be counterproductive. The list is essentially a list of the subclasses, which would exist if we generated "proper" documentation (e.g. with doxygen). Let's discuss this offline as I think it's a wider point.

PR #232, opened 2022-02-21
Uffd dirty tracking

Shillaker:
This PR adds four new dirty tracking modes, all based on [`userfaultfd`](https://www.kernel.org/doc/html/latest/admin-guide/mm/userfaultfd.html):

- `uffd` - uses the `SIGBUS` handler to catch events triggered by accessing missing pages in demand-zero paged memory.
- `uffd-wp` - same as `uffd` but adds write-protected events to catch subsequent writes to write-protected pages. 
- `uffd-thread` - same as `uffd`, but using a background event thread to handle events. This has the benefit of distinguishing between read and write missing page events.
- `uffd-thread-wp` - same as `uffd-thread`, but adds write-protected events.

We leave the default dirty tracking mode as `segfault` as this is still the most reliable and efficient. However, in a subsequent PR I'll be benchmarking our whole menu of dirty tracking options to see which is best in which situation: https://github.com/faasm/faabric/pull/234.

Other small changes:

- Dirty tracker singleton is now a shared pointer to avoid having to initialise all the different types of dirty tracker as we did before.
- Deduped tests around dirty tracking. Merged into one larger parameterised test.


kubasz:
Just so you know, using the poll-based method in another thread for UFFD incurs 2-10x the latency of the SIGBUS-on-same-thread method on each page fault (based on my microbenchmarks). Feel free to take a look at my implementation at <https://github.com/auto-ndp/faasm/blob/master/src/wasm/UffdMemoryArenaManager.cpp> for how the sigbus-based method works

Shillaker:
@kubasz do you know if there's any way to distinguish between a read and a write event arising from a missing page when using the `SIGBUS` set-up? When using a background thread you can inspect the `flags` of the `uffd_msg` and check the `UFFD_PAGEFAULT_FLAG_WRITE` bit, but AFAICT that info can't be worked out in the `SIGBUS` version as all you get is the `siginfo_t` associated with the signal. 

Ideally the dirty tracking would only mark pages dirty if they've been _written_ to, and not just if they'd been accessed. However, in the `SIGBUS` case, it seems there's no way to tell the difference between a single read or a single write.

kubasz:
> @kubasz do you know if there's any way to distinguish between a read and a write event arising from a missing page when using the `SIGBUS` set-up? [...]

I had a look through the kernel source and sadly it seems the answer currently is no: <https://elixir.bootlin.com/linux/latest/source/fs/userfaultfd.c#L407> - with the sigbus flag the fault information is not communicated to the program (which is a bit strange imo, maybe they'd accept a patch that adds this information)

Shillaker:
Ok great, thanks for that Jakub; I ended up looking at the same code but thought/ hoped there might be some magic way to get around it. I agree, this could be some nice low-hanging fruit for a patch, can't see what the objection to it would be.

I've found the following from this work:

- Write-protected mode can't be used on shared memory, which seems to include any mapping from a file descriptor created by `memfd_create`. This makes it unusable for Faabric function memory that's been mapped from a snapshot.
- `SIGBUS` can't distinguish between reads and writes in events on missing pages (as discussed), hence may mark some clean pages as dirty. The inefficiency caused by this will depend on the workload.
- Use of a background thread should work, but is slower than the `SIGBUS` approach, and possibly also slower than the existing segfault approach (as per the [link](https://xzpeter.org/userfaultfd-wp-latency-measurements/) you shared a while ago). 

As it stands, the existing segfault dirty tracking still looks like the safest option as a default, but `userfaultfd` might be better for certain workloads. I'm going to write my own benchmarking script now to check each of the Faabric implementations as they stand in this PR and confirm.

kubasz:
I know that the CRIU and KVM people use UFFD for their own snapshot-restore mechanisms, but I don't know how complex redoing the memfd-based mechanism into a uffd-based mechanism in faasm would be. It would require changing snapshots to be regular memory mappings (not memfds), and then using the uffd copypages functionality to populate the local wasm memory arena with the snapshot data on page faults. It might not be worth the maintenance effort for you - if your primary applications didn't run into the same scaling issues I did it might not be worth it.

.github/workflows/tests.yml
```diff
@@ -81,7 +81,8 @@ jobs:
       TSAN_OPTIONS: "verbosity=1:halt_on_error=1:suppressions=/code/faabric/thread-sanitizer-ignorelist.txt:history_size=7"
       UBSAN_OPTIONS: "print_stacktrace=1:halt_on_error=1"
     container:
-      image: faasm/faabric:0.3.1
+      image: faasm/faabric:0.3.2
+      options: --privileged
```
Shillaker, line 85: Needed to actually run the `userfaultfd` tracking.

docker/faabric-base.dockerfile
```diff
@@ -37,6 +37,7 @@ RUN apt install -y \
     libtool \
     libunwind-13-dev \
     libz-dev \
+    linux-generic-hwe-20.04 \
```
Shillaker, line 40: The latest `userfaultfd` features are part of kernel 5.13, which requires HWE.

src/util/dirty.cpp
```diff
@@ -13,47 +17,166 @@
...
 };
 
-DirtyTracker& getDirtyTracker()
+/**
+ * Thread-safe wrapper around dirty tracking data, necessary for use with the
+ * background event thread in the userfaultfd tracker. Although the logic around
+ * the interaction between the event thread and the main thread should in theory
+ * protect against concurrent accesses, we need this to keep TSan happy.
+ */
+class ThreadsafeDirtyTrackingRecord final : public DirtyTrackingRecord
```
csegarragonz, line 82: Very minor, `ThreadSafeDirtyTrackingRecord` feels like a more natural capitalisation to me.

csegarragonz (approved):

.github/workflows/tests.yml
kubasz: Another option is setting the `/proc/sys/vm/unprivileged_userfaultfd` sysctl to 1

kubasz: But probably not an option on github actions runners actually

include/faabric/util/dirty.h
```diff
@@ -14,6 +16,21 @@
 #define PAGEMAP_ENTRY_BYTES sizeof(uint64_t)
 #define PAGEMAP_SOFT_DIRTY (1Ull << 55)
 
+// The following isn't available using HWE on 20.04, so copied from:
```
kubasz, line 19: There are a few constants in this file that are different (eg. UFFD_API_FEATURES lists all valid feature bits), it might be better to just copy in the entire userfaultfd.h file into the repo and include it above from the local copy rather than the system-wide one

src/util/dirty.cpp
```diff
@@ -13,47 +17,166 @@
...
+
+void* pageAlignAddress(void* faultAddr)
+{
+    uintptr_t addr = (uintptr_t)faultAddr;
+    addr &= -HOST_PAGE_SIZE;
+    auto* alignedAddr = (void*)addr;
+    return alignedAddr;
+}
+
+// Thread-local tracking information for dirty tracking using signal
```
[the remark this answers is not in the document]
Shillaker: I think the style is consistent, just not in the way you're thinking. I use the following convention:

- Class and method comments that will go into documentation (i.e. above class and method declarations) with `/** */`
- All other single or multi-line comments with `//`

Now that I think about it, this is actually Javadoc style that I picked up years ago, and probably not in line with the Moz C/C++ style guide. I've used it in more than just this PR, so if we want to change it, we should do a codebase-wide effort and ideally enforce it in the formatting checks.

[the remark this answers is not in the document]
Shillaker: Given that we're using Doxygen we should go with what they say in this case rather than the C/C++ guide: https://www.doxygen.nl/manual/docblocks.html.

Their first example uses Javadoc style as I mention, so I'd be happy to stick with that. It makes it easier to distinguish between class/ method docs (`/**`) and implementation-specific comments (i.e. `//`).

Let's discuss offline though.

.github/workflows/tests.yml
Shillaker: Ah nice, yes this is a bit of a blunt instrument. We also use `privileged` in Faasm for cgroups, so it's a bit lazy but works across the board.

src/util/dirty.cpp
```diff
@@ -110,6 +221,24 @@ void SoftPTEDirtyTracker::stopTracking(std::span<uint8_t> region)
     // Do nothing
 }
 
+void SoftPTEDirtyTracker::resetPTEs()
+{
+    PROF_START(ClearSoftPTE)
```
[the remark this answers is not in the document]
Shillaker: I'm still using it in my profiling.

include/faabric/util/dirty.h
Shillaker: Hmm yep good point, my thinking was that if we use the system-wide version for everything we can, then we'll reduce the risk of diverging from the API (and some potentially very difficult-to-diagnose issues). However, it makes it much easier to diff and work out what's gone wrong if we have the whole file copied in, so I'll do that.

PR #233, opened 2022-02-23
Add task to generate release body

csegarragonz:
In this PR I introduce a task to generate the body for a new release.

We use the [git-cliff](https://github.com/orhun/git-cliff) project through their docker image to generate a changelog between the current tag and the latest release's tag.

I have customised the format of the release body a bit to fit our Github usage (i.e. squash commits). This is orthogonal to any automation efforts we want to do in the future.

I paste what the release body would look like now for discussion (note that we are missing this actual PR, hence why we see three dangling commits at the beginning).

=====

# Faabric's Changelog

Here is what has changed since last release:

## [0.3.1] - 2022-02-23
 * Ci: generate release body with changelog
 * Ci: add step to generate changelog
 * Bumping code version for release
 * Detect CPU information (#231)
 * Executor context (#228)
 * Conan cache improvements (#230)
 * Remove SGX-related message fields (#229)
 * Master to main (#227)

## [0.3.0] - 2022-02-11
 * Version bump (#226)
 * Generate documentation (#223)
 * Executor decision caching and single-host optimisations (#222)
 * XOR diffing (#224)
 * Use conan action (#225)
 * Use dirty page flags instead of dirty regions (#219)
 * Periodically Check For Migration Opportunities (#214)
 * Add gdbinit config (#217)
 * Remove timed mutex from ptp broker (#220)
 * Cmake minimum version and skylake (#218)
 * Cmake: quote conan revisions (#213)
 * Dirty tracking performance improvements (#210)
 * Introduce Fixed-Capacity Queues (#205)
 * Merge regions overhaul (#201)
 * Add in-proc PAIR sockets (#206)
 * Use matrix build for sanitisers (#208)
 * No need to acquire a lock to set a message id in mpi (#203)
 * Snapshots overhaul (#176)

## [0.2.3] - 2021-12-20
 * Pin Conan dependencies to revision (#200)
 * Locality aware gather (#198)
 * Return executed time with successful status request (#197)
 * Fix reduce bug and add regression test (#195)
 * More race condition fixes (#193)
 * Locality Aware Reduce (#191)
 * Minor fixes (#192)
 * Locality Aware Broadcast (#185)
 * Enable/disable scheduling topology hints through an env. variable (#188)
 * Re-use conan cache for built dependencies (#187)
 * Adding support for sanitised builds (#182)
 * Retry `GET` from Redis for messages in the Execution Graph (#181)
 * Use `NEVER_ALONE` for chained MPI calls (#184)
 * Add Scheduling Topology Hints (#180)
 * Fix race condition in scheduler reset (#179)
 * Get `ExecGraph` hosts (#178)
 * Make sure conan packages get compiled with C++20 enabled (#177)
 * Track Message Type in Execution Graph (#175)
 * Snapshot diff merging fixes (#174)
 * Scheduler hints and fixes for distributed coordination bugs (#169)
 * Fixes for MPI message graph (#173)
 * Adding arbitrary information to execution graph (#166)
 * Add distributed coordination operations (#161)
 * Use faster lld as the linker (#168)

## [0.2.1] - 2021-10-27
 * Cmake Cleanup Faabric Part 2 (#165)
 * Cmake cleanup Faabric Part 1 (#164)
 * Initialise point-to-point mappings from scheduler output (groundwork) (#160)
 * README flair (#159)
 * Allow merge regions to cross pages (#158)
 * Allow generating execution graphs from MPI runs (#154)
 * Generalised point-to-point messaging (#151)
 * Fix issue with hanging shutdown when not really started (#155)
 * Remove sgx check from faabric runner (#153)
 * Add snapshot if exists function to snapshot registry (#152)
 * Multi-threaded server message endpoints (#147)
 * Change default mpi base port and allow value to be set through env variable (#150)
 * Add ignored snapshot regions (#148)
 * Add crash handler to tests (#149)
 * Merge docker compose files (#146)
 * Add a crash handler that prints backtraces to FaabricMain users (#145)
 * Compare pointers rather than ids in Scheduler::notifyExecutorShutdown (#144)
 * Snapshot diff merge operations (#142)
 * Not use emplace for editing vector elements (#143)
 * Specify timeouts when timing out (#140)
 * Add `snapshotExists` function to registry (#136)
 * Set `appid` on messages (#137)
 * Fix race condition in test (#138)
 * Latency fixes (#135)


tasks/git.py
```diff
@@ -24,3 +23,40 @@ def tag(ctx, force=False):
...
+
+
+@task
+def release_body(ctx, file_path="/tmp/release_body.md"):
+    """
+    Generate body for release with detailed changelog
+    """
+    docker_cmd = [
+        "docker run -t -v",
+        "{}/..:/app/".format(PROJ_ROOT),
```
csegarragonz, line 47: Note: this only works if faabric is used as a submodule in faasm. This is because, as a submodule, the `.git` folder is a softlink to `../.git/modules/faabric`.

Shillaker: We shouldn't assume this, Faabric should be treated as a separate library and worked on in a separate checkout. Editing Faabric as a submodule of Faasm should be the exception not the rule.

Shillaker (changes requested):
Nice one, this looks good. I think we can go one step further and actually do the release from the command. See the equivalent command in Faasm: https://github.com/faasm/faasm/blob/main/faasmcli/faasmcli/tasks/release.py#L152

Splitting it into creating and then publishing the release is probably good for now, but can eventually be merged into one step.

Shillaker:
@csegarragonz let's also remove the commit bodies and just include the headline commits. There are links to the PRs that people can click on if they want more detail.

csegarragonz:
@Shillaker we don't need to add `git-cliff` as a dependency, as we rely on the docker image to run it. further, `git-cliff` is written in rust, what means that we couldn't use `pip` to manage its installation anyway.

i add the new release body format in the updated pr description, i agree with you and think it is more concise and readable.

Shillaker:
Ah whoops, sorry I'd assumed `git-cliff` was a python module. Interesting choice to go with Rust...

docs/source/development.md
```diff
@@ -188,3 +181,35 @@ inv container.push
...
+just tagged the code (see previous section).
+
+Then, you can create a release on [Github](https://github.com/faasm/faabric/releases)
+and publish it from the command line. If it is the first time you are creating
+a release you will have to configure a Github access token (see below).
+
+First, generate a draft release:
+
+```bash
+inv git.release_create
```
Shillaker, line 196: I don't think this will work, I have a feeling Invoke will replace all underscores with hyphens.

Shillaker (approved):
This looks great. Only possible issue is with the docs, I think those `inv` commands won't work but may be wrong.

PR #238, opened 2022-03-01
Add icelake to CPU list

Shillaker:


kubasz:
If you want to target the CPU of the machine compiling the code, why not just use `-march=native`?

Shillaker:
Good question, had I known about `native` I would certainly have put that in instead :smile: . Do you know where I can find a list of supported targets? I couldn't find a way to get clang to tell me. 

This is just an experiment at the moment;  I had to remove the `skylake`-only approach to run on a wider range of machines for our experiments. Once I'm done with the changes I'll make it configurable then revert to a generic build for releases.

kubasz:
`llc -march=x86 -mattr=help` prints out the supported CPUs and cpu attributes in LLVM (including clang)

PR #241, opened 2022-03-15
Reintroduce diffs as part of snapshot result

Shillaker:
This primarily fixes a race condition, but also a bug with adjacent merge regions.

The race condition occurs when a remote thread result is received and acted upon _before_ the snapshot diffs from that thread have been applied. This exists because the result and diffs are done in two separate requests, and the server can sometimes process them out of order. The solution is to merge the diffs into the thread result (which is what we used to do but it got lost in one of the big reworkings of the snapshotting logic).

The bug with adjacent merge regions occurred in the process for "filling the gaps" between merge regions. If two merge regions were adjacent, we'd insert another one with zero length between them, rather than skipping it completely. Zero length merge regions are interpreted as running from their offset to the end of the snapshot, so this really messed things up. It's now fixed with a simple change to the loop in the "gap filling" function.

### DIFF
--- a/include/faabric/scheduler/Scheduler.h
+++ b/include/faabric/scheduler/Scheduler.h
@@ -171,11 +171,6 @@ class Scheduler
                          const std::string& key,
                          const std::vector<faabric::util::SnapshotDiff>& diffs);
 
-    void pushSnapshotDiffs(
-      const faabric::Message& msg,
-      const std::string& snapshotKey,
-      const std::vector<faabric::util::SnapshotDiff>& diffs);
-
     void setThreadResultLocally(uint32_t msgId, int32_t returnValue);
 
     std::vector<std::pair<uint32_t, int32_t>> awaitThreadResults(
--- a/include/faabric/scheduler/Scheduler.h
+++ b/include/faabric/scheduler/Scheduler.h
@@ -166,12 +166,10 @@ class Scheduler
 
     faabric::Message getFunctionResult(unsigned int messageId, int timeout);
 
-    void setThreadResult(const faabric::Message& msg, int32_t returnValue);
-
-    void setThreadResultWithDiffs(
-      const faabric::Message& msg,
-      int32_t returnValue,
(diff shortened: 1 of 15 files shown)


include/faabric/scheduler/Scheduler.h
```diff
@@ -166,12 +166,10 @@ class Scheduler
 
     faabric::Message getFunctionResult(unsigned int messageId, int timeout);
 
-    void setThreadResult(const faabric::Message& msg, int32_t returnValue);
-
-    void pushSnapshotDiffs(
-      const faabric::Message& msg,
-      const std::string& snapshotKey,
-      const std::vector<faabric::util::SnapshotDiff>& diffs);
+    void setThreadResult(const faabric::Message& msg,
```
Shillaker, line 169: This is the main change, merging `pushSnapshotDiffs` and `setThreadResult`.

include/faabric/util/snapshot.h
```diff
@@ -15,6 +15,10 @@
 
 namespace faabric::util {
 
+/**
+ * Defines the permitted datatypes for snapshot diffs. Each has a predefined
+ * length, except for the raw option which is used for generic streams of bytes.
+ */
```
Shillaker, line 21: I added these docs as part of a general effort to document things gradually.

src/scheduler/Executor.cpp
```diff
@@ -518,40 +520,19 @@ void Executor::threadPoolThread(int threadPoolIdx)
...
-                      diffs.size(),
-                      faabric::util::funcToString(msg, false),
-                      mainThreadSnapKey,
-                      msg.groupid());
-
-                    snap->queueDiffs(diffs);
-                } else if (isLastInBatch) {
-                    sch.pushSnapshotDiffs(msg, mainThreadSnapKey, diffs);
-                }
-            }
```
Shillaker, line 552: This was the original place where the snapshot diff pushing was happening, it's now been moved later in the method, and this logic lives in the scheduler now.

PR #243, opened 2022-03-16
Fix for scheduling on overloaded hosts

Shillaker:
This fixes an issue where scheduling functions on remote hosts would break if those hosts were overloaded.

### DIFF
--- a/src/scheduler/Scheduler.cpp
+++ b/src/scheduler/Scheduler.cpp
@@ -386,11 +386,13 @@ faabric::util::SchedulingDecision Scheduler::doSchedulingDecision(
               registeredHosts[funcStr];
 
             for (const auto& h : thisRegisteredHosts) {
-                // Work out resources on this host
+                // Work out resources on the remote host
                 faabric::HostResources r = getHostResources(h);
                 int available = r.slots() - r.usedslots();
 
-                // Floor this at zero and claim as many as we can
+                // We need to floor at zero here in case the remote host is
+                // overloaded, in which case its used slots will be greater than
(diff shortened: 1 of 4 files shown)


src/scheduler/Scheduler.cpp
```diff
@@ -486,7 +498,12 @@ faabric::util::SchedulingDecision Scheduler::doSchedulingDecision(
     }
 
     // Sanity check
-    assert(hosts.size() == nMessages);
+    if (hosts.size() != nMessages) {
+        SPDLOG_ERROR(
+          "Serious scheduling error: {} != {}", hosts.size(), nMessages);
+
+        throw std::runtime_error("Not enough scheduled hosts for messages");
```
Shillaker, line 505: This should never happen, but given that it did happen, I found it much nicer to have a message than an assertion error.

src/scheduler/Scheduler.cpp
```diff
@@ -425,9 +432,14 @@ faabric::util::SchedulingDecision Scheduler::doSchedulingDecision(
...
 
-                // Work out resources on this host
+                // Work out resources on the remote host
                 faabric::HostResources r = getHostResources(h);
                 int available = r.slots() - r.usedslots();
+
+                // We need to floor at zero here in case the remote host is
+                // overloaded, in which case its used slots will be greater than
+                // its available slots.
+                available = std::max<int>(0, available);
```
csegarragonz, line 442: Good catch

csegarragonz (approved):

PR #244, opened 2022-03-26
Avoid copying 0MQ message data

Shillaker:
Avoid copying incoming data on receipt of messages.


src/transport/MessageEndpoint.cpp
```diff
@@ -319,8 +319,7 @@ std::optional<Message> MessageEndpoint::recvNoBuffer(zmq::socket_t& socket)
       },
       "recv_no_buffer")
 
-    // Copy the received message to a buffer whose scope we control
-    return Message(msg);
```
Shillaker, line 323: This is the copy we can avoid.

PR #252, opened 2022-03-30
Scheduler-controlled executor shutdown

Shillaker:
Currently the executor lifecycle is quite complex, it works as follows:

- Scheduler spawns executors in response to incoming requests.
- Executors lazily create threads in a thread pool to handle requests.
- Scheduler assigns tasks to a given executor, which then puts those tasks on a queue for their thread pool to execute.
- Thread pool threads listen for tasks on a queue
- _If a thread doesn't receive a task in the given timeout, it shuts itself down_
- _If the thread is the last one in the pool, it notifies the scheduler that the executor has finished_

These last two points are problematic, as they require the thread to tidy up executor state, and for the thread to contact the scheduler, which will eventually loop around to tidying up the thread itself. This loop is confusing, and makes the clean-up code very difficult to follow. More importantly, it requires us to keep track of dead executors and dead threads, which means their resources can't be freed.

This PR changes the last two points to the following

- Threads continue to check for messages on their queue until the executor tells them to stop.
- Executors continue to operate until the scheduler tells them to stop.

This turns the existing communication loop into a top-down shutdown, where each layer listens for a shutdown message from the layer above it. Each layer is also responsible only for clearing up its own resources, and not those of the layer above.

Changes:

- Scheduler now operates a "reaper" thread, which wakes up on an interval to remove stale executors (those that have not executed a message for a while)
- Thread pool threads don't ever shut themselves down, they only stop when they receive a shutdown message from their parent executor.
- Refactor registered hosts tracking to use just user/ function strings rather than the whole message object (which needed to be passed around before).
- Remove dead executors and threads.

The downsides to this are that we now have a stop-the-world like clean-up in the scheduler, but it's worth it for removing the complexity and memory leaks.


include/faabric/scheduler/FunctionMigrationThread.h
```diff
@@ -1,26 +1,23 @@
...
-// are migration opportunities for in-flight apps that have opted in to
-// being checked for migrations.
-class FunctionMigrationThread
+
+/**
+ * Background thread that, every wake up period, will check if there
+ * are migration opportunities for in-flight apps that have opted in to
+ * being checked for migrations.
+ */
+class FunctionMigrationThread : public PeriodicBackgroundThread
```
Shillaker, line 18: I've refactored this to extract the periodic-wakeup-thread bits into `PeriodicBackgroundThread`.

include/faabric/scheduler/Scheduler.h
```diff
@@ -80,15 +82,15 @@ class Executor
       faabric::Message& msg,
       bool createIfNotExists = false);
 
-    void updateMainThreadSnapshot();
```
Shillaker, line 83: This function wasn't implemented.

PR #253, opened 2022-04-06
Add utility functions to convert from byte arrays to hex strings

csegarragonz:
I need these to convert SGX quotes to HTTP requests for validation by the external attestation service.

### DIFF
--- a/include/faabric/util/bytes.h
+++ b/include/faabric/util/bytes.h
@@ -1,8 +1,10 @@
 #pragma once
 
 #include <algorithm>
+#include <iomanip>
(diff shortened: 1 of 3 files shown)


Shillaker (approved):
This looks good, but if it's only related to Faasm stuff, it should live in Faasm. If there's no usage in Faabric, it might get deleted in a clean-up in future, then we'd only realise once the Faasm SGX tests start failing.

csegarragonz:
I will merge this in, and then once the attestation PR is merged in, move this to a utility folder in faasm.

PR #254, opened 2022-04-11
Fix `MessageEndpointServer` crash during `TIMEOUT`

csegarragonz:
Changes:

- Make message server and underlying message handlers configurable on timeout (to enable testing).
- Handle timeout message in server.
- Add test to check with short timeouts.

### DIFF
--- a/include/faabric/transport/MessageEndpoint.h
+++ b/include/faabric/transport/MessageEndpoint.h
@@ -13,10 +13,9 @@
 #define ANY_HOST "0.0.0.0"
 
 // These timeouts should be long enough to permit sending and receiving large
-// messages, note that they also determine the period on which endpoints will
-// re-poll.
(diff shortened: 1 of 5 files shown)


include/faabric/transport/MessageEndpoint.h
```diff
@@ -13,10 +13,9 @@
 #define ANY_HOST "0.0.0.0"
 
 // These timeouts should be long enough to permit sending and receiving large
-// messages, note that they also determine the period on which endpoints will
-// re-poll.
-#define DEFAULT_RECV_TIMEOUT_MS 60000
-#define DEFAULT_SEND_TIMEOUT_MS 60000
```
Shillaker, line 19: These were the same, and not used consistently, so I merged it into a single constant.

PR #259, opened 2022-04-20
Support concurrent MPI executions

csegarragonz:


tests/test/scheduler/test_multiple_mpi_worlds.cpp
```diff
@@ -0,0 +1,291 @@
...
+          rankA1, rankA2, BYTES(buffer), MPI_INT, messageData.size(), &status);
+        std::vector<int> actualB(buffer, buffer + messageData.size());
+        REQUIRE(actualB == messageData);
+        REQUIRE(status.MPI_ERROR == MPI_SUCCESS);
+        REQUIRE(status.MPI_SOURCE == rankA1);
+        REQUIRE(status.bytesSize == messageData.size() * sizeof(int));
+    }
+}
+
+TEST_CASE_METHOD(RemoteMultiWorldMpiTestFixture,
```
csegarragonz, line 198: This is the test that fails. Unfortunately, as explained in the PR, I don't think there is a way to properly test this without actually porting the MPI-native tests to our distributed tests.

PR #260, opened 2022-04-22
Distributed Tests for MPI

### LINKED ISSUE #62: [MPI-native] Mount libraries to run tests
The way we test the `mpi-native` implementation in [Faabric](https://github.com/faasm/faabric/tree/master/mpi-native), only updates the logic within when we bump the version, and create a new release. As a consequence, we won't know if the changes break the `mpi-native` code until we do so.

However, this is hard to circunvent, as we rely heavily on `docker-compose` to scale up or down the number of processes we use in the native deployment (bear in mind this whole subproject is aimed for developing and testing purposes, not production environments). Thus, a possible solution would be to mount the updated libraries in the `docker-compose` file [here](https://github.com/faasm/faabric/blob/master/mpi-native/mpi-native-docker-compose.yml).

Further, recent PRs like #67 and #65 have increased the number of implemented MPI calls, hence these ought to be included in `mpi-native`. For a possible implementation, it's always good to see how it's done in [faasm](https://github.com/faasm/faasm/blob/master/src/wavm/mpi.cpp).

csegarragonz:
In this PR we migrate the remote transport layer used in MPI to use the existing PTP broker. This fixes some bugs that had appeared in MPI in the last months, and also allows for multi-tenant MPI.

The design of the PTP broker made it impossible to keep using our hacked distributed MPI tests, so I also take the opportunity in this PR to migrate the rotten `mpi-native` stuff to proper distributed tests.

I am aware that including both changes (PTP broker, and MPI distributed tests) makes it harder to review the PR, but there was a chicken and egg problem:
* MPI (w/out the PTP broker) is broken atm, so we can not port the `mpi-native` stuff to distributed tests.
* We can not test the migration to the PTP broker w/out distributed tests.

On the bright side, this PR includes:
* Working distributed tests for all MPI test-cases (in Faasm, `mpi-native`, and mocked distributed tests)
* Working distributed test for MPI migration
* Working distributed test for multi-tenant MPI execution (including one test with migration)

To get the distributed test for MPI migration I have slightly changed the way migration works. Now, migrated functions do have a special return value, and we can iterate the execution graph and infer which functions were migrated and where. In addition, these tests rely (a bit) on the timing of both hosts (as they try to reproduce a real migration) so they may be risky to have in GHA.

This PR is blocked by #261. But given its size, I think it is ready for review. At the moment, I change the number of threads in the PTP server to 1 to check that the distributed tests run succesfully.

Closes #62


src/scheduler/MpiContext.cpp
```diff
@@ -25,10 +25,7 @@ int MpiContext::createWorld(faabric::Message& msg)
 
     // Create the MPI world
     scheduler::MpiWorldRegistry& reg = scheduler::getMpiWorldRegistry();
-    scheduler::MpiWorld& world = reg.createWorld(msg, worldId);
-
-    // Broadcast setup to other hosts
-    world.broadcastHostsToRanks();
```
csegarragonz, line 31: Don't need to do this anymore, as the information is already in the PTP broker.

tests/test/scheduler/test_mpi_world.cpp
```diff
@@ -137,7 +137,9 @@ TEST_CASE_METHOD(MpiBaseTestFixture, "Test cartesian communicator", "[mpi]")
     for (int i = 0; i < dims[0]; i++) {
         for (int j = 0; j < dims[1]; j++) {
             std::vector<int> coords = { i, j, 0 };
-            int rank, src, dst;
+            int rank;
```
csegarragonz, line 140: Changes in this file are only fixes of clang tidy warnings (now that I see them). I only include them here as it is related to the PR. Happy to discard them though.

tests/utils/fixtures.h
```diff
@@ -195,9 +195,54 @@ class ConfTestFixture
     faabric::util::SystemConfig& conf;
 };
 
+class PointToPointTestFixture
```
csegarragonz, line 198: Moved these two fixtures above the MPI ones.

tests/test/scheduler/test_remote_mpi_worlds.cpp
```diff
@@ -11,1088 +11,14 @@
...
 
+// The tests in this file are used to test the internal behaviour of MPI when
+// running in a distributed behaviour. They should test very specific things
+// _always_ in mocking mode. For truly multi-host MPI tests you must write
+// an actual distributed test.
+
 using namespace faabric::scheduler;
 
 namespace tests {
-class RemoteCollectiveTestFixture : public RemoteMpiTestFixture
```
csegarragonz, line 17: I removed all the tests in this file that are now covered by distributed tests.

Shillaker (commented):
A couple of TODOs left to do, but other than that LGTM

csegarragonz:
@Shillaker i am not sure i can see your review comments :/

Shillaker (approved):

PR #261, opened 2022-04-26
In-order message reception for the PTP server

csegarragonz:
In this PR we add support for in-order message delivery in the PointToPoint (PTP) messaging layer. Ordering is needed for applications like MPI, and is broken at the `PointToPointServer` where multiple threads read from the PTP port, and push to downstream in-proc sockets.

We include a sequence number in the `Message` header, that we read and update if the `PointToPointBroker` is configured to maintain message ordering. Preliminary benchmarking shows a reduction in throughput of 40-50% when compared to out-of-order message delivery. For the moment, we will test this implementation with MPI and see if further profiling and optimisation is needed.

See [here](https://github.com/faasm/faabric/runs/6179026854?check_suite_focus=true#step:5:1152) for the test failing in GHA (showcasing that the existing PTP layer does not maintain message order).

### DIFF
--- a/include/faabric/transport/MessageEndpoint.h
+++ b/include/faabric/transport/MessageEndpoint.h
@@ -28,6 +28,7 @@
 // - Message sequence number of in-order message delivery default -1 (int)
 #define NO_HEADER 0
 #define HEADER_MSG_SIZE (sizeof(uint8_t) + sizeof(size_t) + sizeof(int))
+#define NO_SEQUENCE_NUM -1
 
 #define SHUTDOWN_HEADER 220
 static const std::vector<uint8_t> shutdownPayload = { 0, 0, 1, 1 };
@@ -71,7 +72,7 @@ class MessageEndpoint
                      uint8_t header,
                      const uint8_t* data,
                      size_t dataSize,
-                     int sequenceNumber = -1);
+                     int sequenceNumber = NO_SEQUENCE_NUM);
 
     Message recvMessage(zmq::socket_t& socket, bool async);
 
@@ -94,7 +95,7 @@ class AsyncSendMessageEndpoint final : public MessageEndpoint
     void send(uint8_t header,
               const uint8_t* data,
               size_t dataSize,
-              int sequenceNum = -1);
+              int sequenceNum = NO_SEQUENCE_NUM);
 
   protected:
     zmq::socket_t socket;
@@ -109,7 +110,7 @@ class AsyncInternalSendMessageEndpoint final : public MessageEndpoint
     void send(uint8_t header,
               const uint8_t* data,
               size_t dataSize,
-              int sequenceNumber = -1);
+              int sequenceNumber = NO_SEQUENCE_NUM);
 
   protected:
     zmq::socket_t socket;
--- a/include/faabric/transport/MessageEndpoint.h
+++ b/include/faabric/transport/MessageEndpoint.h
@@ -28,7 +28,6 @@
 // - Message sequence number of in-order message delivery default -1 (int)
 #define NO_HEADER 0
 #define HEADER_MSG_SIZE (sizeof(uint8_t) + sizeof(size_t) + sizeof(int))
-#define NO_SEQUENCE_NUM -1
 
 #define SHUTDOWN_HEADER 220
 static const std::vector<uint8_t> shutdownPayload = { 0, 0, 1, 1 };
--- a/include/faabric/transport/MessageEndpoint.h
+++ b/include/faabric/transport/MessageEndpoint.h
@@ -22,8 +22,12 @@
 // things haven't yet completed (usually only when there's an error).
 #define LINGER_MS 100
 
+// The header structure is:
+// - Message code (uint8_t)
+// - Message body size (size_t)
+// - Message sequence number of in-order message delivery default -1 (int)
 #define NO_HEADER 0
-#define HEADER_MSG_SIZE (sizeof(uint8_t) + sizeof(size_t))
+#define HEADER_MSG_SIZE (sizeof(uint8_t) + sizeof(size_t) + sizeof(int))
 
 #define SHUTDOWN_HEADER 220
 static const std::vector<uint8_t> shutdownPayload = { 0, 0, 1, 1 };
@@ -66,7 +70,8 @@ class MessageEndpoint
     void sendMessage(zmq::socket_t& socket,
                      uint8_t header,
                      const uint8_t* data,
-                     size_t dataSize);
+                     size_t dataSize,
+                     int sequenceNumber = -1);
 
     Message recvMessage(zmq::socket_t& socket, bool async);
 
@@ -98,7 +103,10 @@ class AsyncInternalSendMessageEndpoint final : public MessageEndpoint
     AsyncInternalSendMessageEndpoint(const std::string& inProcLabel,
                                      int timeoutMs = DEFAULT_SOCKET_TIMEOUT_MS);
 
-    void send(uint8_t header, const uint8_t* data, size_t dataSize);
+    void send(uint8_t header,
+              const uint8_t* data,
+              size_t dataSize,
+              int sequenceNumber = -1);
 
   protected:
     zmq::socket_t socket;
--- a/include/faabric/transport/MessageEndpoint.h
+++ b/include/faabric/transport/MessageEndpoint.h
@@ -91,7 +91,10 @@ class AsyncSendMessageEndpoint final : public MessageEndpoint
                              int portIn,
                              int timeoutMs = DEFAULT_SOCKET_TIMEOUT_MS);
 
-    void send(uint8_t header, const uint8_t* data, size_t dataSize);
+    void send(uint8_t header,
+              const uint8_t* data,
+              size_t dataSize,
+              int sequenceNum = -1);
 
   protected:
     zmq::socket_t socket;
--- a/include/faabric/transport/MessageEndpointClient.h
+++ b/include/faabric/transport/MessageEndpointClient.h
@@ -14,9 +14,14 @@ class MessageEndpointClient
                           int syncPort,
                           int timeoutMs = DEFAULT_SOCKET_TIMEOUT_MS);
 
-    void asyncSend(int header, google::protobuf::Message* msg);
+    void asyncSend(int header,
+                   google::protobuf::Message* msg,
+                   int sequenceNum = -1);
 
-    void asyncSend(int header, const uint8_t* buffer, size_t bufferSize);
+    void asyncSend(int header,
+                   const uint8_t* buffer,
+                   size_t bufferSize,
+                   int sequenceNum = -1);
 
     void syncSend(int header,
                   google::protobuf::Message* msg,
--- a/include/faabric/transport/MessageEndpointClient.h
+++ b/include/faabric/transport/MessageEndpointClient.h
@@ -16,12 +16,12 @@ class MessageEndpointClient
 
     void asyncSend(int header,
                    google::protobuf::Message* msg,
-                   int sequenceNum = -1);
+                   int sequenceNum = NO_SEQUENCE_NUM);
 
     void asyncSend(int header,
                    const uint8_t* buffer,
                    size_t bufferSize,
-                   int sequenceNum = -1);
+                   int sequenceNum = NO_SEQUENCE_NUM);
 
     void syncSend(int header,
                   google::protobuf::Message* msg,
--- a/include/faabric/transport/PointToPointBroker.h
+++ b/include/faabric/transport/PointToPointBroker.h
@@ -138,9 +138,7 @@ class PointToPointBroker
 
     void initSequenceCounters(int groupId);
 
-    void incrementSentMsgCount(faabric::PointToPointMessage& msg,
-                               int groupId,
-                               int recvIdx);
+    int getAndIncrementSentMsgCount(int groupId, int recvIdx);
 
     void incrementRecvMsgCount(int groupId, int sendIdx);
 
--- a/include/faabric/transport/PointToPointBroker.h
+++ b/include/faabric/transport/PointToPointBroker.h
@@ -117,7 +117,7 @@ class PointToPointBroker
 
     void resetThreadLocalCache();
 
-    void setMustOrderMessages(bool mustOrderMsgs);
+    bool setIsMessageOrderingOn(bool mustOrderMsgs);
 
   private:
     faabric::util::SystemConfig& conf;
@@ -134,7 +134,7 @@ class PointToPointBroker
 
     Message doRecvMessage(int groupId, int sendIdx, int recvIdx);
 
-    std::atomic<bool> orderMsg;
+    std::atomic<bool> _isMessageOrderingOn;
 
     void initSequenceCounters(int groupId);
 
--- a/include/faabric/transport/PointToPointBroker.h
+++ b/include/faabric/transport/PointToPointBroker.h
@@ -106,7 +106,8 @@ class PointToPointBroker
                      int sendIdx,
                      int recvIdx,
                      const uint8_t* buffer,
-                     size_t bufferSize);
+                     size_t bufferSize,
+                     int sequenceNum = -1);
 
     std::vector<uint8_t> recvMessage(int groupId, int sendIdx, int recvIdx);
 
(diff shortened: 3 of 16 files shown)


src/transport/PointToPointBroker.cpp
```diff
@@ -502,6 +519,87 @@ std::set<int> PointToPointBroker::getIdxsRegisteredForGroup(int groupId)
...
+        } else {
+            recvMsgCount = std::make_unique<std::vector<int>>(groupSize, 1);
+        }
+    }
+
+    assert(sendIdx < recvMsgCount->size());
+    recvMsgCount->at(sendIdx)++;
+}
+
 void PointToPointBroker::sendMessage(int groupId,
```
Shillaker, line 603: Proposed change: header message to contain sequence number:

- Add `int` to `HEADER_MSG_SIZE` at the end.
- Set to `-1` when not set.
- Then add to the `transport::Message` class alongside the `size`, set on receive.
- Update this `PointToPointBroker::sendMessage` to take a sequence number too, thread that through to the message endpoint `sendMessage`.

src/transport/PointToPointBroker.cpp
```diff
@@ -502,6 +519,87 @@ std::set<int> PointToPointBroker::getIdxsRegisteredForGroup(int groupId)
...
+        // We assume that group indexes will always go from 0 to `groupSize - 1`
+        // We also want to start counting from one. This way, an unset sequence
+        // number (0 by default in protobuf) will indicate that it is unset.
+        if (recvMsgCount != nullptr) {
+            recvMsgCount->clear();
+            recvMsgCount->resize(groupSize, 1);
+        } else {
+            recvMsgCount = std::make_unique<std::vector<int>>(groupSize, 1);
+        }
+    }
```
Shillaker, line 571: Could share this with the other lazy initialisation, it would look something like:

```cpp
// Receive version (will be similar for send)
if(groupId != currentGroupId) {
    recvMsgCount = initSequenceCounter(groupId);
}
```

src/transport/PointToPointBroker.cpp
```diff
@@ -566,10 +672,74 @@ std::vector<uint8_t> PointToPointBroker::recvMessage(int groupId,
 
     Message message = recvEndpoints[label]->recv();
```
Shillaker, line 673: This message will be the body message, but the sequence number is in the header. We can add a bit in the transport layer to set it on the body.

src/transport/PointToPointBroker.cpp
```diff
@@ -566,10 +672,74 @@ std::vector<uint8_t> PointToPointBroker::recvMessage(int groupId,
...
+        if (seqNum == expectedSeqNum) {
+            returnData = std::move(recvData);
+            break;
+        }
+
+        // If not, we must insert the received message in the out of order
+        // received messages. We must also check if the sequence number we want
+        // is in the list.
+
+        // First insert
```
Shillaker, line 712: Fast path:

- Avoid message copy by taking ownership.
- Do `outOfOrderMsgs.emplace_back(message)`.
- Retrieval code will have a special case, `if(outOfOrderMsgs.back().seqNum == expectedSeqNum) { // it's the one we want }`.
- Otherwise do a simple iteration over the list, check each `msg.seqNum`.

src/transport/PointToPointBroker.cpp
```diff
@@ -566,10 +672,74 @@ std::vector<uint8_t> PointToPointBroker::recvMessage(int groupId,
...
+    }
+
+    int expectedSeqNum = getExpectedSeqNum(groupId, sendIdx);
+    std::vector<uint8_t> recvData;
+    std::vector<uint8_t> returnData;
+
+    // Loop to receive the right sequence number
+    while (true) {
+        // First, receive from the transport layer
+        recvData = doRecvMessage(groupId, sendIdx, recvIdx);
```
Shillaker, line 694: We want to get the message here, and avoid a copy.

src/transport/PointToPointBroker.cpp
```diff
@@ -566,10 +672,74 @@ std::vector<uint8_t> PointToPointBroker::recvMessage(int groupId,
...
+        // First insert
+        auto insertIt = std::find_if(
+          outOfOrderMsgs.begin(),
+          outOfOrderMsgs.end(),
+          [&seqNum](const std::pair<int, std::shared_ptr<std::vector<uint8_t>>>&
+                      element) { return element.first > seqNum; });
+        outOfOrderMsgs.emplace(
+          insertIt,
+          std::make_pair(seqNum,
+                         std::make_shared<std::vector<uint8_t>>(recvData)));
```
Shillaker, line 721: Double check, but I think this is a copy, e.g. if you change `recvData`, the content of the pointer target won't change.

src/transport/PointToPointBroker.cpp
csegarragonz: I also just realised that we need to initialise both sequence counters at the same time.

In the general case when we are sending and receiving ordered messages, the `groupId != currentGroupId` check will only fail once per thread and group id. So regardless of whether we hit it sending or receiving, we need to reset all coutners.

src/transport/PointToPointBroker.cpp
csegarragonz: Note that if we include the sequence number in the message header, there's no need to modify the `PointToPointMessage` protobuf object, as we can just add the sequence number in the actual message.

tests/dist/transport/functions.cpp
```diff
@@ -62,6 +62,158 @@ int handlePointToPointFunction(
...
+        }
+    } else {
+        SPDLOG_ERROR("Unexpected group index: {}", groupIdx);
+        return 1;
+    }
+
+    return 0;
+}
+
+void doExperiment(std::map<std::string, std::vector<double>>& results,
```
csegarragonz, line 111: I include the benchmark in the PR. Maybe this is not the right place to have it though, and happy to remove it.

src/transport/PointToPointBroker.cpp
```diff
@@ -564,10 +643,54 @@ std::vector<uint8_t> PointToPointBroker::recvMessage(int groupId,
...
+    if (wantedIt != outOfOrderMsgs.end()) {
+        incrementRecvMsgCount(groupId, sendIdx);
+        Message returnMsg = std::move(*wantedIt);
+        outOfOrderMsgs.erase(wantedIt);
+        return returnMsg.dataCopy();
+    }
+
+    // Given that we don't have the message, we query the transport layer until
+    // we receive it
+    while (true) {
```
Shillaker, line 679: `while(true)` loops hanging are quite difficult to debug if there's no logging. I think we should have some trace logging on entering the while loop and receiving out-of-order messages.

include/faabric/transport/PointToPointBroker.h
```diff
@@ -128,6 +131,18 @@ class PointToPointBroker
       groupFlags;
 
     std::shared_ptr<faabric::util::FlagWaiter> getGroupFlag(int groupId);
+
+    Message doRecvMessage(int groupId, int sendIdx, int recvIdx);
+
+    std::atomic<bool> orderMsg;
```
Shillaker, line 137: Why does `orderMsg` need to be an `atomic<bool>` here? Are we modifying it from multiple threads?

In general it's best to name bools with getters and setters as `_isX` (variable), `isX()` (getter) and `setX()` (setter), in this case the following might be clearer:

- `_isMessageOrderingOn` (variable)
- `isMessageOrderingOn` (getter)
- `setMessageOrderingOn` (setter)

src/transport/PointToPointBroker.cpp
```diff
@@ -502,16 +518,73 @@ std::set<int> PointToPointBroker::getIdxsRegisteredForGroup(int groupId)
...
+void PointToPointBroker::initSequenceCounters(int groupId)
+{
+    if (currentGroupId != NO_CURRENT_GROUP_ID) {
+        SPDLOG_WARN("Changing the current group Id in PTP broker ({} -> {})",
+                    currentGroupId,
+                    groupId);
+    }
+    currentGroupId = groupId;
+    int groupSize = getIdxsRegisteredForGroup(groupId).size();
+    // We assume that group indexes will always go from 0 to `groupSize - 1`
```
Shillaker, line 530: This will always be true, but writing "we assume" here, suggests that there may not be groups for which this is the case. I think it's clearer not to have this line at all.

src/transport/PointToPointBroker.cpp
```diff
@@ -502,16 +518,73 @@ std::set<int> PointToPointBroker::getIdxsRegisteredForGroup(int groupId)
...
+    // We also want to start counting from one. This way, an unset sequence
+    // number (0 by default in protobuf) will indicate that it is unset.
```
Shillaker, line 532: "This way, an unset sequence number (0 by default in protobuf) will indicate that it is unset." this seems like a tautology. When do we need to distinguish between unset values and set values? It seems like the `orderMsg` bool should tell us whether we should pay attention to sequence numbers.

src/transport/PointToPointBroker.cpp
```diff
@@ -564,10 +643,54 @@ std::vector<uint8_t> PointToPointBroker::recvMessage(int groupId,
...
+
+        // If the sequence numbers match, exit the loop
+        int seqNum = recvMsg.getSequenceNum();
+        if (seqNum == expectedSeqNum) {
+            incrementRecvMsgCount(groupId, sendIdx);
+            return recvMsg.dataCopy();
+        }
+
+        // If not, we must insert the received message in the out of order
+        // received messages.
```
Shillaker, line 691: Inconsistent use of full stops at the end of comments here. Style should be no full stop.

tests/dist/transport/test_point_to_point.cpp
```diff
@@ -69,6 +69,60 @@ TEST_CASE_METHOD(DistTestsFixture,
...
+        faabric::Message& msg = req->mutable_messages()->at(i);
+
+        msg.set_appid(appId);
+        msg.set_appidx(i);
+        msg.set_groupid(groupId);
+        msg.set_groupidx(i);
+
+        // Add to expected decision
+        expectedDecision.addMessage(expectedHosts.at(i), req->messages().at(i));
+    }
```
Shillaker, line 111: Most of this function looks very similar to the other test above, could you refactor to share the logic rather than duplicate?

tests/test/transport/test_point_to_point.cpp
```diff
@@ -143,6 +144,85 @@ TEST_CASE_METHOD(PointToPointClientServerFixture,
...
+    msgB.set_appid(appId);
+    msgB.set_groupid(groupId);
+    msgB.set_groupidx(idxB);
+
+    decision.addMessage(LOCALHOST, msgA);
+    decision.addMessage(LOCALHOST, msgB);
+
+    // Set up the mappings and configure in-order delivery
+    broker.setAndSendMappingsFromSchedulingDecision(decision);
+    broker.setMustOrderMessages(true);
```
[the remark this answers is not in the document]
Shillaker: I would expect this test to have two sections, one with ordering on, one with ordering off, then check that the sender is setting the sequence numbers right (this can still be checked without the server being used right?).

tests/dist/transport/functions.cpp
```diff
@@ -62,6 +62,52 @@ int handlePointToPointFunction(
...
+{
+    faabric::Message& msg = req->mutable_messages()->at(msgIdx);
+
+    int groupId = msg.groupid();
+    uint8_t groupIdx = (uint8_t)msg.groupidx();
+    int numMsg = 10000;
+
+    // Set the broker to receive messages in order
+    auto& broker = faabric::transport::getPointToPointBroker();
+    broker.setMustOrderMessages(true);
```
Shillaker, line 79: AFAICT this doesn't get set back to `false` anywhere. It should really do the following:

```cpp
bool originalMessageOrderingOn = broker.isMessageOrderingOn();
broker.setMessageOrderingOn(true);

...

broker.setMessageOrderingOn(originalMessageOrderingOn);
```

include/faabric/transport/MessageEndpoint.h
```diff
@@ -86,7 +91,10 @@ class AsyncSendMessageEndpoint final : public MessageEndpoint
                              int portIn,
                              int timeoutMs = DEFAULT_SOCKET_TIMEOUT_MS);
 
-    void send(uint8_t header, const uint8_t* data, size_t dataSize);
+    void send(uint8_t header,
+              const uint8_t* data,
+              size_t dataSize,
+              int sequenceNum = -1);
```
Shillaker, line 97: This value of `-1` to signify an unset sequence number is used in a few places as a magic number. It would be good to define it as a constant.

Shillaker (changes requested):

src/transport/PointToPointBroker.cpp
csegarragonz: You are right, the comment and check were left behind from a previous version of the PR.

tests/test/transport/test_point_to_point.cpp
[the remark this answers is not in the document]
csegarragonz: Yes, the sequence numbers are implicitly checked by running the tests with the ordering on. Do you mean explicitly checking for the sequence number in the test? 

Irrespective of your answer, I add the requested section in the test.

include/faabric/transport/MessageEndpointClient.h
```diff
@@ -14,9 +14,14 @@ class MessageEndpointClient
...
-
-    void asyncSend(int header, const uint8_t* buffer, size_t bufferSize);
+    void asyncSend(int header,
+                   google::protobuf::Message* msg,
+                   int sequenceNum = -1);
+
+    void asyncSend(int header,
+                   const uint8_t* buffer,
+                   size_t bufferSize,
+                   int sequenceNum = -1);
```
Shillaker, line 24: Could use constant here instead of the magic number for `-1`.

include/faabric/transport/PointToPointBroker.h
```diff
@@ -106,7 +106,8 @@ class PointToPointBroker
                      int sendIdx,
                      int recvIdx,
                      const uint8_t* buffer,
-                     size_t bufferSize);
+                     size_t bufferSize,
+                     int sequenceNum = -1);
```
Shillaker, line 110: Constant vs. `-1`

include/faabric/transport/PointToPointClient.h
```diff
@@ -26,7 +26,7 @@ class PointToPointClient : public faabric::transport::MessageEndpointClient
 
     void sendMappings(faabric::PointToPointMappings& mappings);
 
-    void sendMessage(faabric::PointToPointMessage& msg);
+    void sendMessage(faabric::PointToPointMessage& msg, int sequenceNum = -1);
```
Shillaker, line 29: Magic number.

src/transport/PointToPointBroker.cpp
```diff
@@ -502,16 +518,70 @@ std::set<int> PointToPointBroker::getIdxsRegisteredForGroup(int groupId)
     return groupIdIdxsMap[groupId];
 }
 
+void PointToPointBroker::initSequenceCounters(int groupId)
+{
+    if (currentGroupId != NO_CURRENT_GROUP_ID) {
+        SPDLOG_WARN("Changing the current group Id in PTP broker ({} -> {})",
+                    currentGroupId,
+                    groupId);
```
Shillaker, line 526: Should this be a warn? It looks like it's the expected behaviour when the ID changes, could it be a `DEBUG` instead?

src/transport/PointToPointBroker.cpp
```diff
@@ -611,6 +746,13 @@ void PointToPointBroker::resetThreadLocalCache()
     clients.clear();
 }
 
+bool PointToPointBroker::setIsMessageOrderingOn(bool mustOrderMsgs)
+{
+    bool oldValue = _isMessageOrderingOn.load(std::memory_order_acquire);
+    _isMessageOrderingOn.store(mustOrderMsgs, std::memory_order_release);
+    return oldValue;
```
Shillaker, line 753: Can we use `exchange` here rather than `load` and `store`?

Shillaker (changes requested):
Couple of minor points but other than that LGTM

include/faabric/transport/MessageEndpointClient.h
csegarragonz: Sorry, some of the -1s slipped under the radar

src/transport/PointToPointBroker.cpp
csegarragonz: I was not aware of `exchange`, thanks.

Shillaker (approved):

PR #263, opened 2022-05-09
Clear Point-to-Point broker

csegarragonz:
In this PR I add support to clean the thread-local and shared records in the Point-to-Point broker as part of the executor cleanup and scheduler reset respectively.

### DIFF
--- a/src/scheduler/Executor.cpp
+++ b/src/scheduler/Executor.cpp
@@ -584,11 +584,12 @@ void Executor::threadPoolThread(std::stop_token st, int threadPoolIdx)
             } else {
                 reset(msg);
             }
-            // Clear the thread local PTP cache
-            faabric::transport::getPointToPointBroker().resetThreadLocalCache();
 
             releaseClaim();
         }
+
+        // Clear the thread local PTP cache
+        faabric::transport::getPointToPointBroker().resetThreadLocalCache();
 
         // Return this thread index to the pool available for scheduling
         {
--- a/src/scheduler/Executor.cpp
+++ b/src/scheduler/Executor.cpp
@@ -584,6 +584,8 @@ void Executor::threadPoolThread(std::stop_token st, int threadPoolIdx)
             } else {
                 reset(msg);
             }
+            // Clear the thread local PTP cache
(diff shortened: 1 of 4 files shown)


src/scheduler/Executor.cpp
```diff
@@ -588,6 +588,9 @@ void Executor::threadPoolThread(std::stop_token st, int threadPoolIdx)
             releaseClaim();
         }
 
+        // Clear the thread local PTP cache
```
csegarragonz, line 591: I am clearing the thread local cache every time an execuing thread is done executing a task. This feels like the right choice for non-threads, but it may not be for thread execution (?).

Shillaker: I don't think this is a good idea, each executor is either repeatedly executing jobs from the same application (in which case the cache is useful and we want it to persist), or will automatically get cleared up by the scheduler reaping process, or will be cleared from an explicit flush operation. Its associated TLS will automatically get cleaned up when the underlying thread is removed in the last two cases.

Unless it's definitely needed to avoid resource exhaustion, I don't see what benefit this adds, and think it's sufficient just to add the clean-up to the scheduler. I would be surprised if it was this TLS that's the problem (i.e. I suspect it's the stuff that's outside the executor lifecycle that builds up but may be wrong).

tests/test/transport/test_point_to_point.cpp
```diff
@@ -160,6 +160,13 @@ TEST_CASE_METHOD(PointToPointClientServerFixture,
     REQUIRE(receivedDataB == sentDataB);
     REQUIRE(receivedDataC == sentDataC);
 
+    // Clear local records
+    broker.resetThreadLocalCache();
+
+    // Clear broker
+    broker.clear();
+    REQUIRE(broker.getIdxsRegisteredForGroup(groupId).empty());
+
```
Shillaker, line 169: I would expect to see a stand-alone test for this change that confirms that the ptp broker is cleared when we perform a scheduler reset.

Shillaker (changes requested):

Shillaker (approved):

PR #270, opened 2022-09-08
Race Condition

mmathys:
Update `Scheduler.cpp` such that the race condition doesn't occur anymore.


csegarragonz:
Superseeded by #288

PR #271, opened 2022-09-28
Bump base to `ubuntu:22.04`

csegarragonz:
In this PR I bump `faabric-base`'s parent image from `ubuntu:20.04` to `ubuntu:22.04`. The most notable changes are:
* Remove the use of `apt-key add` as it is marked as deprecated in 22.04. See [here](https://askubuntu.com/questions/1286545/what-commands-exactly-should-replace-the-deprecated-apt-key) for more details.

Update: this PR is in draft until the SGX-SDK supports 22.04 (next release). In the meantime, I port some of the other changes to #272


bin/create_venv.sh
```diff
@@ -0,0 +1,32 @@
...
+}
+
+pushd ${PROJ_ROOT} >> /dev/null
+
+if [ ! -d ${VENV_PATH} ]; then
+    python3 -m venv ${VENV_PATH}
+fi
+
+pip_cmd install -U pip
+pip_cmd install -r requirements.txt
```
eigenraven, line 30: `wheel` should be installed before requirements.txt, otherwise installing packages via requirements falls back to a much slower path

csegarragonz: Yeah, thanks for pointing this out! I will request a review from you now, see if you spot something else.

docker/faabric-base.dockerfile
```diff
@@ -1,18 +1,48 @@
...
+    && apt upgrade -y \
+    && apt install -y \
+        curl \
+        gpg \
+        software-properties-common \
+        wget \
+    # apt-key add is now deprecated for security reasons, we add individual
+    # keys manually.
+    # https://askubuntu.com/questions/1286545/what-commands-exactly-should-replace-the-deprecated-apt-key
+    && wget -O /tmp/llvm-snapshot.gpg.key \
```
eigenraven, line 14: Installing llvm and cmake from 20.04 repositories in 22.04 can be prone to breakage, apt sometimes starts removing the wrong packages if you force it to resolve versions from mixed distributions like this. For LLVM, 13 and 11 are in the base 22.04 repositories, there might be no other option to install llvm 10. For cmake and conan, I switched to installing them via pip in my containers as both have pip packages, and that also made them installable on non-x86 architectures (but I don't think that's important for your usecases).
Everything else looks fine to me.

csegarragonz: Ok. I realised LLVM 10 was not in 22.04, so I reverted to instaling _both_ 10 and 13 from 20.04 repositories. Thanks for pointing this out.

eigenraven (approved):

csegarragonz:
Drafting the PR, as the SGX-SDK is still not compatible with 22.04, so porting some of the changes to a separate PR.

eigenraven:
FYI a new SGX sdk was just released with 22.04 support <https://github.com/intel/linux-sgx/releases/tag/sgx_2.18>

PR #274, opened 2022-10-07
Asio+Beast-based endpoint server

eigenraven:
Based off my current code in <https://github.com/auto-ndp/faabric>, this implementation of the faabric endpoint replaces the pistache-based endpoint with an asynchronous one based on boost::beast.
This lets the local endpoint handle many more requests per second with fewer threads, as the threads should no longer be waiting for executors to finish execution before handling more HTTP requests.

One exception to this is calls that are forwarded to other machines, as the redis calls are still synchronous (meaning it will behave as bad as without this PR) - a separate asynchronous redis fetching implementation is needed. People have built asio wrappers around hiredis's interface before, so it's just a matter of writing the code.


src/endpoint/FaabricEndpoint.cpp
```diff
@@ -1,92 +1,261 @@
...
+            SPDLOG_ERROR("Error reading an HTTP request: {}", ec.message());
+            return;
+        }
+        SPDLOG_TRACE("Read HTTP request of {} bytes", bytesTransferred);
+        handleRequest(parser.release());
+    }
+
+    void sendResponse(faabric::util::BeastHttpResponse&& response)
+    {
+        // response needs to be freed after the send completes
```
csegarragonz, line 91:
```suggestion
        // Response needs to be freed after the send completes
```

src/endpoint/FaabricEndpoint.cpp
```diff
@@ -1,92 +1,261 @@
...
+        if (ec) {
+            SPDLOG_ERROR("Couldn't write HTTP response: {}", ec.message());
+            return;
+        }
+        SPDLOG_TRACE("Write HTTP response of {} bytes", bytesTransferred);
+        if (needsEof) {
+            doClose();
+            return;
+        }
+        // reset parser to a fresh object, it has no copy/move assignment
```
csegarragonz, line 118:
```suggestion
        // Reset parser to a fresh object, it has no copy/move assignment
```

src/endpoint/FaabricEndpoint.cpp
```diff
@@ -1,92 +1,261 @@
...
+        parser.~parser();
+        new (&parser) decltype(parser)();
+        doRead();
+    }
+
+    void doClose()
+    {
+        beast::error_code ec;
+        stream.socket().shutdown(asio::socket_base::shutdown_send, ec);
+        // ignore errors on connection closing
```
csegarragonz, line 128:
```suggestion
        // Ignore errors on connection closing
```

src/endpoint/FaabricEndpoint.cpp
```diff
@@ -1,92 +1,261 @@
...
+                   asio::ip::tcp::socket&& socket,
+                   std::shared_ptr<HttpRequestHandler> handlerIn)
+      : ioc(iocIn)
+      , stream(std::move(socket))
+      , buffer()
+      , parser()
+      , handler(handlerIn)
+    {}
+
+    void run()
```
csegarragonz, line 45: I am a bit confused, why should an `HttpConnection` have a `run()` method?

src/endpoint/FaabricEndpoint.cpp
```diff
@@ -1,92 +1,261 @@
...
+
+    void doClose()
+    {
+        beast::error_code ec;
+        stream.socket().shutdown(asio::socket_base::shutdown_send, ec);
+        // ignore errors on connection closing
+    }
+};
+
+class EndpointListener : public std::enable_shared_from_this<EndpointListener>
```
csegarragonz, line 132: If we rename to `FaabricHttpEndpoint` we can rename this to `HttpEndpointListener`.

Can you also add a little comment on top of the class explaining what it does?

src/endpoint/FaabricEndpoint.cpp
```diff
@@ -1,92 +1,261 @@
...
+}
 
-FaabricEndpoint::FaabricEndpoint(int portIn, int threadCountIn)
-  : port(portIn)
-  , threadCount(threadCountIn)
-  , httpEndpoint(
-      Pistache::Address(Pistache::Ipv4::any(), Pistache::Port(portIn)))
-{}
+namespace {
+class HttpConnection : public std::enable_shared_from_this<HttpConnection>
```
csegarragonz, line 26: Could you add a comment explaiing what the class does?

src/endpoint/FaabricEndpoint.cpp
```diff
@@ -1,92 +1,261 @@
...
-    httpEndpoint.serveThreaded();
+    const auto address = asio::ip::make_address_v4("0.0.0.0");
+    const auto port = static_cast<uint16_t>(this->port);
+
+    std::make_shared<EndpointListener>(state->ioc,
+                                       asio::ip::tcp::endpoint{ address, port },
+                                       this->requestHandler)
+      ->run();
+
+    std::optional<asio::signal_set> signals;
```
csegarragonz, line 229: These two bits of code could do with some comments.

src/endpoint/FaabricEndpoint.cpp
csegarragonz: OK, I see that an `HttpConnection` is what we run on `accept`. I don't know if `run` is the best name, but maybe with a bit of explanation at the begining of the class I would have been all right.

src/scheduler/Scheduler.cpp
```diff
@@ -1314,6 +1335,103 @@ faabric::Message Scheduler::getFunctionResult(unsigned int messageId,
     return msgResult;
 }
 
+void Scheduler::getFunctionResultAsync(
```
csegarragonz, line 1338: Do we need to keep the old `getFunctionResult`?

src/scheduler/Scheduler.cpp
```diff
@@ -1314,6 +1335,103 @@ faabric::Message Scheduler::getFunctionResult(unsigned int messageId,
...
+        {
+            faabric::util::UniqueLock resultsLock(localResultsMutex);
+            auto it = localResults.find(messageId);
+            if (it == localResults.end()) {
+                break; // Fallback to redis
+            }
+            mlr = it->second;
+        }
+        // Asio wrapper for the MLR eventfd
+        struct MlrAwaiter : public std::enable_shared_from_this<MlrAwaiter>
```
csegarragonz, line 1361: `class`?

csegarragonz (commented):
Thanks for taking the time to upstream this! I've left some comments.

More generally, is this a replacement for the syncrhonous `getFunctionResult` method in the scheduler? Can we delete that one?

Also, does this change the format of the HTTP response? This is, do we need to change the HTTP response consumers code?

include/faabric/endpoint/FaabricEndpoint.h
```diff
@@ -1,37 +1,61 @@
...
+  public:
+    virtual void onRequest(HttpRequestContext&& ctx,
+                           faabric::util::BeastHttpRequest&& request) = 0;
+};
+
 class FaabricEndpoint
 {
   public:
     FaabricEndpoint();
+    FaabricEndpoint(
```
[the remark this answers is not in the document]
eigenraven: If that's what you want, I could include this refactor, but my goal with this PR was to not break the faabric API as used by faasm.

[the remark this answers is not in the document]
csegarragonz: That's a very good point

csegarragonz (approved):
LGTM. Let's wait for Simon

Shillaker (approved):
Awesome, this looks great, thanks @eigenraven. Just one small question in the CMake file around removing pistache completely.

cmake/ExternalProjects.cmake
```diff
@@ -120,13 +119,13 @@ set(ZSTD_LZ4_SUPPORT OFF CACHE INTERNAL "")
```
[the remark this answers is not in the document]
eigenraven: Looks like it, I didn't notice you rewrote the http test utils to use curl instead of pistache.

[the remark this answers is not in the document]
csegarragonz: @eigenraven Let me know when you are done with it and I will merge this in.

PR #282, opened 2022-11-02
Fix clang-10 formatting check

csegarragonz:
I realised that the `format-code` task wasn't working for C++ code with the `--check` flag. In this PR I fix the issue and now, with the `--check` flag,  `clang-format-10` returns an error if formatting changes are needed.

### DIFF
--- a/tasks/format_code.py
+++ b/tasks/format_code.py
@@ -51,7 +51,12 @@ def format(ctx, check=False):
         .split("\n")[:-1]
     )
 
-    clang_cmd = "clang-format-10 -i {}".format(" ".join(files_to_check))
+    if check:
+        clang_cmd = "clang-format-10 --dry-run --Werror {}".format(
+            " ".join(files_to_check)
+        )
+    else:
+        clang_cmd = "clang-format-10 -i {}".format(" ".join(files_to_check))
     run(clang_cmd, shell=True, check=True, cwd=PROJ_ROOT)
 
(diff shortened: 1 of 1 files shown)


tasks/format_code.py
```diff
@@ -51,7 +51,12 @@ def format(ctx, check=False):
         .split("\n")[:-1]
     )
 
-    clang_cmd = "clang-format-10 -i {}".format(" ".join(files_to_check))
+    if check:
+        clang_cmd = "clang-format-10 --dry-run --Werror {}".format(
+            " ".join(files_to_check)
+        )
+    else:
+        clang_cmd = "clang-format-10 -i {}".format(" ".join(files_to_check))
```
Shillaker, line 59: Could be cleaner as:

```python
clang_cmd = [
    "clang-format-10",
    "--dry-run --Werror" if check else "-i",
    files_to_check,
]
clang_cmd = " ".join(clang_cmd)
```

Is the `-i` not needed in the dry run command too though?

Shillaker (changes requested):

tasks/format_code.py
csegarragonz: Agree, your suggestion makes it cleaner. Afaict `-i` stands for "modify (files) in place" whereas `--dry-run` means "check for formatting changes, but don't apply any", so we don't need the `-i` in the `--dry-run`.

Shillaker (approved):

Issue #199, opened 2021-12-15
Switch from using `uint8_t` to `std::byte`


Shillaker:
[`std::byte`](https://en.cppreference.com/w/cpp/types/byte) is the standard way to represent raw memory, which we currently do with `unit8_t`. 

`std::byte` avoids the temptation to use any character or arithmetic semantics, and should thus reduce the risk of dodgy memory manipulation that may otherwise be possible.

PR #284, opened 2022-11-15
Replace VLA usage with heap-allocated buffers

eigenraven:
Variable length arrays are a compiler extension and not part of the base c++ language, they cause dynamic stack allocations which can silently overflow the stack, causing hard-to-debug errors at runtime which I have ran into a couple of times with large network messages.

The macros.h macros were only used once, so I just replaced the macro with its contents.
This has the additional benefit that SerializeToString (string is used as a bytes container in protobuf, not text) will only compute the size of the message once unlike the previous approach which computed it twice while serializing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@ else()
 endif()
 
 # Top-level CMake config
-set(CMAKE_CXX_FLAGS "-Wall")
+set(CMAKE_CXX_FLAGS "-Wall -Werror=vla")
 set(CMAKE_CXX_FLAGS_DEBUG "-g")
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
--- a/include/faabric/transport/macros.h
+++ b/include/faabric/transport/macros.h
@@ -6,20 +6,6 @@
         throw std::runtime_error("Error deserialising message");               \
     }
 
-#define SERIALISE_MSG(_msg)                                                    \
-    size_t serialisedSize = _msg.ByteSizeLong();                               \
-    uint8_t serialisedBuffer[serialisedSize];                                  \
-    if (!_msg.SerializeToArray(serialisedBuffer, serialisedSize)) {            \
-        throw std::runtime_error("Error serialising message");                 \
-    }
-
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ else()
 endif()
 
 # Top-level CMake config
-set(CMAKE_CXX_FLAGS "-Wall")
+set(CMAKE_CXX_FLAGS "-Wall -Werror=vla")
```
csegarragonz, line 26: So, what does this do exactly? And what errors does it help prevent?

csegarragonz (approved):
LGTM. Just one question as I did not understand the PR description.

CMakeLists.txt
eigenraven: It makes creating a variable-length array a compiler error to prevent any more VLAs sneaking in in the future, e.g.:
```
/usr/local/code/faasm/tests/test/threads/test_levels.cpp:44:20: error: variable length arrays are a C99 feature [-Werror,-Wvla-extension]
    uint8_t buffer[bufferSize];
                   ^~~~~~~~~~
```

csegarragonz: ah! now i get it. thanks for that.

eigenraven: They're easy to miss, as a VLA is created any time you give a non-constant expression as the array size in a `type name[size];` declaration. `std::array<type, size>` won't accept non-constant sizes in the first place and it makes it easy to access the size of the array via `var.size()` so it's also a good replacement.

csegarragonz: Yes, its good that we can enforce this from the compiler. This reminded me of #199. This PR is not the place to change it, but do you think its worth the re-factor `uint8_t -> std::byte`?

I'd like to progressively chip away on the issues that have been sitting there for months.

eigenraven: Yes, it won't impact correctness in any way so it's not high priority but it will make it clearer which variables hold raw memory areas and make it less likely to make mistakes accessing them as 8-bit integers. One downside is that the library apis mostly don't use it so there'll have to be casts at that boundary.

Shillaker (approved):
Very nice :+1:

PR #294, opened 2023-02-22
Reset executor when a function invokes threads

csegarragonz:
In a request of type `THREADS`, the "caller" (or main) thread (the thread that calls `executeThreads`, which is _not_ part of the `THREADS` request) [executes in the same `Executor`](https://github.com/faasm/faabric/blob/main/src/scheduler/Scheduler.cpp#L818-L843) than the "callee" threads (the threads spawned as a consequence of `executeThreads`, i.e. the `THREADS` request).

This means that the `batchCounter` atomic counter in the `Executor` will take different values:
* First, when the "caller"/main thread is invoked, it will take value 1.
* Second, when the main thread issues a `THREADS` request, it will take the value of the size of the `THREADS` request.

This is because we were overwritting the value of the atomic counter with a `store()`.

This is problematic because we use `batchCounter` to define the variable `isLastInBatch` that we use to determine when a batch is finished. A finished batch can either be:
i. The batch corresponding to the main thread.
ii. The batch corresponding to the `THREADS` request.

Each finished batch requires a different cleanup:
i. When the main thread finishes we need to `reset()` and release the claim on the the executor.
ii. When a `THREADS` batch finishes, we need to synchronise the diffs and set the results.

In addition, if the batch is a `THREADS` request and is not executing in the main host, it means that we also need to `release()` (but not `reset()`) the executor.

At the moment we are only doing (ii) and not (i). I add a regression test that fails against the current `main` branch in faabric.

### DIFF
--- a/src/scheduler/Executor.cpp
+++ b/src/scheduler/Executor.cpp
@@ -206,7 +206,7 @@ void Executor::executeTasks(std::vector<int> msgIdxs,
                             std::shared_ptr<faabric::BatchExecuteRequest> req)
 {
     const std::string funcStr = faabric::util::funcToString(req);
-    SPDLOG_ERROR("{} executing {}/{} tasks of {} (single-host={})",
+    SPDLOG_TRACE("{} executing {}/{} tasks of {} (single-host={})",
                  id,
                  msgIdxs.size(),
                  req->messages_size(),
@@ -438,7 +438,7 @@ void Executor::threadPoolThread(std::stop_token st, int threadPoolIdx)
               faabric::transport::PointToPointGroup::getGroup(msg.groupid());
         }
 
-        SPDLOG_WARN("Thread {}:{} executing task {} ({}, thread={}, group={})",
+        SPDLOG_TRACE("Thread {}:{} executing task {} ({}, thread={}, group={})",
                      id,
                      threadPoolIdx,
                      task.messageIndex,
--- a/src/scheduler/Executor.cpp
+++ b/src/scheduler/Executor.cpp
@@ -206,7 +206,7 @@ void Executor::executeTasks(std::vector<int> msgIdxs,
                             std::shared_ptr<faabric::BatchExecuteRequest> req)
 {
     const std::string funcStr = faabric::util::funcToString(req);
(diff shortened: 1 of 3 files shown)


src/scheduler/Executor.cpp
```diff
@@ -497,14 +501,24 @@ void Executor::threadPoolThread(std::stop_token st, int threadPoolIdx)
...
+        bool isLastThreadInExecutor = false;
+        if (isThreads) {
+            oldTaskCount = threadBatchCounter.fetch_sub(1);
+            isLastThreadInBatch = oldTaskCount == 1;
+            isLastThreadInExecutor = batchCounter.load(std::memory_order_release) == 0;
+        } else {
+            oldTaskCount = batchCounter.fetch_sub(1);
+            isLastThreadInExecutor = oldTaskCount == 1;
+        }
+        assert(oldTaskCount >= 1);
```
csegarragonz, line 521: `taskCount` should be > 0, thus `oldTaskCount` should always be greater than `1`. This assertion would have caught the bug.

src/scheduler/Executor.cpp
```diff
@@ -558,9 +572,9 @@ void Executor::threadPoolThread(std::stop_token st, int threadPoolIdx)
         }
 
         // If this is not a threads request and last in its batch, it may be
-        // the main function in a threaded application, in which case we
-        // want to stop any tracking and delete the main thread snapshot
-        if (!isThreads && isLastInBatch) {
+        // the main function (thread) in a threaded application, in which case
```
csegarragonz, line 575: The word thread may be a bit overloaded in this context.

tests/test/scheduler/test_executor.cpp
```diff
@@ -1018,17 +1018,25 @@ TEST_CASE_METHOD(TestExecutorFixture,
...
     {
         requestType = faabric::BatchExecuteRequest::THREADS;
         nMessages = 3;
         expectedResets = 0;
     }
 
-    SECTION("Function") {}
+    SECTION("Simple function") {}
+
+    SECTION("Function that spawns threads")
```
csegarragonz, line 1033: This section fails against the `main` branch

PR #300, opened 2023-04-04
Centralised Control-Plane

csegarragonz:


include/faabric/util/json.h
```diff
@@ -3,7 +3,27 @@
 #include <faabric/proto/faabric.pb.h>
 #include <faabric/util/exception.h>
 
+#include <google/protobuf/util/json_util.h>
+
 namespace faabric::util {
+void messageToJsonPb(const google::protobuf::Message& msg,
```
csegarragonz, line 9: These two methods use protobuf's embedded JSON (de)-serialisation.

It is a very big improvement in terms of LoC and complexity versus the current RapidJSON approach.

I implement it here as the RapidJSON approach only works for `faabric::Message` whereas protobuf's works for any class inheriting from `google::protobuf::Message`.

However, removing the old version is not immediate as _all_ CLI and experiment code relies on the syntax we define when (de)-serialising messages to JSON, so I'd rather add this here, and remove RapidJSON in a separate PR.

include/faabric/endpoint/FaabricEndpoint.h
```diff
@@ -38,10 +38,9 @@ class FaabricEndpoint
   public:
     FaabricEndpoint();
 
-    FaabricEndpoint(
-      int port,
-      int threadCount,
-      std::shared_ptr<HttpRequestHandler> requestHandlerIn = nullptr);
+    FaabricEndpoint(int port,
```
csegarragonz, line 41: Before, we allowed users of the `FaabricEndpoint` not to specify a request handler, and we would use `FaabricEndpointHandler` by default.

Unfortunately, this means that any other class implementing a different endpoint (thus using `FaabricEndpoint` but a different endpoint handler than `FaabricEndpointHandler`) still has a link-time dependency to `FaabricEndpointHandler` (i.e. needs to provide all the symbols used by `FaabricEndpointHandler`). In turn, this means that we need to link with `faabric::scheduler` even if not using any symbols therein defined.

The aim of this PR is to split from the scheduler (eventually turning the existing scheduler into a form of local scheduler), so we don't want to take that link-time dependency.

tests/utils/faabric_utils.h
```diff
@@ -2,7 +2,7 @@
 
 #include <catch2/catch.hpp>
 
-#include "fixtures.h"
```
csegarragonz, line 5: Throughout the tests we included "faabric_utils.h" to use the test fixtures. However, the fixtures are defined in "fixtures.h", and it only workerd because we had this include here, which we actually don't need for "faabric_utils.h".

.github/workflows/tests.yml
```diff
@@ -71,33 +71,44 @@ jobs:
...
-      credentials:
-        username: ${{ secrets.ACR_SERVICE_PRINCIPAL_ID }}
-        password: ${{ secrets.ACR_SERVICE_PRINCIPAL_PASSWORD }}
-      options: --privileged
-    services:
-      redis:
-        image: redis
+      UBSAN_OPTIONS: print_stacktrace=1:halt_on_error=1
     steps:
+      - name: "Install Conan"
```
csegarragonz, line 84: I move the `tests` job from running inside a container with additional services to using `docker compose` manually.

The `planner` runs in a separate container, but may be updated frequently. To avoid having to re-tag the code everytime we change anything in the planner, we want to mount the binaries we build as part of the tests (with the latest checkout of the code) and use those to run the planner.

The first problem this arises is that, even though GHA's service containers support mounting volumes, the planner service requires the `planner_server` binary to be there before it can start. And once it is running it can not be restarted from _inside_ the tests container.

To overcome this problem, we add a custom entrypoint for the planner that allows setting the directory where it will look for the `planner_server` binary. If its not there, it will wait until it becomes available.

Then, we come to a second problem. GHA's service containers don't allow overwriting the entrypoint. See [here](https://github.com/orgs/community/discussions/26688) for a further discussion.

The solution I have found is to use `docker compose` manually.

Shillaker (approved):

PR #301, opened 2023-04-11
Re-factor tests includes

csegarragonz:
To use test fixtures, we currently include `faabric_utils.h` around, instead of including `fixtures.h`. This works because in the former we also include `fixtures.h`. However, `fixtures.h` is actually _not_ needed in `faabric_utils.h`, and removing it from there breaks the compilation of all the tests.

In addition, this was making the file `tests/dist/fixtures.h` use `tests/util/fixtures.h` transitively through `faabric_utils.h`. This was silently overwriting the latter with the former, what could have eventually caused compilation errors.

In this PR, I re-factor the includes in the tests to:
1. Include `fixtures.h` instead of `faabric_utils.h` when we want to use the fixtures.
2. Follow the include order we soft-enforce across the codebase.
3. Re-factor `tests/dist/fixtures.h` to `tests/dist/dist_test_fixtures.h`.

I understand that we currently don't enforce 2 strongly anywhere, but I figured out that I would still amend it here nontheless.

### DIFF
--- a/tests/utils/faabric_utils.h
+++ b/tests/utils/faabric_utils.h
@@ -2,8 +2,6 @@
 
 #include <catch2/catch.hpp>
 
-#include "fixtures.h"
-
 #include <faabric/scheduler/ExecGraph.h>
 #include <faabric/state/State.h>
 #include <faabric/state/StateServer.h>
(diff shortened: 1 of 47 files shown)


tests/utils/faabric_utils.h
```diff
@@ -2,8 +2,6 @@
 
 #include <catch2/catch.hpp>
 
-#include "fixtures.h"
```
csegarragonz, line 5: This is the offending line

PR #302, opened 2023-04-11
Use protobuf JSON utilities for JSON (de-)serialisation

### LINKED ISSUE #167: Replace use of RapidJSON with protobuf JSON utilities
The use of RapidJSON to manually parse JSON is now no longer necessary, and instead we can use [protobuf tools](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.util.json_util#JsonStringToMessage.details) to convert JSON coming from a client directly into a protobuf object.

[This file](https://github.com/faasm/faabric/blob/master/src/util/json.cpp) contains our only use of RapidJSON, so working upwards from there we should be able to completely remove the dependency. To make it a clean change, we could of course keep the existing functions but just change their implementation to use the protobuf equivalent. 

## Breaking changes to clients

Various clients rely on the current JSON field naming scheme (cpp client, Python client, and all the experiments), so if we can preserve this that would also reduce the impact of this change.

csegarragonz:
In this PR I switch from RapidJSON to protobuf-based JSON (de-)serialisation.

This change greatly simplifies the serialisation logic, and also makes it possible to (de-)serialise any other protobuf message that inherits from `google::protobuf::Message`.

The only downside is that the serialisation keys we manually set (i.e. `py_user` or `py_func`) are deeply embedded throughout the organisation's codebase. For example, in `faasmcli` or in the experiments repo. By default, protobuf sets its own keywords based on the proto file, but we can also force the serialisation name. We do the latter. I think I am not forgetting any keys, but there's no way to be really sure. I also add a regression test to make sure the keys we use elsewhere are always there in the form we expect them to be.

Closes #167


include/faabric/util/json.h
```diff
@@ -1,23 +1,19 @@
...
 #include <faabric/util/exception.h>
 
+#include <google/protobuf/util/json_util.h>
+
 namespace faabric::util {
-std::string messageToJson(const faabric::Message& msg);
+std::string messageToJson(const google::protobuf::Message& msg);
 
-faabric::Message jsonToMessage(const std::string& jsonIn);
+void jsonToMessage(const std::string& jsonStr, google::protobuf::Message* msg);
```
csegarragonz, line 10: I need to change this signature, as we can not return an instance of the abstract class `google::protobuf::Message`.

Instead, we take the argument as a pointer.

PR #319, opened 2023-06-02
Generate code coverage report with CodeCov

csegarragonz:
In this PR I add preliminary support to generate code coverage reports as part of the GHA runs, and upload them to CodeCov.

In subsequent PRs, as I get more used to it, I will tune in the behaviour of the codecov bot.

For the moment, this is what I can see in my browser:
image


csegarragonz:
@Shillaker log in to CodeCov is done through GitHub, so you should also be able to access the dashboard, let me know if that is not the case.

(If you manage to get into the dashboard, make sure to change the barnch to `codecov`)

PR #329, opened 2023-07-02
Simplify and re-factor ExecGraph

csegarragonz:
In this PR I change the process and inner-workings of requesting an execution graph.

First, execution graphs are now requested via an HTTP request to the planner (instead than a random worker).
If the planner has all messages for the execution graph available, it will return the execution graph.
Otherwise, it will return an error.

Second, I move all the execution graph logic from the scheduler to `src/util`. Getting the execution graph now becomes a matter of just getting the message result for all messages involved in an execution, and building the exec graph. Thus, the only method required is getting a function's result (which currently is done through the scheduler, but could change).


include/faabric/util/ExecGraph.h
csegarragonz, line 1: I merge on this header file the definitions re. the exec. graph in `include/faabric/scheduler/Scheduler.h` and `include/faabric/util/exec_graph.h`.

I neither add nor remove any definitions.

PR #378, opened 2024-02-15
mpi: manually heap-allocate payloads for local messages

csegarragonz:
This PR makes sure that, when sending a local MPI message, we make the minimum number of copies (of the message payload).

The minimum number of copies is two: (i) from the sender-owned buffer, to a runtime-owned buffer, and (ii) from the runtime-owned buffer to the receiver-owned buffer.

In the future, we could consider using something different than `std::malloc`.

### DIFF
--- a/tests/test/mpi/test_mpi_world.cpp
+++ b/tests/test/mpi/test_mpi_world.cpp
@@ -242,23 +242,6 @@ TEST_CASE_METHOD(MpiTestFixture, "Test send and recv on same host", "[mpi]")
     world.send(
       rankA1, rankA2, BYTES(messageData.data()), MPI_INT, messageData.size());
 
-    /*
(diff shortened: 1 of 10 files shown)


lgarithm:
TODO: 
- use memory pool instead of malloc and free.

tests/test/mpi/test_mpi_world.cpp
```diff
@@ -242,21 +242,6 @@ TEST_CASE_METHOD(MpiTestFixture, "Test send and recv on same host", "[mpi]")
     world.send(
       rankA1, rankA2, BYTES(messageData.data()), MPI_INT, messageData.size());
 
-    SECTION("Test queueing")
```
csegarragonz, line 245: This does not apply as the message buffer is not in the queue any more.

PR #391, opened 2024-03-08
upstream a cleaned version of MPI allreduce benchmark

lgarithm:
add mpi allreduce benchmark

### DIFF
--- a/tests/dist/mpi/CMakeLists.txt
+++ b/tests/dist/mpi/CMakeLists.txt
@@ -8,6 +8,7 @@ target_link_libraries(faabric_dist_tests_mpinative_lib PUBLIC faabric::test_util
 
 # Second we set all the source files for the dist tests library
 set(ALL_MPI_EXAMPLES
+    ${CMAKE_CURRENT_LIST_DIR}/benchmarks/mpi_allreduce.cpp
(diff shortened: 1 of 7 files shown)


lgarithm:
Hi @csegarragonz, could you run the github action to see if this breaks the CI, thanks.

csegarragonz:
@lgarithm there seems to be a problem with the permissions when I approve your run, i'll take a look

csegarragonz:
it [seems](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow) that gha secrets are not passed to runners running a fork of the repo.

truth is that we do not really need the secrets for the main tests workflow, as we are only pulling from ACR.
we still need them for release workflows, but that's ok.

hopefully #392 fixes this.

csegarragonz:
@lgarithm could you rebase to the latest `main`?

lgarithm:
@csegarragonz I pushed the rebase.

csegarragonz:
@lgarithm this is looking good.

can you rebase again (forgot to remove another usage of secrets), and then run:

```
inv format-code
```

(from _inside_ the container) to format the c++ code?

PR #385, opened 2024-02-26
mpi: use boost's lockfree ringbuffer for local messaging

csegarragonz:
todo: probably want to add some locking around the recv rather than just crazyily spinlock

closes #383

PR #382, opened 2024-02-23
util(queue): explore different in-memory queue implementations

csegarragonz:

PR #395, opened 2024-03-13
wip: test transport layer improvements

csegarragonz:
This PR combines:
* spinlock queue for mpi: #385 
* c-struct for mpi messages: #379
* mimalloc memory allocator: #387 (this seems to not improve our benchmarks)
* transport layer improvements:
  *  c-struct for ptp messages: #388

we will merge them separately, but ideally we want to have both PRs above green, and then measure the benefits (hopefully we are on par with #382)


lgarithm:
The results for the 3 commits are here: https://github.com/lgarithm/faabric/pull/2#issuecomment-1994419719

PR #432, opened 2024-04-21
mpi: fix setting local ranks

csegarragonz:
In very constrained environments like GHA it may sometimes happen that some ranks _completely_ finish (i.e. call destroy) before others have even called Init.

With our current implementation, this meant that the world may have been removed from the registry, even though there were still active local ranks to run.

To fix this, we set the number of active local ranks once, at the beginning, and decrement it on a per-thread basis. Note that this does not invaliate the fix to the previous race condition because, in fact, what we fixed is that we now _always_ init a world when we execute in it.

### DIFF
--- a/include/faabric/mpi/MpiWorld.h
+++ b/include/faabric/mpi/MpiWorld.h
@@ -212,7 +212,11 @@ class MpiWorld
     std::string thisHost;
     faabric::util::TimePoint creationTime;
 
-    std::atomic<int> activeLocalRanks = 0;
+    // Latch used to clear the world from the registry when we are migrating
+    // out of it (i.e. evicting it). Note that this clean-up is only necessary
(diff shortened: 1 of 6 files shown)


lgarithm:
```
---------------------------------------------------------------------------------------------------------
local/S       local/L       remote/S      remote/L      commit                                         
---------------------------------------------------------------------------------------------------------
0.294         8.855         0.020         8.665         mpi                                            
0.130 (0.44)  5.470 (0.62)  0.008 (0.40)  4.254 (0.49)  5b65b3e77d5726d6e22748db930e75642ed7672a # main
0.139 (0.47)  5.480 (0.62)  0.008 (0.40)  4.227 (0.49)  4db5ae32160aa1299d920c17b4196716d1ba29f7 # main
```

